Issue #2, opened 2018-05-12
Scaffold Playground Layout


ningyuansg:
For now, we should focus on the playground.

PR #8, opened 2018-05-16
Scaffold playground page

ningyuansg:
Addresses #2, _'Scaffold Playground Layout'_.

- Playground is a child of application
- Playground is associated with new state `IPlaygroundState` which is a child of `IState`
- `IPlaygroundState` stores the contents of the playground editor, allowing it to persist between pages
- Use the library `react-ace` to implement the editor

### DIFF
--- a/src/actions/playground.ts
+++ b/src/actions/playground.ts
@@ -8,7 +8,7 @@ export const UPDATE_EDITOR_VALUE: string = 'UPDATE_EDITOR_VALUE'
 
 /**
  * Represents an `Action` which updates the `editorValue` of a
- * `IPlaygroundState` 
+ * `IPlaygroundState`
  * @property type           - Unique string identifier for this `Action`
  * @property newEditorValue - The new string value for `editorValue`
  */
@@ -18,7 +18,7 @@ export interface IUpdateEditorValue extends Action {
 }
 
 /**
- * An `ActionCreator` returning an `IUpdateEditorValue` `Action` 
+ * An `ActionCreator` returning an `IUpdateEditorValue` `Action`
  * @param newEditorValue - The new string value for `editorValue`
  */
 export const updateEditorValue: ActionCreator<IUpdateEditorValue> = (newEditorValue: string) => ({
--- a/src/actions/playground.ts
+++ b/src/actions/playground.ts
@@ -1,12 +1,13 @@
 import { Action, ActionCreator } from 'redux'
 
-export type UPDATE_EDITOR_VALUE = 'UPDATE_EDITOR_VALUE'
-const UPDATE_EDITOR_VALUE: UPDATE_EDITOR_VALUE = 'UPDATE_EDITOR_VALUE'
+export const UPDATE_EDITOR_VALUE: string = 'UPDATE_EDITOR_VALUE'
+
 export interface IUpdateEditorValue extends Action {
-  type: UPDATE_EDITOR_VALUE
+  type: string
   newEditorValue: string
 }
+
 export const updateEditorValue: ActionCreator<IUpdateEditorValue> = (newEditorValue: string) => ({
-  UPDATE_EDITOR_VALUE,
+  type: UPDATE_EDITOR_VALUE,
   newEditorValue
 })
--- a/src/actions/playground.ts
+++ b/src/actions/playground.ts
@@ -1,12 +1,26 @@
 import { Action, ActionCreator } from 'redux'
 
+/**
+ * The `type` attribute for an `Action` which updates the `IPlaygroundState`
+ * `editorValue`
+ */
 export const UPDATE_EDITOR_VALUE: string = 'UPDATE_EDITOR_VALUE'
 
+/**
+ * Represents an `Action` which updates the `editorValue` of a
+ * `IPlaygroundState` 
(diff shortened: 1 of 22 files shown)


src/reducers/playground.ts
```diff
@@ -0,0 +1,34 @@
+import { Action, Reducer } from 'redux'
+import { IUpdateEditorValue, UPDATE_EDITOR_VALUE } from '../actions/playground'
+
+/**
+ * A state for the playground container
```
evansb: In general, since we have established a convention, I think the comment can be terser and only to describe non-obvious name or to explain a complex routine/function.

I think we can go with no comment by default, the reviewer will be the one responsible for requesting documentation if deemed necessary, what do you think?

src/actions/playground.ts
```diff
@@ -0,0 +1,27 @@
...
+export const UPDATE_EDITOR_VALUE: string = 'UPDATE_EDITOR_VALUE'
+
+/**
+ * Represents an `Action` which updates the `editorValue` of a
+ * `IPlaygroundState`
+ * @property type           - Unique string identifier for this `Action`
+ * @property newEditorValue - The new string value for `editorValue`
+ */
+export interface IUpdateEditorValue extends Action {
+  type: string
```
evansb: This should be redundant, all `Action` is expected to have a type field, please verify

src/actions/playground.ts
```diff
@@ -0,0 +1,27 @@
...
+  newEditorValue: string
```
evansb: You should follow the Redux FSA(Flux Standard Action) convention. The payload of the action should be under the name `payload`. A payload is typically an object in complex action, but in this case I think a string suffice.

src/components/Application.tsx
```diff
@@ -3,23 +3,24 @@ import * as React from 'react'
...
 import NavigationBar from './NavigationBar'
 import NotFound from './NotFound'
 
 export interface IApplicationProps extends RouteComponentProps<{}> {
-  application: IApplicationState
+  title: string
 }
 
-const Application: React.SFC<IApplicationProps> = ({ application }) => {
+const Application: React.SFC<IApplicationProps> = (props: IApplicationProps) => {
```
evansb: Adding `IApplicationProps` after `props:` is redundant because we already declare the type as `React.SFC<IApplicationProps>`. TypeScript is able to infer the parameter type. So you can unpack `{ title }` straight away in the parameter.

src/actions/playground.ts
```diff
@@ -0,0 +1,27 @@
+import { Action, ActionCreator } from 'redux'
+
+/**
+ * The `type` attribute for an `Action` which updates the `IPlaygroundState`
+ * `editorValue`
+ */
+export const UPDATE_EDITOR_VALUE: string = 'UPDATE_EDITOR_VALUE'
```
evansb: `string` is redundant

evansb: When in doubt, you can check with your  editor /IDE, if upon hovering to the variable you see a type, that means the type can be inferred.

Of course sometimes you want to explicitly declare it anyway, for documentation purposes

src/components/Playground.tsx
```diff
@@ -0,0 +1,36 @@
...
+import 'brace/theme/github'
+
+/**
+ * A prop that is passed to the Playground
+ * @property editorValue - The string content of the react-ace editor
+ * @property updateCode  - A callback function for the react-ace editor's `onChange`
+ */
+export interface IPlaygroundProps {
+  editorValue: string
+  updateCode: (newCode: string) => void
```
evansb: Usually the convention for event handler begins with `handle<event-name>`, i.e `handleEditorChange`

src/components/Playground.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ * A component representing the Playground
+ */
+export default class Playground extends React.Component<IPlaygroundProps, {}> {
```
evansb: You should always default to `React.SFC` until you need state

src/actions/playground.ts
remo5000: To clarify, do you mean the naming convention only? Or do you mean that we should keep the`payload` object as an arbitrarily typed object (like in most example online)? I feel that the typing should remain strict for a particular payload.

src/components/Playground.tsx
remo5000: In this case, our editor would need to call upon a `dispatch` to update the `editorValue` within the `IPlaygroundState`, whenever someone edits code in the ace editor. Is a Component warranted in this case?

src/components/Application.tsx
remo5000: I feel like it would be better to keep a `props` argument, so that adding properties to `IApplicationProps` does not require an additional change in the arguments. What do you think?

remo5000:
@evansb I've made the changes requested, thanks for the feedback

src/components/Playground.tsx
evansb: Yeah, so in this case the editor component is stateful, but the playground component itself should be stateless

remo5000: Ah, I understand now. Will do!

Issue #14, opened 2018-05-17
Make all state props `readonly`


remo5000:
This will allow us to conform to the Redux principle [https://redux.js.org/introduction/three-principles#state-is-read-only](url)

Should be pretty straightforward, but I'm adding it here in case we forget.

Issue #11, opened 2018-05-17
Use `bindActionCreators`


remo5000:
In accordance with the guide, and more importantly, cleaner code.

Issue #7, opened 2018-05-16
State definitions should be consolidated


remo5000:
States, that are ultimately part of `IState`, should be defined in one folder (similar to how source-academy-2 defined `Shape`).

ningyuansg:
Addressed in a3e60f3.

PR #19, opened 2018-05-20
Add a Functioning IDE for the playground

remo5000:
### Features added
- Added `slang` (hence the numerous file additions)
- Added a REPL with the same evaluation and clear functionality as source-academy
- Abstracted the editor and REPL into an  `IDE` container (to maintain reusability)

### Issues fixed
- #7
- #6
- #11
- #14

#### Notes:
- An additional rule of tslint, `no-empty` has been turned off. We feel that semantically, a no-op is easier to understand with empty curly braces as opposed to `null`. Input on this would be appreciated, as we can easily reinstate the rule.
- We violate Redux's suggestion of immutable states in the case of the Source's Context object. Should be okay, since the Context is not used to render anything, and making a deep copy of Context would be slow for large programs.

#### What's next:
- ~~~We still need to implement display. That's next in line.~~~ Done.
- ~~~Following that are style changes, and the use of SCSS (#5).~~~ Done with #21.


src/reducers/playground.ts
```diff
@@ -1,28 +1,106 @@
...
+      return {
+        ...state,
+        output: []
+      }
+    case CLEAR_CONTEXT:
+      return {
+        ...state,
+        context: createContext()
+      }
+    case HANDLE_CONSOLE_LOG:
```
evansb: Not sure what's the benefit of grouping console logs. In fact, under certain circumstances this might not be what you want (e.g what if the console log is done asynchronously?)

src/reducers/playground.ts
```diff
@@ -1,28 +1,106 @@
...
+       * (2) state.output[-1] is not RunningOutput
+       * (3) state.output[-1] is RunningOutput */
+      lastOutput = state.output.slice(-1)[0]
+      if (lastOutput === undefined || lastOutput.type !== 'running') {
+        newOutput = state.output.concat({
+          type: 'running',
+          consoleLogs: [action.payload]
+        })
+      } else {
+        lastOutput.consoleLogs = lastOutput.consoleLogs.concat(action.payload)
```
evansb: this mutates `oldstate.output[oldstate.output-1].consoleLogs`, which in Redux is an antipattern

ningyuansg:
Sorry, we realise this PR is kinda big. Will cut them down in size in the future (~1 issue/PR).

evansb:
Since this PR is blocking, let's resolve the comments later.

Hopefully this will be the last huge PR, in general it's very hard to review but I understand the intent so this is ok

src/reducers/playground.ts
remo5000: The intent would be to group the `display()` outputs of a particular program together in one Card, as compared to one card per output (as in sourceacademy), so as to increase the readability of one's program outputs (code output, displays and then the result). 

I did not consider asynchronous outputs previously, it could cause problems with the solution here. @ningyuansg any thoughts?

ningyuansg: We are trying to have one output per program. That is, each program results in exactly one card being displayed (with the exception of CodeOutput). 

I recorded a comparison [here](https://youtu.be/8RArdxjIuHA). The distinction between logs and results/errors will be more prominent after we add text colour with code tags.

The aim is to emulate normal behaviour in most shells. When a program is run, the user is only given back the prompt after stdout/stderr has finished printing their stuff. The area of output between prompts (`user $`) is analogous to our `<Card />` block elements.

Async certainly deviates from this behaviour, but the reducer is smart enough to handle it at least without crashing. Async is also pretty rare in CS1101S. It was present only as a 6th MC sidequest last year, and considering next year's CS1101S is 4 MCs the source language may not need to support async.

src/components/IDE/index.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+
+import ControlContainer from '../../containers/IDE/ControlContainer'
+import EditorContainer from '../../containers/IDE/EditorContainer'
+import ReplContainer from '../../containers/IDE/ReplContainer'
+
+const IDE: React.SFC<{}> = () => (
```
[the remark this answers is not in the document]
ningyuansg: Created issue #34 for this refactor. Right now there are quite a number of changes directly related to this component, we predict that refactoring the name right now will cause quite a mess. It should be a smoother process once we merge in the changes to IDE and start working on other components (dashboard, etc.) in the future. Meanwhile, we keep track of this as an issue (so it's not forgotten).

PR #28, opened 2018-05-23
Add corrections for #19

remo5000:
#19 

### Changes
- Removed mutation of state


remo5000:
@evansb We'll be pushing the corrections here for review!

Issue #26, opened 2018-05-23
Ternary expressions should only accept boolean first arguments


ningyuansg:
Probably part of RTTC.

ningyuansg:
Hi Prof @martin-henz,

Currently, the specifications state that ternary expressions will only accept boolean expressions as their first argument (_"Preceding `?`, Source only allows boolean expressions"_). Should we also implement the same restriction for `if` statements?

Meaning, should the type of `a` in the following program be required evaluate to a boolean value?
```javascript
if (a) {
    0;
} else {
    1;
}
```

martin-henz:
Yes, that was a gap in the specs. It should say:

Preceding ? and following if, Source only allows boolean expressions.

I've pushed the change in the specs.

PR #29, opened 2018-05-23
Restrict conditions to booleans values, for IfStatements and ConditonalExpression

ningyuansg:
This PR depends on #24 as it uses the convenience function `isBool` and test file `src/slang/utils/__tests__rttc.ts`.

---

### Summary

- In the programs `a ? b : c;` or `if (a) { 1; } else { 0; }`, `a` must be a boolean value, or a error is raised, and the program is stopped.

### Issues fixed

- #26

### DIFF
--- a/src/slang/utils/__tests__/rttc.ts
+++ b/src/slang/utils/__tests__/rttc.ts
@@ -1,6 +1,6 @@
 import { BinaryOperator, UnaryOperator } from 'estree'
 import { mockClosure, mockRuntimeContext } from '../../../mocks/context'
-import { checkBinaryExpression, checkLogicalExpression, checkUnaryExpression } from '../rttc'
+import * as rttc from '../rttc'
 
 const num = 0
 const bool = true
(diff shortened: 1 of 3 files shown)


src/slang/utils/__tests__/rttc.ts
```diff
@@ -169,7 +169,21 @@ test('Invalid logical type combinations return TypeError', () => {
...
+  const context = mockRuntimeContext()
+  const errors = operatorValues.map(opVal => rttc.checkIfStatement(context, opVal))
+  errors.map(error => expect(error).toBe(undefined))
+})
+
+test('Invalid ternary/if test expressions return TypeError', () => {
+  const operatorValues = [num, str, func]
+  const context = mockRuntimeContext()
+  const errors = operatorValues.map(opVal => rttc.checkIfStatement(context, opVal))
+  errors.map(error => expect(error).toBeDefined())
```
evansb: let's make this check if error is instance of `TypeError`

Issue #38, opened 2018-05-26
Follow Up on Review of PR29


evansb:
https://github.com/source-academy/cadet-frontend/pull/29#pullrequestreview-123547153

PR #45, opened 2018-05-26
Improve tests for slang rttc

ningyuansg:
### Features

- 100% coverage for `src/slang/utils/rttc.ts`
- Better use of jest methods, `toBeUndefined`, `toBeInstanceOf`
- Better mock Closure which uses the Closure constructor

```
  misc.ts            |    44.44 |        0 |        0 |    44.44 |... 48,49,51,53,59 |
  object.ts          |        0 |      100 |        0 |        0 |               3,4 |
 src/slang/utils     |    70.87 |    68.29 |       75 |    69.39 |                   |
  node.ts            |    33.33 |     7.14 |    44.44 |    33.33 |... 08,114,115,116 |
  rttc.ts            |      100 |      100 |      100 |      100 |                   |
---------------------|----------|----------|----------|----------|-------------------|

Test Suites: 10 passed, 10 total
Tests:       32 passed, 32 total
Snapshots:   184 passed, 184 total
Time:        5.612s
```

### Issues fixed

- #38 which is a follow-up on the review for #29

### DIFF
--- a/src/slang/utils/rttc.ts
+++ b/src/slang/utils/rttc.ts
@@ -12,7 +12,7 @@ import {
 const LHS = ' on left hand side of operation'
 const RHS = ' on right hand side of operation'
 
-class TypeError implements SourceError {
+export class TypeError implements SourceError {
   public type = ErrorType.RUNTIME
   public severity = ErrorSeverity.ERROR
   public location: es.SourceLocation
@@ -26,7 +26,7 @@ class TypeError implements SourceError {
   }
 
   public elaborate() {
-    return 'TODO'
+    return this.explain()
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/slang/utils/rttc.ts
```diff
@@ -51,15 +51,12 @@ export const checkUnaryExpression = (
...
-  switch (operator) {
-    case '+':
-      return isNumber(value) ? undefined : new TypeError(node, '', 'number', typeOf(value))
-    case '-':
-      return isNumber(value) ? undefined : new TypeError(node, '', 'number', typeOf(value))
-    case '!':
-      return isBool(value) ? undefined : new TypeError(node, '', 'boolean', typeOf(value))
-    default:
-      return
+  if ((operator === '+' || operator === '-') && !isNumber(value)) {
```
remo5000: I feel like the switch case used previously is clearer, although simplifying the logic would help with more operators

ningyuansg: So the thing is that with the switch case, in order to get 100% test coverage there would need to be a call to `checkUnaryExpression(context, ''. value)`. Otherwise, the default case would not be covered by the tests. 

So assuming we want 100% test coverage, it was either that, or to change the switch case to the if/else here. Since there isn't much sense in checking for an empty operator, I decided to go with the latter instead.

remo5000: Can the default case be used to test for an unspecified unary operator (e.g `delete` or `typeof`)? If so, it can be meaningfully used

ningyuansg:
Had to do some git stuff to get github to mark this PR as merged after merging on the CLI. 

(refer to 1559e07 for the changes)

ningyuansg:
Nevermind, it still marks this PR as closed instead of merged.

src/slang/utils/rttc.ts
ningyuansg: That sounds like a good idea, tbh. 

Unspecified operators are currently checked for in the linting step at at [`noUnspecifiedOperators`](https://github.com/source-academy/cadet-frontend/blob/1559e07a196e3260fafa35ac308cf589861f2c10/src/slang/rules/noUnspecifiedOperator.ts). Checking the operators again during runtime seems redundant at first, but I feel like it's extra safety without any real performance hits.

ningyuansg: @evansb Can we check, what's the best practice here?

Issue #52, opened 2018-05-28
Use specific slices of states in containers


remo5000:
If the container pertains to a particular slice of the state, I believe that we can use that part only in `mapStateToProps`.

PR #65, opened 2018-05-30
Use specific slices of State in containers

remo5000:
Issues fixed:
#52


remo5000:
My bad, turns out that the state is just casted.

PR #73, opened 2018-05-31
Scaffold announcement posts

remo5000:
Blocked by #70. 

Will change the base and unblock one simplify-menu is merged.


src/components/device/Announcements.tsx
```diff
@@ -38,4 +33,13 @@ class Announcements extends React.Component<IAnnouncementsProps, {}> {
   }
 }
 
+
+const AnnouncementCard: React.SFC<{ann : Announcement, key : number}> = props => (
```
remo5000: @ningyuansg Im getting a warning for using `key` here:
```
Warning: AnnouncementCard: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)
```

PR #77, opened 2018-06-01
Add green screen mode to playground

ningyuansg:
### Features

- Add a green screen mode to the playground
- Green screen mode hides the navigation bar, and changes backgrounds to `#00ff00`
- Some elements have a border of `#00e000` dark green. It is otherwise difficult to tell where things like the editor and the repl end
- Green screen mode is activated with a sequence of keys while focus is in the playground (currently, the sequence of keys is _'hulk'_)
- Green screen mode is irreversible for the page (refresh to return to the default colour scheme)

2018-06-01-124413_1920x1080_scrot

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "react": "^16.3.1",
     "react-ace": "^6.1.1",
     "react-dom": "^16.3.1",
+    "react-hotkeys": "^1.1.4",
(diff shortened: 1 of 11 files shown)


ningyuansg:
@remo5000 Good point, done with 01309b9.

PR #82, opened 2018-06-05
Add game

ningyuansg:
### Features

- Game (previously story-xml) is represented by the Game component
- Pixi stuff only loads when the game component is called
- The canvas is preserved between routes (i.e. clicking away into another component and back to game does not re-render game)
- The canvas resizes dynamically with the browser window, maintaining it's aspect ratio
- Redirects within the game do not reload the page (react-router-style redirects)
- Move history from `index.tsx` into its own file


src/components/game/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+     * will show nothing but a black screen. This means that if the user
+     * navigate aways from the game tab, and then back again, the game would not
+     * work.
+     *
+     * So, we save a reference to the first canvas that is loaded. Thereafter,
+     * when this component is mounted, use that canvas instead of the new canvas
+     * mounted with this div. This is a bit hacky, and refs aren't favoured in
+     * react, but it also prevents excessive loading of the game
+     */
+    if (this.props.canvas === undefined) {
```
remo5000: @evansb I think the loading of the game takes up quite a bit of time, and this would be a good way to maintain continuity in the SPA. Is there any better way/optimisation that you might be able to advise us on?

Issue #81, opened 2018-06-04
Overall structure of Source Academy UX


martin-henz:
**Current structure**

Cadet | DEVICE                                                    </> PLAYGROUND

Under "DEVICE":
Missions | Side Quests | Paths | Contests             Announcements |  Materials | Profile

**Proposal**

Source Academy | News | Material | Playground                           <USER_ICON> Thomas Tan

Under "Source Academy":
Missions | Side Quests | Paths | Contests

Notes:
* "Materials" renamed to "Material" (that's grammatically better, see https://english.stackexchange.com/questions/247497/material-vs-materials-which-is-correct)
* "Cadet" renamed to "Source Academy"
* "Announcements" renamed to "News"
* The <USER_ICON> could come from the IVLE student picture.
* Clicking <USER_ICON> or "Thomas Tan" will take the users to their Profile
* Before the first login, there would be "Login" instead of "<USER_ICON> Thomas Tan"
* I'm assuming that we treat News, Material and Playground as public information; anyone can see this without login.
* Before the first login, the clicking of "Source Academy" is disabled, and no Missions etc are displayed.

martin-henz:
The toplevel could also consist of pictograms:
Source Academy | News | Material | Playground <USER_ICON> Thomas Tan
-> (for example)
* ![](https://sourceacademy.comp.nus.edu.sg/images/logo.gif)
* ![](https://www.imca-int.com/wp-content/themes/imcaseventeen/img/imca-icons/post.svg)
* ![](https://www.amd.com/system/files/11050-partner-marketing-materials-icon-960x458.png)
* ![](http://www.fullbasketproperty.com/uploads/amenities/b9026fbb68aac397c6615e0a9a9cedb4.png)
* <USER_ICON> Thomas Tan

remo5000:
Sounds good!

remo5000:
I think a combination of an icon and the word works well.  I have added the admin and game buttons as well (as they have been scaffolded). Is this something like what you had in mind? @martin-henz 

![image](https://user-images.githubusercontent.com/12512488/40961325-d474eba8-68d4-11e8-8dfc-c93d5df55d4b.png)

remo5000:
@martin-henz 
- We shall address the space issue when resolving #76.
- The Game button is for the "story". I can move it into Source Academy.
- The article makes a lot of sense, I'll remove the all caps. Thank you for the feedback! 😄 

[...]

PR #85, opened 2018-06-05
Change UX structure

remo5000:
### Features
- Moved Announcements and Materials out of device
---
### Issues fixed:
- #81 
- #87 (Because Announcements had to be moved out)


src/components/commons/ContentDisplay.tsx
```diff
@@ -13,10 +14,18 @@ class ContentDisplay extends React.Component<IContentDisplayProps, {}> {
   public render() {
     return (
       <div className="ContentDisplay row center-xs">
-        <div className="col-xs-10">{this.props.display}</div>
+        <div className="col-xs-10 contentdisplay-content-parent">
```
remo5000: @ningyuansg Do you think we should rename `ContentDisplay`?

PR #89, opened 2018-06-06
Create resizable workspace side content

ningyuansg:
### Features

- Side content renders on right of workspace by default
- Side content is resizable vertically
- Side content can be hidden by dragging to the top, or expanded by dragging to the bottom
- Absolute height of side content is persistent on dismount/remounting
- Draggable bottom border of side content is never offscreen (changes dynamically with screen height)
- Cleaned up scss a bit


src/components/workspace/side-content/index.tsx
```diff
@@ -0,0 +1,36 @@
+import { Card, Text } from '@blueprintjs/core'
+import * as React from 'react'
+
+const SideContent: React.SFC<{}> = () => (
+  <div className="side-content">
```
remo5000: Can this be changed to `SideContent`, to be consistent with the other components?

remo5000:
Added snapping for the Editor as well 👨‍💻

PR #106, opened 2018-06-11
Integrate IVLE login

ningyuansg:
### Features

- The login button now connects to the IVLE API
- Once logged in, the username is fetched via a saga

### Dev notes

1. The default port for `yarn start` has been changed to 80. This allows the callback url given to IVLE to work correctly. Since most OS's require root to serve at port 80, you will probably have to `sudo` with yarn start.
2. The IVLE LAPI token is saved at `src/utils/secrets.ts`. Add yours, then set git's assume-unchanged on it.
3. **Username fetching will not work on most browsers**. This is because the IVLE api does not set an appropriate [`allow-control-allow-origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) property, so requests from the browser that are from a domain that is _not_ nus.edu.sg will fail.
4. **To bypass this, set an appropriate flag on the chromium browser and view the site through there.** Instructions from SO [here](https://stackoverflow.com/a/3177718/6910451).
5. As discussed, actions sent to and taken by saga should be prefixed with fetch (e.g `fetchUsername`) _where appropriate_, and actions dispatched by saga to change the state are prefixed with set (e.g. `setUsername`)


ningyuansg:
@rrtheonlyone Let me know what you think. Right now, the plan is to fetch content from our news tab from IVLE's announcements, and our materials from IVLE's files. With this PR, there is no check if a student takes CS1101S or not. I'm not familiar with the IVLE API docs (in fact, I'm not sure I used the right one, since the docs for this say there's a new API, but I couldn't find that new one), so you'll have to guide us on this one.

src/sagas/index.ts
```diff
@@ -7,8 +7,8 @@ import * as actions from '../actions'
 import * as actionTypes from '../actions/actionTypes'
 import { IState } from '../reducers/states'
 import { Context, interrupt, runInContext } from '../slang'
-import { history } from '../utils/history'
 import { showSuccessMessage, showWarningMessage } from '../utils/notification'
+import { IVLE_API_KEY } from '../utils/secrets'
```
ningyuansg: 1. We should request an API key specifically for this frontend rather than use our own student ones (obviously).
2. I considered hashing this, but it'll be exposed through the browser GET call anyways.

src/components/academy/index.tsx
```diff
@@ -28,8 +33,18 @@ export const Academy: React.SFC<AcademyProps> = props => (
   </div>
 )
 
-const checkLoggedIn = (props: AcademyProps) =>
-  props.token === undefined ? <Route component={redirectToLogin} /> : undefined
+const checkLoggedIn = (props: IAcademyProps) => {
+  const token = qs.parse(props.location.search).token
+  if (token !== undefined) {
+    props.changeToken(token) // just received a callback from IVLE
```
remo5000: I think this call could use an API call (without further redirect) to check the validity of the token. Redirection is a possible security threat. For example, (`http://localhost/academy?token=SPONGEBOBSQUAREPANTS`) creates a valid token in the state, allowing me to view the game (albeit without any relevant content). It is a matter of debate as to *when* the token is to be validated (as discussed in real life @ningyuansg), but adding an extra check here prevents the user from seeing data that is already present in the website (e.g frontend stuff)

ningyuansg: - API calls take time, so adding more calls reduces performance
- Invalid token will throw an error with `fetch` anyways, in these cases we catch the error, set `session.token` to undefined, and redirect the user to login (perhaps with a toaster error message)
- If the data is already present in our state, the user must have been logged in before, in a valid manner. Otherwise, the data would not have been fetched. There is no way for 'illegitimate' data to be fetched by an invalid token. Thus, I don't believe this is a security risk

You can document this in an issue though

src/sagas/index.ts
```diff
@@ -50,8 +50,21 @@ function* interpreterSaga(): SagaIterator {
...
+    window.location.href = `${apiLogin}?apikey=${key}&url=${callback}`
+    yield undefined
+  })
+
+  yield takeEvery(actionTypes.FETCH_USERNAME, function*() {
+    const apiUsername = 'https://ivle.nus.edu.sg/api/Lapi.svc/UserName_Get'
+    const key = IVLE_API_KEY
+    const token = yield select((state: IState) => state.session.token)
+    const username = yield call(() =>
+      fetch(`${apiUsername}?APIKey=${key}&Token=${token}`).then(response => response.json())
```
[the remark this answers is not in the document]
ningyuansg: Username fetch happens right after login, which depends on IVLE being up (i.e. its guaranteed by a 'soft' condition). So I don't think the error handling is a priority until we deal with fetching more substantial IVLE data e.g. news (announcements), files (material).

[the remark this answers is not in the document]
remo5000: Ah ok

src/components/academy/index.tsx
remo5000: Sure. I'm not very clear about such problems myself, it would be nice to get some more input

PR #110, opened 2018-06-14
Store IVLE API key in environment variable

ningyuansg:
### Features

- The IVLE Lapi key is now saved as an environment variable `REACT_APP_IVLE_KEY`. The prefix REACT_APP is a requirement from the react framework. This is reflected in the README.
- New pre-push test script, now stashes unstaged changes before running format and jest tests.
- Update README's project structure.

### Issues fixed

- #109

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -31,8 +31,11 @@ main() {
 run_cmd() {
     echo_cyan "> $1"
 
-    $1 2>&1 1> /dev/null | sed 's/^/  /'
+    print=$($1 2>&1 1> /dev/null)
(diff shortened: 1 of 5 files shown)


scripts/test.sh
```diff
@@ -0,0 +1,60 @@
+#! /usr/bin/env bash
+
+export CI=true
+
+main() {
+    run_cmd "git stash push --keep-index --message precommit"
+    echo "  If you cancel this pre-push hook, use \`git stash pop\` to retrieve your"
```
remo5000: Good addition!

Issue #121, opened 2018-06-21
Parser rules not checked when minified


remo5000:
Expected: An error about only allowing empty list

Got: An array returned. Nothing can be done with the array because of a `no-member` rule.

ningyuansg:
This bug only occurs after minifying the files.

ningyuansg:
Most of the stuff in `src/slang/rules` don't work. Possibly the errors are not being pushed at `src/slang/parser.ts`.

ningyuansg:
![issue121](https://user-images.githubusercontent.com/22320312/41703144-e7f4a7c6-7564-11e8-8ee2-b924fe7f0d84.png)

Seems something went wrong adding stuff to the `walkers` object.

ningyuansg:
Bugged [Scope](https://stackoverflow.com/questions/45506745/what-is-scopes-in-dispatch-of-redux) (minified):
```javascript
0: Closure (Object.keys.forEach.t)
    t: "ArrayExpression"
1: Closure (s.default.forEach.e)
    e: {name: "no-non-empty-list", disableOn: 9, checkers: {…}}
2: Closure
    a: {default: {…}, __esModule: true}
    c: {ErrorType: {…}, ErrorSeverity: {…}, Closure: ƒ, ArrowClosure: ƒ, __esModule: true}
    d: class d
    f: (e,n)=> {…}
    h: class h
    i: {simple: ƒ, ancestor: ƒ, recursive: ƒ, full: ƒ, fullAncestor: ƒ, …}
    l: class l
    m: (n,r)=> {…}
    o: {…}
    p: {Program: ƒ, ExpressionStatement: ƒ, IfStatement: ƒ, FunctionDeclaration: ƒ, VariableDeclaration: ƒ, …}
    r: {…}
    t: {DisallowedConstructError: ƒ, FatalSyntaxError: ƒ, MissingSemicolonError: ƒ, TrailingCommaError: ƒ, freshId: ƒ, …}
    u: class u
3: Global
```

Ok Scope (not minified):
```javascript
0: Closure (compose)
    w1: (node, context) => {…}
    w2: (node, context) => {…}
1: Closure (./src/slang/parser.ts)
    DisallowedConstructError: class DisallowedConstructError
    FatalSyntaxError: class FatalSyntaxError
    MissingSemicolonError: class MissingSemicolonError
    TrailingCommaError: class TrailingCommaError
    compose: ƒ compose(w1, w2)
[...]

PR #126, opened 2018-06-21
Fix parser not checking rules/* when minified

ningyuansg:
### Features

- Added acorn as a dependency. We were able to use this before without having it as a project dependency because it was a required by some other dependency we were using. It's probably better practice this way, in case whatever was using acorn stopped using acorn (there would be a module resolution error then)
- Fixed yet another error with the minifying process. For details, see #121.

### Issues fixed

- #121

### DIFF
--- a/src/slang/parser.ts
+++ b/src/slang/parser.ts
@@ -95,7 +95,7 @@ export class TrailingCommaError implements SourceError {
   }
 }
 
-export function parse (source: string, context: Context) {
+export function parse(source: string, context: Context) {
   let program: es.Program | undefined
   try {
     program = acornParse(source, createAcornParserOptions(context))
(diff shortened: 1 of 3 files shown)


src/slang/parser.ts
```diff
@@ -163,43 +146,72 @@ const createAcornParserOptions = (context: Context): AcornOptions => ({
...
-  const keys = Object.keys(rule.checkers)
-  keys.forEach(key => {
-    walkers[key] = compose(walkers[key], (node, context) => {
-      if (typeof rule.disableOn !== 'undefined' && context.chapter >= rule.disableOn) {
-        return
+function createWalkers(
+  allowedSyntaxes: { [nodeName: string]: number },
+  parserRules: Array<Rule<es.Node>>
+) {
+  const newWalkers: { [name: string]: (node: es.Node, ctxt: Context) => void } = {}
```
remo5000: Is it possible to use ES6 Map here? since you do use the `[]` member notation to assign functions.

remo5000: `Array.from(x.entries()).reduce((obj, pair) => {obj[pair[0]] = pair[1]; return obj;}, {})`
should be able to convert from Map to Object

PR #127, opened 2018-06-22
Cache Redux Store

remo5000:
### Features
- Parts of the state such as `token`, `username` and the playground code persist when refreshing the browser. As a result, no need to login again when refreshing the page (!!)

### TODO
- There is an issue with redux-persist in the memory it uses (WebStorage). It works when running, but it is not available during tests, that are not on the web. I'll have to use a different type of storage for this.


ningyuansg:
Opened #129 for tests not working with redux-persist.

PR #134, opened 2018-06-25
Fix playground links making editor un-editable

remo5000:
### Features
The parsed program is now updated only once, when the page loads, rather
than when the component is updated.

This is not related to the readonly tag in the URL, which is not an implemented feature (yet).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "cadet-frontend",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "scripts-info": {
     "format": "Format source code",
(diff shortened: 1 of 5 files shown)


ningyuansg:
Need to fix shared url not being loaded when the url is loaded while the user is already in the playground. To replicate:

1. Load the playground (/playground)
2. Paste a shared url http://localhost/playground#lib=2&prgrm=CYSwzgDgNghgngCgOQAsCmUoHsCESCUA3EA

Expected bahaviour:
- The shared program `display('hello!');` loads into the editor without page refresh

Encountered behaviour:
- Nothing happens
- In order for the shared program to appear, user needs to manually reload the page

PR #137, opened 2018-06-26
Segregate Playground and Assessment Workspaces

remo5000:
### Features
- Playground and Assessment have separate Workspaces in the redux state
- Workspace is now stateless (not connected)

### TODO
- The `display` function needs to pass in a `location` argument, to be kept in the Context
- The Assessments Workspace does not allow editing of the Editor code (forced to be the solution template)
- The slice of the state for the Assessment's workspace must get reset every time one switches away from the question (e.g when they go to the menu, or when they go to the next question)


src/actions/interpreter.ts
```diff
@@ -1,21 +1,31 @@
...
-export const handleConsoleLog = (log: string) => ({
+// TODO fix this immediately after location
+// is implemented completely
+export const handleConsoleLog = (
+  log: string,
+  workspaceLocation: WorkspaceLocation = 'assessment'
+) => ({
   type: actionTypes.HANDLE_CONSOLE_LOG,
-  payload: log
+  payload: { log, workspaceLocation }
```
ningyuansg: `logString` is more descriptive imo

src/actions/workspace.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Used to differenciate between the sources of actions, as
+ * two workspaces can work at the same time. To generalise this
+ * or add more instances of `Workspace`s, one can add a string,
+ * and call the actions with the respective string (taken
+ * from the below enum).
+ *
+ * Note that the names must correspond with the name of the
+ * object in IWorkspaceManagerState.
+ */
+export type WorkspaceLocation = 'assessment' | 'playground'
```
ningyuansg: Can you use values from the enum instead of string literals?

src/components/Playground.tsx
```diff
@@ -6,12 +6,33 @@ import * as React from 'react'
...
+
+export interface IStateProps {
   editorValue: string
+  sideContentHeight: number
+  isRunning: boolean
+  activeTab: number
+  editorWidth: string
+  output: InterpreterOutput[]
+  replValue: string
+}
```
ningyuansg: Quite a lot of properties here, I think they could be alphabeticalised

remo5000:
Thanks for the quick review, have applied the changes 👍. I'll start on the other features now.

remo5000:
@ningyuansg I've added the "reset" feature. Do you think I could afford to keep `currentAssessmentId` and `currentQuestionId` in a tuple instead of separate properties?

src/reducers/states.ts
```diff
@@ -120,16 +128,25 @@ export const defaultApplication: IApplicationState = {
...
   context: createContext(latestSourceChapter),
-  editorValue: '',
+  editorValue: undefined,
   editorWidth: '50%',
   isRunning: false,
   output: [],
   replValue: '',
   sideContentActiveTab: 0,
-  sideContentHeight: undefined,
+  sideContentHeight: 60,
```
ningyuansg: Does this cause overflow? A fresh workspace should render such that the side content takes up a size that is just big enough to prevent overflow of its contents.

src/sagas/index.ts
```diff
@@ -15,7 +16,7 @@ function* mainSaga() {
   yield* apiFetchSaga()
   yield* interpreterSaga()
```
ningyuansg: Can consider renaming as workspaceSaga

ningyuansg:
Properties are more readable imo. Not sure about efficiency.

remo5000:
Changes made! @ningyuansg

Issue #138, opened 2018-06-27
redux-persist stores extra parts of the state


remo5000:
### Expected
Only environment, historyHelper, token, username stored.

### Actual
![image](https://user-images.githubusercontent.com/12512488/41953940-1bbf425a-7a0b-11e8-862b-7a9bbeacd389.png)

### Suggested approach
Abandon redux-persist, and write it our own way (there is a video on this by the creator of redux)

remo5000:
Resolved by #141

PR #139, opened 2018-06-27
Enable "display" standard library function

remo5000:
### Features
- `display` and `timed` work
- Context now has a generic type that can be used to provide external properties, named `externalContext`. In this case, it is a `WorkspaceLocation`.

### Issues fixed
- Fixed an issue with WorkspaceLocation enums (untracked as blocking PR not merged)

### Caveat
- `display` and `timed` do not work unless `redux-persist` is turned off. This is related to #138 , where it attempts to stores the Context (when it is not supposed to), or something related to the workspace, and hence throw an error on trying to serialise an object with a circular structure. Once #138  is resolved, the functions will work.

### DIFF
--- a/src/actions/workspaces.ts
+++ b/src/actions/workspaces.ts
@@ -13,7 +13,7 @@ import * as actionTypes from './actionTypes'
  */
 export enum WorkspaceLocations {
   assessment = 'assessment',
-  playground = 'playground',
+  playground = 'playground'
 }
 export type WorkspaceLocation = keyof typeof WorkspaceLocations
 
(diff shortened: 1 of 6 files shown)


remo5000:
Interesting. The handleConsoleLog actions seem to run before the render method is called, causing the displays to show up but the stop button not re-rendering. Really weird, good catch!

remo5000:
I've confirmed this to be a bug present previously, even in the master branch. I've opened up an issue (#142). 

[GIF](https://gfycat.com/CorruptWelloffKestrel)

Issue #129, opened 2018-06-23
Fix tests not working with redux-persist


ningyuansg:
See PR #127.

> There is an issue with redux-persist in the memory it uses (WebStorage). It works when running, but it is not available during tests, that are not on the web. I'll have to use a different type of storage for this.

remo5000:
I looked into this. The part of the code that throws the error checks for the propery `selfStorage` in the `Window`. Normally, the `selfStorage`'s value is `localStorage`, which cannot be created in a test environment as it is defaulted to a `WebStorage`. I'm thinking of overriding this by setting the property in the window as some sort of memory storage.

PR #141, opened 2018-06-28
Implement more reliable caching

remo5000:
### Features
- No more dependency on redux-persist (that caused the issues below)
- State is mapped more transparently onto a stored type (`IStoredState`) and stored onto localStorage
- Throttling is enabled for saving the state, and **for updating React-Ace editor values**.
- Empty playground reducer added for consistency, and preventing a jest error (wrong key 'playground' in state as there is no such reducer. The reducers passed to combineReducers must have names that [map to slices of the state](https://redux.js.org/api-reference/combinereducers)).

--- 

### Issues fixed:
- #138 
- #129

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     }
   },
   "dependencies": {
-    "@types/lodash": "^4.14.110",
     "acorn": "^5.7.1",
     "astring": "^1.3.0",
     "common-tags": "^1.7.2",
@@ -73,6 +72,7 @@
     "@types/estree": "^0.0.39",
     "@types/invariant": "^2.2.29",
     "@types/jest": "^22.2.3",
+    "@types/lodash": "^4.14.110",
     "@types/lz-string": "^1.3.32",
     "@types/node": "^9.6.5",
     "@types/pixi.js": "^4.7.3",
--- a/package.json
+++ b/package.json
@@ -29,11 +29,13 @@
     }
   },
   "dependencies": {
+    "@types/lodash": "^4.14.110",
     "acorn": "^5.7.1",
     "astring": "^1.3.0",
     "common-tags": "^1.7.2",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -29,11 +29,13 @@
     }
   },
   "dependencies": {
+    "@types/lodash": "^4.14.110",
```
ningyuansg: Should be a dev dependency.

src/localStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const loadStoredState = (): ISavedState | undefined => {
+  try {
+    const serializedState = localStorage.getItem('storedState')
+    if (serializedState === null) {
+      return undefined
+    } else {
+      // tslint:disable-next-line
+      console.log('Loading State: \n' + serializedState)
```
ningyuansg: To remove

src/localStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const saveState = (state: IState) => {
+  try {
+    const stateToBeSaved: ISavedState = {
+      token: state.session.token,
+      username: state.session.username,
+      historyHelper: state.session.historyHelper
+    }
+    // tslint:disable-next-line
+    console.log('Saving State: \n' + JSON.stringify(stateToBeSaved))
```
ningyuansg: To remove

src/localStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+    const serialized = JSON.stringify(stateToBeSaved)
+    localStorage.setItem('storedState', serialized)
+  } catch (err) {
+    // TODO catch possible errors
```
ningyuansg: Can we add an issue for this?

PR #148, opened 2018-07-02
Add a Grading listing Interface

remo5000:
### Features
- A table to view grading information (overviews)
- Table can be filtered and resized
- ContentDisplay has an optional `fullWidth` flag

### TODO
- Custom table filters (numeric and boolean logic)
- Hide IDs


ningyuansg:
I'm getting

```
./node_modules/ag-grid-react/lib/agGridReact.js
Module not found: Can't resolve 'react-dom-factories' in '/home/ning/github/cadet-frontend/node_modules/ag-grid-react/lib'
```

ningyuansg:
Header and content alignment is broken for large browser widths.

2018-07-02-194032_1920x1080_scrot

remo5000:
Nice catch, I couldn't find a native solution to this, but found a CSS one. I've made it stick to the left, so it always aligns with the columns. Big displays will just cause a space on the right, which is nonconsequential

remo5000:
Rebased.

PR #149, opened 2018-07-03
Refactor Assessment files naming

remo5000:
### Features
- Assessment/index is now the listing, instead of Assessment/AssessmentListing. This is because everything is redirected to the listing page first, making it the "index".
- Assessment/AssessmentWorkspace is now the component that spawns a workspace. It used to be the index, but is not. 
- Renamed props and styles to reflect the above changes


remo5000:
Travis failed due to a renaming issue, should be fixed now.

PR #166, opened 2018-07-05
Remove throttle for editorValue, fix workspace bugs

### LINKED ISSUE #147: Delay of 1s in saving editor value causes accidental missing semicolon errors
- The editor value is saved every second
- The evaluation of the editor or repl takes the program from the editor value
- Therefore, there is a brief window of max 1s where
  - The user has keyed in some text in the editor/repl
  - This new text is not yet saved
  - The user evaluates the editor/repl (commonly with shift + enter)
  - The program being evaluated is therefore not reflective of the _current_ editor/repl values

This generally causes quite a lot of missing semicolon errors, since users tend to `shift + enter` as soon as they end off a line (with a semicolon `;`).

### LINKED ISSUE #154: Increase update rate for REPL editor value
Typing in something, then evaluating it, will cause the old text to be placed back into the text editor. This also causes buggy behaviour (invisible REPL value) after code execution. 

This behaviour happens because the EVAL_REPL action is called before the UPDATE_REPL_VALUE action (due to the throttle).

### LINKED ISSUE #163: The default playground value raises missing semi-colon error
Steps to replicate:

1. Load up fresh playground
2. Press run

The default value is a single line comment:

```javascript
// Type your code here!
```

Changing the editorValue removes this error, this is probably something related to that slice of the state, then.

### LINKED ISSUE #165: Share button is broken

### LINKED ISSUE #170: Chapter select is missing from playground

ningyuansg:
### Features

- Editor and repl value updates are no longer throttled. Fixes #154, fixes #147.
- Fix default playground value raising semi-colon error. Fixes #163.
- Fix share button not working. Fixes #165.
- Fix missing chapter select. Fixes #170.

### DIFF
--- a/src/components/Playground.tsx
+++ b/src/components/Playground.tsx
@@ -18,6 +18,7 @@ export interface IStateProps {
   editorWidth: string
   isRunning: boolean
   output: InterpreterOutput[]
+  queryString?: string
   replValue: string
   sideContentHeight?: number
 }
@@ -28,6 +29,7 @@ export interface IDispatchProps {
   handleEditorEval: () => void
   handleEditorValueChange: (val: string) => void
   handleEditorWidthChange: (widthChange: number) => void
+  handleGenerateLz: () => void
   handleInterruptEval: () => void
   handleReplEval: () => void
   handleReplOutputClear: () => void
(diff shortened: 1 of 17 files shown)


src/components/Playground.tsx
```diff
@@ -97,9 +101,7 @@ class Playground extends React.Component<IPlaygroundProps, PlaygroundState> {
   }
 
   private chooseEditorValue(props: IPlaygroundProps): string {
-    return parsePrgrm(this.props) || this.props.editorValue === undefined
-      ? defaultPlaygroundText
-      : this.props.editorValue
+    return parsePrgrm(this.props) || this.props.editorValue
```
remo5000: As of 20f72c82d07933717c9f7c39455f330a7c50817d, the playground cannot be edited after the playground link is accessed. If i recall, it is because of how an empty string acts as `false` in the `||` check. Can you try to revert this line to see if it fixes it?

PR #173, opened 2018-07-06
Fix workspace bugs

### LINKED ISSUE #153: Question ID higher than the maximum question ID crashes Workspace
This is expected by design, just that it would be nice to account for this bug (in case some power user types in the exact question they want to do). A negative number doesn't get parsed into a number, so that side of the cases is taken care of.

### LINKED ISSUE #164: Key property needed by MCQChooser
Just need to use the key parameter in `map()`

### LINKED ISSUE #167: Suppress or fix scrolling warning
```
Automatically scrolling cursor into view after selection change this will be disabled in the next version set editor.$blockScrolling = Infinity to disable this message
```
Appears due to auto-scroll. Not sure if we can edit the code responsible for this.

### LINKED ISSUE #168: REPL value missing when obstructed
#### Steps to reproduce
Method One:
- Type something into the REPL
- Move the slider down so that the REPL editor box is not completely visible
- Type into the REPL again

Method Two:
- Move the slider down so that the REPL editor box is not completely visible
- Hold shift-enter for 3 seconds to create a bunch of output boxes
- Move the slider up to a normal position
- Type into the REPL again

Two methods to hopefully pinpoint the source faster.

### LINKED ISSUE #171: Unnecessary component re-rendering
### Root cause
Having connected the components at a higher level (Assessment/Playground/Grading) instead of connecting them at a lower one like before, there is a problem of constant prop updates. Any single change causes an update of the entire component, and all its children. 

### Solution
Use `PureComponent` instead of `Component`, as it does a shallow check on the props to see if there needs to be an update. This is internally implemented using `shouldComponentUpdate`. We can always define our own method for this, but it is [recommended against](https://reactjs.org/docs/react-api.html#reactpurecomponent).

### LINKED ISSUE #172: Repl scrolls down when slider is moved
Possibly caused by #171

remo5000:
### Issues fixed
- Fixes #164
- Fixes #153
- Fixes #167 
- Fixes #172 
- Fixes #171
- Fixes #168

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "cadet-frontend",
-  "version": "0.1.2",
+  "version": "0.1.3",
   "scripts-info": {
     "format": "Format source code",
     "start": "Start the Webpack development server",
--- a/src/components/assessment/AssessmentWorkspace.tsx
+++ b/src/components/assessment/AssessmentWorkspace.tsx
@@ -84,9 +84,13 @@ class AssessmentWorkspace extends React.Component<
         </Card>
       </Dialog>
     )
-    const question: IQuestion = this.props.assessment.questions[this.props.questionId]
+    /* If questionId is out of bounds, set it to the max. */
+    const questionId = (this.props.questionId >= this.props.assessment.questions.length) 
+      ? this.props.assessment.questions.length - 1 
+      : this.props.questionId
+    const question: IQuestion = this.props.assessment.questions[questionId]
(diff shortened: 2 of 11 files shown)


ningyuansg:
<div align='center'>
  
</div>

The `overflow: scroll` property for `.pt-card` breaks rendering of the `pt-card` on my browser. I removed this once in one of the earlier PRs, but I see that you need this for the markdown editor. Can you target the markdown editor more selectively, and we'll see if that works without breaking `.pt-cards` for me.

Debian X11 Chrome 67.

PR #186, opened 2018-07-10
Use js-slang instead of slang files

remo5000:
### Features
- No more `src/slang` codebase 
- A `slangHelper.ts` file to define application-specific external builtins (current only display and prompt) and export a curried `createContext`.


remo5000:
My bad, fixing it [here](https://github.com/source-academy/js-slang/pull/10)

PR #191, opened 2018-07-11
Use context properties instead of state

### LINKED ISSUE #39: Implement interrupt/stop using context.runtime.isRunning?
Refer to review on #27.

remo5000:
### Features
- Uses Context's chapter property instead of a `sourceChapter` property in the state
- Moved assessmentId check into AssessmentWorkspace
- Uses Context's isRunning to check `isRunning`

### Issues fixed
Fixes #39 

### Directly following
- Make workspace reset accept a library parameter


src/reducers/workspaces.ts
```diff
@@ -205,17 +202,16 @@ export const reducer: Reducer<IWorkspaceManagerState> = (
...
+       * Called to force a state change, which
+       * will cause a re-render
+       */
       return {
-        ...state,
-        [location]: {
-          ...state[location],
-          isRunning: false
-        }
+        ...state
```
ningyuansg: As discussed, there is a chance that this fails due to its async nature.

interrupt(context) in the saga inserts an interrupt error into the context. The scheduler only stops on the next evaluation step when it comes across this interrupt error. Then, the scheduler stops execution and sets context.isRunning to false. END_INTERRUPT_EXECUTION depends on the scheduler having done all this before the reducer takes this action of recreating the state, so that the props has a new value for isRunning and therefore triggers a re-render of the eval/run/stop buttons.

Instead, make END_INTERRUPT_EXECUTION explicitly (and redundantly) set isRunning in the context to false, and also push an interrupt error into the context ([reference](https://github.com/source-academy/js-slang/blob/13b4e1adbfaed6f1f369e4bd21265867290c3f3b/src/index.ts#L62)).

src/actions/actionTypes.ts
```diff
@@ -15,7 +15,17 @@ export const GENERATE_LZ_STRING = 'GENERATE_LZ_STRING'
...
+/**
+ * Called to signal the beginning of an interruption,
+ * i.e the trigger for the interpreter to interrupt execution.
+ */
+export const BEGIN_INTERRUPT_EXECUTION = 'BEGIN_INTERRUPT_EXECUTION'
+/**
+ * Called to signal the end of an interruption,
+ * i.e called after the interpreter is told to stop interruption,
+ * to cause UI changes.
+ */
```
ningyuansg: Can we move these comments to the reducers? I think there are at least three places we can comment on the actions:

- actionTypes
- actionCreators
- reducers & saga

And I think we should standardise them being in the reducers and saga because

1. Most of them are already there, and
2. That's where the logic is

remo5000:
Thanks, changes made 😄

Issue #199, opened 2018-07-16
Improve UX for browsing of repl history


ningyuansg:
1. When browsing up or down, the location of the cursor should be set to the last row and column
    - Currently, the cursor remains on the same column, unless the new text navigated to is too short. Then, it will take the position at the end of that text
2. Another hotkey should be set to allow navigation up and down lines
    - Currently, the user has to use the left and right arrow keys

PR #200, opened 2018-07-16
Implement history search in repl via up/down arrows

### LINKED ISSUE #196: up and down arrows in REPL
Would be nice if the up arrow in REPL would take the user to previously entered program. After a series of up keys, the down key takes the user to more recently entered programs. This behaviour is common in shell tools, which are basically REPLs.

ningyuansg:
- Resolves #196 

Some possible improvements are documented in #199.


src/reducers/workspaces.ts
```diff
@@ -167,25 +251,24 @@ export const reducer: Reducer<IWorkspaceManagerState> = (
...
+      } else {
+        newReplHistoryRecords = state[location].replHistory.records
       }
-    case EVAL_EDITOR:
-      return {
-        ...state,
-        [location]: {
-          ...state[location]
-        }
+      if (newReplHistoryRecords.length > 50) {
```
remo5000: You can consider using a deque or a circular buffer, so that the size does not have to be a concern.

src/reducers/workspaces.ts
```diff
@@ -167,25 +251,24 @@ export const reducer: Reducer<IWorkspaceManagerState> = (
...
-          ...state[location],
-          output: newOutput
-        }
+      let newReplHistoryRecords: string[]
+      if (action.payload.value !== '') {
+        newReplHistoryRecords = [action.payload.value].concat(state[location].replHistory.records)
+      } else {
+        newReplHistoryRecords = state[location].replHistory.records
       }
-    case EVAL_EDITOR:
```
remo5000: I should have added some documentation for this, my bad. It is here to force a re-render in the workspace component, showing the "Stop" button. Can it be added back? The Stop button does not show up when running e.g `const fib = (x) => x <= 2? 1 : fib(x-1) + fib(x-2); fib(20);`

ningyuansg:
Changes made with c6b0fdc and d4eb7ae.

PR #204, opened 2018-07-16
Fetch assessments

ningyuansg:
### Features

- AssessmentOverviews are now fetched from the backend
- Assessments are now fetched from the backend

Currently raises a runtime error when rendering a mcq question due to incompatible response model from cadet (backend). See https://github.com/source-academy/cadet/issues/136. Should be ok once it's fixed there.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,4 @@
 REACT_APP_IVLE_KEY=your_ivle_key_here
 REACT_APP_VERSION=$npm_package_version
 REACT_APP_BACKEND_URL=http://something.com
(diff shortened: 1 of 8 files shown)


ningyuansg:
@remo5000 Bump

ningyuansg:
There is a new .env variable defined in the .env.example file, `REACT_APP_USE_BACKEND`. If it is empty, the mock assessments will be used. Otherwise, the saga will make actual backend calls to the given `REACT_APP_BACKEND_URL`.

PR #209, opened 2018-07-17
Create slice of state for Grading

### LINKED ISSUE #205: Stop button is not rendered for when workspace isRunning
The commit to master resulting in this bug is 7a2c087.

remo5000:
### Features
- Far cleaner management of actions and state (no need for edge cases when resetting, only more specific cases
- Grading does not care about the `assessment` slice of the state, refers to `grading` only
- Use of `workspaceLocation` in containers instead of `location`

Fixes #205.


src/reducers/workspaces.ts
```diff
@@ -229,28 +263,21 @@ export const reducer: Reducer<IWorkspaceManagerState> = (
...
-          replHistory: {
-            ...state[location].replHistory,
-            records: newReplHistoryRecords
-          }
+          output: newOutput
         }
       }
     case EVAL_EDITOR:
-      // Forces re-render of workspace on editor eval
+      return {
```
ningyuansg: Can we preserve the comment `// Forces re-render of...`? I'm 100% gonna forget what this does in the future and end up deleting it again.

remo5000: My bad, got removed in a rebase. I intended to keep it too

PR #210, opened 2018-07-17
Add sound and runes support

remo5000:
### Features
- All libraries loaded using one index script (therefore, only one `<script>` reference
- Added runes externals 
- Added a `CanvasOutput` (from sourceacademy2) to allow output of 512x512 canvases
- Basic sound and rune functionalities work

### DIFF
--- a/src/components/workspace/CanvasOutput.tsx
+++ b/src/components/workspace/CanvasOutput.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react'
 
 /**
- * Takes the output of the rendered graphics (in a hidden canvas tag under <body>) 
+ * Takes the output of the rendered graphics (in a hidden canvas tag under <body>)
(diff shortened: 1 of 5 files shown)


src/components/workspace/CanvasOutput.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react'
+
+/**
+ * Takes the output of the rendered graphics (in a hidden canvas tag under <body>)
+ * and makes it into a new <canvas> output for viewing.
+ */
+class CanvasOutput extends React.Component<{}, {}> {
+  private $canvas: HTMLCanvasElement | null
+  private $parent: HTMLElement | null
```
ningyuansg: Are `$variables` a convention for variables pointing to HTMLElements? If so I'll open an issue changing the other HTMLElement variable names.

PR #215, opened 2018-07-18
Load editorValue from solutionTemplate or answer

ningyuansg:
This branch is based off #204. After that is merged, it will be rebased off master and be cleaner (5 commits).

### Features

- The default editor value is undefined for assessment workspaces.
- If an assessment workspace has an editor value, load that editor value
    - This represents a user navigating back into the assessment workspace after having briefly navigated away e.g. to the playground, but not to other assessment questions
- Otherwise, if the question being loaded has an answer, load that answer
    - This represents a user who has just navigated into a question that they have saved an answer for in the past. The saved answer was sent into the backend, and is now provided to us via the backend api as the property answer.
- Otherwise, load the solution template
    - This represents a user who has just started a question for the first time

Note that all of these features apply only for programming questions. The mcq questions need fixes w.r.t. response bodies.

### Todos

- [x] Saving an answer should also change the answer property of the IAssessment in our local state. Right now it only does the saving via an api call to the backend. The relevant action is saga-handled SUBMIT_ANSWER.


ningyuansg:
@remo5000 Rebased to master

ningyuansg:
@remo5000 Fixed the issue where evaluating an assessment workspace that has just been mounted always results in a missing semicolon syntax error, because the editor value in the state which eval editors uses still has the default value of undefined. 31a84c2

src/reducers/states.ts
```diff
@@ -58,7 +58,7 @@ export interface IWorkspaceManagerState {
 
 interface IWorkspaceState {
   readonly context: Context
-  readonly editorValue: string
+  readonly editorValue: string | undefined
```
remo5000: Can this be changed to `null`?

src/utils/notification.ts
```diff
@@ -4,15 +4,15 @@ const notification = Toaster.create({
...
-export const showSuccessMessage = (message: string, timeout = 500) => {
+export const showSuccessMessage = (message: string, timeout: number) => {
   notification.show({
     intent: Intent.SUCCESS,
     message,
     timeout
   })
 }
 
-export const showWarningMessage = (message: string, timeout = 500) => {
```
remo5000: As discussed, could you try `timeout: number = 500`? If it doesn't work, then keeping the current change is fine too

PR #219, opened 2018-07-20
Update shape of Library

remo5000:
### Features
- `globals` are used properly (see [issue in cadet](https://github.com/source-academy/cadet/issues/149))
- `files` property in `Library` has been removed  
- Reverted component updating for AssessmentWorkspace and GradingWorkspace


ningyuansg:
Not merging until backend implements #149, since the site would probably crash otherwise.

ningyuansg:
Note to self: do #207 once this is merged

PR #225, opened 2018-07-23
Add Favicon

remo5000:
### Features
- image

### TODO
- Should we keep a rounded black background?
- Should it be bigger?
- Should we use the yellow version instead of this one?


martin-henz:
"Cadet" should not appear anywhere as the name of the system. It should be "Source Academy". We should also not let things like "frontend" or "backend" spill over into the user namespace.

ningyuansg:
Actually that's pretty trivial so I'll go ahead and make the changes

PR #228, opened 2018-07-23
Submit and show MCQ Answer

remo5000:
### Features
- MCQs can be selected
- Selection will cause an update using the already defined API call `submitAnswer`
  - Save button does not show up for MCQ 
- Selecting MCQ in the `GradingWorkspace` does not cause anything to happen 
- `resetWorkspace` action creator generalised, allows an argument to specify values to be set in the new workspace

### DIFF
--- a/src/components/assessment/AssessmentWorkspace.tsx
+++ b/src/components/assessment/AssessmentWorkspace.tsx
@@ -122,6 +122,10 @@ class AssessmentWorkspace extends React.Component<
           ? ((question as IProgrammingQuestion).answer as string)
           : (question as IProgrammingQuestion).solutionTemplate
         : null
+    const chosenMCQAnswer = 
+      question.type === QuestionTypes.mcq && question.answer !== null
+        ? ((question as IMCQQuestion).answer as number)
+        : null
     const workspaceProps: WorkspaceProps = {
       controlBarProps: this.controlBarProps(this.props, questionId),
       editorProps:
@@ -136,6 +140,7 @@ class AssessmentWorkspace extends React.Component<
       handleEditorWidthChange: this.props.handleEditorWidthChange,
       handleSideContentHeightChange: this.props.handleSideContentHeightChange,
       mcqProps: {
+        chosenOption: chosenMCQAnswer,
         mcq: question as IMCQQuestion,
         handleMCQSubmit: (option: number) =>
           this.props.handleSave(this.props.assessment!.questions[questionId].id, option)
--- a/src/components/assessment/AssessmentWorkspace.tsx
+++ b/src/components/assessment/AssessmentWorkspace.tsx
@@ -124,7 +124,7 @@ class AssessmentWorkspace extends React.Component<
         : null
     const chosenMCQAnswer =
       question.type === QuestionTypes.mcq && question.answer !== null
-        ? (question as IMCQQuestion).answer as number
+        ? ((question as IMCQQuestion).answer as number)
         : null
     const workspaceProps: WorkspaceProps = {
       controlBarProps: this.controlBarProps(this.props, questionId),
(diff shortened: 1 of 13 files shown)


remo5000:
I use equality to check if a button should be "pressed" (Is there a better way around this?). As such, all the checks returned false due to the http request returning `answer` as a string. I've opened up an issue [here](https://github.com/source-academy/cadet/issues/162), and added a workaround (and TODO) in the saga

src/components/workspace/MCQChooser.tsx
```diff
@@ -33,6 +48,22 @@ class MCQChooser extends React.PureComponent<IMCQChooserProps, {}> {
...
+  /**
+   * A function to generate an onClick function that causes
+   * and mcq submission with a given answer id.
+   *
+   * Post-condition: the local state will be updated to store the
+   * mcq option selected.
+   *
+   * @param i the id of the answer
+   */
+  private onButtonClickFactory = (i: number) => (e: any) => {
```
remo5000: Should I stick to this or make a tslint exception?

src/components/workspace/MCQChooser.tsx
```diff
@@ -4,15 +4,30 @@ import * as React from 'react'
...
 
 export interface IMCQChooserProps {
+  chosenOption: number | null
   mcq: IMCQQuestion
-  mcqSubmit?: (choiceId: number) => void
+  handleMCQSubmit: (choiceId: number) => void
 }
 
-class MCQChooser extends React.PureComponent<IMCQChooserProps, {}> {
+type State = {
```
remo5000: Local state preserves the mcq option, that might seem to 'flutter' back to its original value in between saving and the state being updated.

src/components/assessment/AssessmentWorkspace.tsx
```diff
@@ -122,6 +122,10 @@ class AssessmentWorkspace extends React.Component<
           ? ((question as IProgrammingQuestion).answer as string)
           : (question as IProgrammingQuestion).solutionTemplate
         : null
+    const chosenMCQAnswer =
+      question.type === QuestionTypes.mcq && question.answer !== null
+        ? ((question as IMCQQuestion).answer as number)
+        : null
```
ningyuansg: Does this work

```typescript
const chosenMCQAnswer = (question as IMCQQuestion).answer
  ? (question as IMCQQuestion).answer
  : null
```

If it does can you also change the `const editorValue = ...` below similarly?

Don't think we need to cast it as a number. Should just wait for backend to fix.

src/components/workspace/MCQChooser.tsx
ningyuansg: I think it's ok

src/sagas/backend.ts
```diff
@@ -34,7 +38,14 @@ function* backendSaga(): SagaIterator {
   yield takeEvery(actionTypes.FETCH_ASSESSMENT, function*(action) {
     const accessToken = yield select((state: IState) => state.session.accessToken)
     const id = (action as actionTypes.IAction).payload
-    const assessment = yield call(getAssessment, id, accessToken)
+    const assessment: IAssessment = yield call(getAssessment, id, accessToken)
+    // TODO remove parsing once backend/#162 is resolved.
```
ningyuansg: Open an issue?

src/components/assessment/AssessmentWorkspace.tsx
remo5000: Actually, `(question as IMCQQuestion).answer as number || null` achieves both the typing and null checking. The same cannot be done for the editorValue as the question type must be checked. We could simplify the expression but it will be harder to read IMO.

PR #232, opened 2018-07-24
Implement Logout

### LINKED ISSUE #132: Add a logout feature
In addition to removing the token from the front-end, the Refresh token should be sent to the server upon logout

### LINKED ISSUE #145: Logging in resets playground editor value to default
The text in the editor of the playground does not persist after a log in. Instead, the text in the editor reverts to its default value.

**Expected bahaviour**: The text in the editor of the playground is the same before and after logging in

**Steps to replicate**:

1. Navigate to the playground
2. Edit the text in the editor
3. Navigate to the academy (mounts the log-in component)
4. Log-in via IVLE (redirects to academy)
5. Navigate to the playground
6. The text in (2) has now disappeared, and the editor is instead showing the default text.

ningyuansg:
### Features

- There is a logout button and a logout dispatch (for me to use at #230 for expired refresh tokens)
- Playground is persisted in local storage

Resolves #132, resolves #145.

### DIFF
--- a/src/components/academy/Status.tsx
+++ b/src/components/academy/Status.tsx
@@ -34,7 +34,10 @@ const StatusPopover = (props: StatusProps) => (
       <Text>
         <h4>{`Source Academy, ${titleCase(props.role)}`}</h4>
       </Text>
-      {controlButton('', IconNames.LOG_OUT, props.handleLogOut)}
+      {controlButton('Log Out', IconNames.LOG_OUT, props.handleLogOut, {
+        minimal: false,
+        iconOnRight: true
+      })}
     </div>
   </Popover>
 )
--- a/src/components/academy/Status.tsx
+++ b/src/components/academy/Status.tsx
@@ -8,6 +8,7 @@ import { controlButton } from '../commons'
 type StatusProps = OwnProps
 
 type OwnProps = {
+  handleLogOut: () => void
   username: string
   role: Role
 }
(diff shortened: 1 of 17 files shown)


src/components/academy/Status.tsx
```diff
@@ -20,15 +21,25 @@ const Status: React.SFC<StatusProps> = props => (
...
     </div>
-    <Popover popoverClassName="Popover-share pt-dark" inheritDarkTheme={true}>
-      <div className="navbar-button-text hidden-xs">
-        {controlButton(titleCase(props.username), IconNames.USER)}
-      </div>
+    {StatusPopover(props)}
+  </>
+)
+
+const StatusPopover = (props: StatusProps) => (
```
remo5000: Can this be lowercase

src/createStore.ts
```diff
@@ -41,7 +41,13 @@ function createStore(history: History): Store<IState> {
           ...defaultState,
           session: {
             ...defaultState.session,
-            ...loadedStore
+            ...loadedStore.session
+          },
+          workspaces: {
+            ...defaultState.workspaces,
+            playground: loadedStore.playgroundWorkspace
```
remo5000: Do you think we should use the spread instead (`{...defaultPlaygroundWorkspace, ...loadedStore.playgroundWorkspace}`)? This will allow the stored playgroundWorkspace to be Partial as well. But perhaps it might be easier to work with the exclusive decision, especially when debugging

src/components/academy/Status.tsx
ningyuansg: Fixed with 6a916e4.

ningyuansg:
No, I don't think the runtime error on the site when there are source errors in the context is acceptable. I'll @ you when I fix that.

src/createStore.ts
ningyuansg: Fixed with 2f0e5ca.

PR #235, opened 2018-07-25
Allow external libraries to be shared

### LINKED ISSUE #207: Implement link sharing for externals

remo5000:
### Features
- Playground links have a `ext` component that stores the external library name to be used
- Application is now a `Component` so that the `parsePlayground` function can be called outside `render()`
- Refactored actions that rely on internal `ControlProps` types (e.g `IExternal` and `IChapter`) by making arguments specific

### TODO
- [x] Implement race condition and timeout for libraries being loaded onto the page
- [x] Rebase and target master 

### Issues fixed
- Fixes #207


src/sagas/index.ts
```diff
@@ -140,23 +140,65 @@ function* workspaceSaga(): SagaIterator {
...
+        }
+        /** Prevent checking instantaenously by waiting for a bit. */
+        yield call(delay, 250)
+      }
+      return true
+    }
+    /** Create a race condition between the js files being loaded and a 3 second timeout. */
+    const { loadedScripts, timeout } = yield race({
+      loadedScripts: call(helper),
+      timeout: call(delay, 4000)
```
ningyuansg: The comment above should say 4s? Anyways I think the comments from lines 143 to here are a little unnecessary. Instead maybe comment (and not necessarily here),

- where does getReadyWebGLForCanvas come from?
- why is there a need to check if it's available (runtime errors)?

src/sagas/index.ts
```diff
@@ -140,23 +140,65 @@ function* workspaceSaga(): SagaIterator {
...
+    yield all([call(checkWebGLAvailable)])
+  })
+
   /**
    * Handles the side effect of resetting the WebGL context when context is reset.
    *
    * @see clearContext and files under 'public/externalLibs/graphics'
    */
   yield takeEvery(actionTypes.CLEAR_CONTEXT, function*(action) {
+    yield all([call(checkWebGLAvailable)])
```
ningyuansg: Is `all` here to ensure `checkWebGLAvailable` has finished running before this goes to the next line? Doesn't `call` already wait for that? If not, does `yield*` work?

remo5000: Yep, I couldn't get `call` to be blocking but I'll try `yield*`

ningyuansg:
@remo5000 Can you update https://github.com/source-academy/sicp/issues/5 with the externals sharing stuff?

PR #238, opened 2018-07-26
Add Grading API calls

remo5000:
Awaiting https://github.com/source-academy/cadet/issues/167.

### Features
- Add sagas to fetch Grading from backend
- Use XP instead of grade
- Remove `graded` option (to be added once backend endpoint is available)

### TODO
- [x] Add GradingOverview calls and type changes
- [x] Add Grading calls and type changes
- [x] Test with an updated staging server
  - [x] Adjustments in grading overview (awating backend)
  - [x] Assessment name and category


ningyuansg:
@remo5000 You gotta rebase off #240 before you use API call in `./src/sagas/backend.ts` because the function `request` has been changed completely in that PR.

remo5000:
@ningyuansg will do

src/components/academy/grading/GradingEditor.tsx
```diff
@@ -1,36 +1,42 @@
...
-  handleGradingCommentsChange: (s: string) => void
-  handleGradingXPChange: (i: number | undefined) => void
-  handleGradingInputSave: (s: string, i: number | undefined) => void
+  handleGradingSave: (
+    submissionId: number,
+    questionId: number,
+    g: number,
+    s: string,
+    i: number | undefined
+  ) => void
```
ningyuansg: Full param names (since prettier put them on separate lines already). I think abbreviated param names is acceptable just for stuff like the tests.

src/components/academy/grading/GradingEditor.tsx
```diff
@@ -1,36 +1,42 @@
...
- * as well as the XP in the numeric input.
+ * as well as the grade adjustment in the numeric input.
  *
- * XP can be undefined to show the hint text.
+ * adjustment can be undefined to show the hint text.
  */
 type State = {
   mdeState: ReactMdeTypes.MdeState
-  XPInput: number | undefined
+  adjustmentInput: number | undefined
```
ningyuansg: Can this be `number | null`, like that time with the editorValue

src/containers/academy/grading/GradingEditorContainer.ts
```diff
@@ -1,26 +1,16 @@
...
+import GradingEditor, { DispatchProps } from '../../../components/academy/grading/GradingEditor'
 import { IState } from '../../../reducers/states'
 
-const mapStateToProps: MapStateToProps<StateProps, {}, IState> = state => {
-  return {
-    gradingCommentsValue: state.workspaces.grading.gradingCommentsValue,
-    gradingXP: state.workspaces.grading.gradingXP
-  }
-}
+const mapStateToProps: MapStateToProps<{}, {}, IState> = state => ({})
```
ningyuansg: You can pass `null` to `connect` in lieu of an empty `mapStateToProps`.

src/sagas/backend.ts
```diff
@@ -136,6 +143,79 @@ function* backendSaga(): SagaIterator {
...
+      })
+      yield put(actions.updateGrading(submissionId, newGrading))
+    } else if (resp !== null) {
+      let errorMessage: string
+      switch (resp.status) {
+        case 400:
+          errorMessage = 'Invalid or missing parameter(s) or submission and/or question not found'
+          break
+        case 401:
+          errorMessage = 'Got 403 response. Only staff can save gradings.'
```
ningyuansg: Typo here (should be '403'). Also, can you help me change [this line](https://github.com/source-academy/cadet-frontend/blob/c61fd6009f480bccc4ed06470a364e5ae9c98b4b/src/sagas/backend.ts#L130) to `case 401`?

Anyways I think this case clause can be omitted. Anyone using the site properly would be a staff by the point they reach here, so it doesn't make sense to remind them that only a staff can save gradings. This should fall in the default clause of 'oops we don't know what happened, here's the error code, go open an issue please'.

src/sagas/backend.ts
```diff
@@ -266,6 +346,104 @@ async function postAnswer(
...
+ * GET /grading
+ * @returns {Array} GradingOverview[]
+ */
+async function getGradingOverviews(tokens: Tokens): Promise<GradingOverview[] | null> {
+  const response = await request('grading', 'GET', {
+    accessToken: tokens.accessToken,
+    refreshToken: tokens.refreshToken,
+    shouldRefresh: true
+  })
+  if (response && response.ok) {
```
ningyuansg: `if (response)` is sufficient. Non-null response implies `response.ok`. [Reference](https://github.com/source-academy/cadet-frontend/blob/c61fd6009f480bccc4ed06470a364e5ae9c98b4b/src/sagas/backend.ts#L482).

src/sagas/backend.ts
```diff
@@ -266,6 +346,104 @@ async function postAnswer(
...
+ * GET /grading/${submissionId}
+ * @returns {Grading}
+ */
+async function getGrading(submissionId: number, tokens: Tokens): Promise<Grading | null> {
+  const response = await request(`grading/${submissionId}`, 'GET', {
+    accessToken: tokens.accessToken,
+    refreshToken: tokens.refreshToken,
+    shouldRefresh: true
+  })
+  if (response && response.ok) {
```
ningyuansg: `if (response)`

src/utils/castBackend.ts
```diff
@@ -0,0 +1,15 @@
+import { ExternalLibraryName, Library } from '../components/assessment/assessmentShape'
+
+export const castLibrary = (lib: any): Library => ({
```
ningyuansg: Good to document differences in frontend and backend data, like [here](https://github.com/source-academy/cadet-frontend/blob/c61fd6009f480bccc4ed06470a364e5ae9c98b4b/src/sagas/backend.ts#L308).

remo5000:
There was a problem with typing, as the input as a number was stored. I've stored the input as a string,  and parse it whenever this data leaves the component (submission)

src/sagas/backend.ts
```diff
@@ -136,6 +143,79 @@ function* backendSaga(): SagaIterator {
...
+      grade,
+      comment,
+      adjustment
+    } = (action as actionTypes.IAction).payload
+    const tokens = yield select((state: IState) => ({
+      accessToken: state.session.accessToken,
+      refreshToken: state.session.refreshToken
+    }))
+    const resp = yield postGrading(submissionId, questionId, grade, comment, adjustment, tokens)
+    if (resp !== null && resp.ok) {
```
ningyuansg: Use `resp && resp.ok` for consistency

src/sagas/backend.ts
```diff
@@ -166,6 +173,79 @@ function* backendSaga(): SagaIterator {
...
+          }
+        }
+        return gradingQuestion
+      })
+      yield put(actions.updateGrading(submissionId, newGrading))
+    } else if (resp !== null) {
+      let errorMessage: string
+      switch (resp.status) {
+        case 400:
+          errorMessage = 'Invalid or missing parameter(s) or submission and/or question not found'
```
ningyuansg: These error messages are for the end user. If they're not providing useful end-user-type information I think just let them be caught in the default case.

src/sagas/backend.ts
```diff
@@ -166,6 +173,79 @@ function* backendSaga(): SagaIterator {
...
+          break
+        case 403:
+          errorMessage = 'Got 403 response. Only staff can save gradings.'
```
ningyuansg: Check role before sending the POST instead of based on the POST response (see SUBMIT_ANSWER)

remo5000: Sure, but I think it's good to note that in a normal case the student would not be able to reach the UI to submit the grading anyway

remo5000: Updated the codes to match the new ones too

PR #250, opened 2018-07-31
Generate story with information from backend

ningyuansg:
Awaiting https://github.com/source-academy/cadet/issues/179.


ningyuansg:
@remo5000 I'm using this PR to bump v0.1.7 to v0.1.8, so we need to merge #266 first if you want to include it in the next staging bump.

PR #263, opened 2018-08-03
Redesign MCQ choosing

remo5000:
### Features
- Only MCQ option are on the MCQChooser component
- MCQ Questions are differentiated based on the presence of a `solution` (and hints):
  - Buttons have intent when the question has a solution
  - Hints are displayed as notifications when options are chosen
  - For ungraded questions, there is no intent on the buttons and there is no notification displayed.

PR #268, opened 2018-08-06
Add List Visualiser

remo5000:
### Requirements
- [x] js-slang must specify the `draw` function to be available (not exactly necessary, but is essential for testing)

### Features
- calling `draw` on a list visualises it in a side content tab
- The visualiser is only available on the playground
- The "Next", "Previous" and "Clear" buttons have been scrapped, as that introduces a two-way interaction between the application (that calls the external functions) and the library (that manipulates the dom and refers to the buttons by class and id). Either ways, I don't see the use case for visualising multiple lists in quick succession.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "draft-js": "^0.10.5",
     "flexboxgrid": "^6.3.1",
     "flexboxgrid-helpers": "^1.1.3",
-    "js-slang": "0.1.3",
+    "js-slang": "0.1.5",
     "lodash": "^4.17.10",
     "lz-string": "^1.4.4",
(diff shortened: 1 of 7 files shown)


ningyuansg:
Are changes to MCQ meant to be part of this PR or included by accident?

remo5000:
It's by accident, those changes should have showed up from #263

remo5000:
Rebased to the squash-commit of #263

ningyuansg:
Please resolve https://github.com/source-academy/js-slang/pull/27 and publish before this branch is merged.

remo5000:
@ningyuansg Bumped!

PR #277, opened 2018-08-09
Fix dates not being parsed properly

### LINKED ISSUE #270: opendate not parsed properly

remo5000:
### Features
- momentJS is used to parse Dates
  - The key problem is that although the `Date` object works, there is some inconsistency in how browsers use them. It also is erroneous in some cases, but I'm not too familiar with this particular date format to verify.

### Issues fixed
Fixes #270 

### TODO
- [x] Specify the ISO 8601 requirement in documentation

### DIFF
--- a/src/components/assessment/__tests__/index.tsx
+++ b/src/components/assessment/__tests__/index.tsx
@@ -31,6 +31,12 @@ const mockPresentAssessment: IAssessmentProps = {
   assessmentOverviews: mockAssessmentOverviews
 }
 
+const mockPresentAssessmentForStudent: IAssessmentProps = {
+  ...defaultProps,
+  assessmentOverviews: mockAssessmentOverviews,
+  isStudent: true
+}
+
 test('Assessment page "loading" content renders correctly', () => {
   const app = (
     <MemoryRouter initialEntries={['/unknown']}>
@@ -55,6 +61,16 @@ test('Assessment page with multiple loaded missions renders correctly', () => {
   const app = (
     <MemoryRouter initialEntries={['/unknown']}>
       <Assessment {...mockPresentAssessment} />
+    </MemoryRouter>
+  )
+  const tree = mount(app)
+  expect(tree.debug()).toMatchSnapshot()
+})
+
+test('Assessment page does not show attempt Button for upcoming assessments for student user', () => {
+  const app = (
+    <MemoryRouter initialEntries={['/unknown']}>
+      <Assessment {...mockPresentAssessmentForStudent} />
     </MemoryRouter>
   )
   const tree = mount(app)
--- a/src/components/assessment/__tests__/index.tsx
+++ b/src/components/assessment/__tests__/index.tsx
@@ -12,6 +12,7 @@ const defaultProps: IAssessmentProps = {
   assessmentOverviews: undefined,
   handleAssessmentOverviewFetch: () => {},
   handleSubmitAssessment: (id: number) => {},
+  isStudent: false,
   ...mockRouterProps('/academy/missions', {})
 }
 
--- a/src/components/assessment/index.tsx
+++ b/src/components/assessment/index.tsx
@@ -115,7 +115,9 @@ class Assessment extends React.Component<IAssessmentProps, State> {
 
       /** Opened assessments, that are released and can be attempted. */
       const isOverviewOpened = (overview: IAssessmentOverview) =>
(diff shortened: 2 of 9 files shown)


indocomsoft:
Working for the close-dates.

src/utils/dateHelpers.ts
```diff
@@ -1,28 +1,28 @@
...
+ * @param {string} d an ISO 8601 compliant date string
  *   e.g 2018-07-06T10:20:09.961Z
  * @returns {string} A user-friendly readable date string,
  *   e.g 07/06, 20:09
  */
 export const getPrettyDate = (d: string): string => {
-  const date = new Date(d)
-  return `${date.getDate()}/${date.getMonth()}, ${date.getHours()}:${date.getMinutes()}`
+  const date = moment(d)
+  return date.format('DD/M, HH:mm')
```
ningyuansg: Can we have something more human readable?

src/utils/dateHelpers.ts
```diff
@@ -1,28 +1,28 @@
...
+ * @param {string} d an ISO 8601 compliant date string
  *   e.g 2018-07-06T10:20:09.961Z
  * @returns {boolean} true if the date specified by the paramter
  *   is before the time of execution of this function.
  */
 export const beforeNow = (d: string): boolean => {
-  const date = new Date(d)
-  const now = new Date()
-  return date <= now
+  return moment(d).isBefore(moment())
```
ningyuansg: I don't think we were aware of this earlier, but backend sends us overviews even for all assessments, even those that are not yet open.

For these 'pending' assessment overviews, we need to place them in their own collapsible category, and hide their 'attempt' buttons (but we should not hide the attempt buttons for staff and admins https://github.com/source-academy/cadet/pull/195). Can you implement this change in this PR?

remo5000:
@ningyuansg implemented it, please take a look as I'm not completely of the semantics of showing the cards

src/components/assessment/index.tsx
```diff
@@ -132,7 +166,8 @@ class Assessment extends React.Component<IAssessmentProps, State> {
         ) : null
       display = (
         <>
-          {openCardsCollapsible}
+          {unopenedCardsCollapsible}
```
ningyuansg: Would prefer 'upcoming', or 'pending'.

src/utils/dateHelpers.ts
```diff
@@ -22,7 +24,8 @@ export const beforeNow = (d: string): boolean => {
  * @returns {string} A user-friendly readable date string,
  *   e.g 07/06, 20:09
  */
-export const getPrettyDate = (d: string): string => {
-  const date = moment(d)
-  return date.format('DD/M, HH:mm')
+export const getPrettyDate = (dateString: string): string => {
+  const date = moment(dateString)
+  const prettyDate = date.format('DD/M, HH:mm')
```
ningyuansg: With 07/06 it is not immediately obvious if we're talking 7th June or 6th July. Can we have the full month name?

src/components/assessment/__tests__/index.tsx
```diff
@@ -12,6 +12,7 @@ const defaultProps: IAssessmentProps = {
   assessmentOverviews: undefined,
   handleAssessmentOverviewFetch: () => {},
   handleSubmitAssessment: (id: number) => {},
+  isStudent: false,
```
ningyuansg: Should write a test for both `isStudent` values, and make sure that the button does not render for students.

src/components/assessment/index.tsx
ningyuansg: Even 'future' is ok

src/utils/dateHelpers.ts
remo5000: Oh, that's what you meant!

src/components/assessment/__tests__/index.tsx
remo5000: Yep, sounds good

remo5000: Default is `false`, so I added a test for when the user is a student

src/utils/dateHelpers.ts
remo5000: Now it'll show "7th June, HH:MM". I think HHMM is good because there is less ambiguity regarding AM/PM

src/components/assessment/index.tsx
remo5000: Used upcoming

remo5000:
@ningyuansg Do let me know if there are any further changes

PR #282, opened 2018-08-10
View comments

### LINKED ISSUE #267: Use markdown to render textual content
The markdown renderer already exists, but is used to render only the `GradingEditor` text output. We can move this to utils or commons and make use of it in other places as well.

remo5000:
### Features
- `IQuestion` now has a `comment` field
- Comments are displayed if the comments passed to the front-end are non-null

### TODO
- [x] Rename `comments` to `comment`
- [x] Use Markdown (from the PR that would resolve #267)

### DIFF
--- a/src/components/workspace/side-content/index.tsx
+++ b/src/components/workspace/side-content/index.tsx
@@ -35,7 +35,7 @@ class SideContent extends React.PureComponent<SideContentProps, {}> {
       const click = (i: number) => () => this.props.handleChangeActiveTab(i)
       const buttons = this.props.tabs.map((tab, i) => (
         <Tooltip key={i} content={tab.label}>
-          <Button icon={tab.icon as IconName} className="pt-minimal" onClick={click(i)} />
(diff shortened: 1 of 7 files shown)


remo5000:
As discussed, we'll open up an issue to make the comments tab the default tab.

src/components/workspace/side-content/index.tsx
```diff
@@ -18,31 +18,28 @@ class SideContent extends React.PureComponent<SideContentProps, {}> {
...
+          >
+            {this.props.tabs.map((tab, key) => (
+              <Tab
+                id={key}
+                key={key}
+                title={tab.label}
+                panel={<div className="side-content-text">{tab.body}</div>}
+              />
+            ))}
+          </Tabs>
```
ningyuansg: Can you revert the commits making this a blueprintjs tab? I meant only to change the implementation, without changing the look, i.e.

- Different tabs are still indicated by icons
- The icons are still centered
- There is still a visual divide between icons and side content content. It's just an awkward empty space right now.

PR #292, opened 2018-08-12
Change some game redirects to LumiNUS, Piazza

### LINKED ISSUE #287: Fix 404 for game redirects to /news and /material

ningyuansg:
Resolves #287.


src/utils/constants.ts
```diff
@@ -6,3 +6,13 @@ export const IVLE_KEY = process.env.REACT_APP_IVLE_KEY
 export const VERSION = process.env.REACT_APP_VERSION
 export const BACKEND_URL = process.env.REACT_APP_BACKEND_URL
 export const USE_BACKEND = process.env.REACT_APP_USE_BACKEND
+
+export const LINKS = {
```
remo5000: We don't use a type here, so enum would still work

PR #308, opened 2018-08-17
Fix News Link at Top Left

ReignOfComputer:
Fixes https://github.com/source-academy/cadet-frontend/issues/305


ningyuansg:
Thank you!

ningyuansg:
(Since this bug is not critical we will push this to prod with the next version)

PR #313, opened 2018-08-19
Disable MCQ selection in Grading workspace

remo5000:
### Features
- Buttons are clickable (to show hints), but the option does not get highlighted.
- An optional prop is used (like in ControlBar) as it is not an expected default behaviour


ningyuansg:
Good enhancement, but may as well use tooltips onhover to show hints, and just disable all the buttons.

remo5000:
Good idea, I'll add that in later.

ningyuansg:
Good comment, noted with thanks.

remo5000:
I just tried it out. I think the way `hint` is used makes it better off as a notification. For example,

image

We would want the student to see the explanation for why the answer is correct, only after clicking it.

ningyuansg:
I meant in the context of the grading workspace, that graders are able to view hints by hovering over the choice. No changes to student's view.

remo5000:
Don't think I have time to work on this for now, will reopen if I do though. The main pain point is creating a behaviour duality within the `MCQChooser`.

PR #338, opened 2018-08-26
Integrate color function from CP3108 project

thenaesh:
This PR integrates the color function written by @adamwth into the frontend. This will allow the rune contests to have the full spectrum of colors.


thenaesh:
@remo5000 The grader thing we'll need to KIV first, I don't know how the grader is going to deal with colours that vary very slightly. It may take a while and quite a lot of testing, so I can't do it so soon.

PR #348, opened 2018-09-03
Adjust xp linearly according to adjusted grade.

blackening:
### Features
image
- Automatic scaling from grade -> xp.
- Bump version to 1.0.9


blackening:
Ok, as requested, done. Was held up with yarn test. I suspect yarn test doesn't work on windows...

PR #363, opened 2018-09-07
Trivial Wording Fix for Introduction

ReignOfComputer:
Addresses https://github.com/source-academy/cadet-frontend/issues/362


remo5000:
@ReignOfComputer The travis build fails for some reason (I don't think it should, from looking at your changes)

PR #373, opened 2018-09-13
Add Tree library

lywjoel:


public/externalLibs/index.js
```diff
@@ -73,6 +75,13 @@ function loadLib(externalLibraryName) {
         '/externalLibs/graphics/webGLcurve.js',
       ];
       break;
+    case "TREE":
```
remo5000: The name of the library is used for distinguishing between clashing libraries like runes and graphics. I think loading it in the `loadAllLibs()` function should suffice as there seems to be no clash with other libraries. Could this case be removed?

lywjoel: done! won't i need the case name so i can load the library under the EXTERNAL tag though

remo5000: Nope, because it's going to be loaded regardless of what library is used. So all you need to do is specify the symbols, you can use `NONE` as the library name

PR #375, opened 2018-09-15
Grading overview upgrades

blackening:
Partially implements https://github.com/source-academy/cadet-frontend/issues/351

image

Defaults are as shown above.

For support to show whether a submission has already been graded, pending https://github.com/source-academy/cadet/issues/260.

### DIFF
--- a/src/components/academy/grading/index.tsx
+++ b/src/components/academy/grading/index.tsx
@@ -10,7 +10,7 @@ import { RouteComponentProps } from 'react-router'
 
 import GradingWorkspaceContainer from '../../../containers/academy/grading/GradingWorkspaceContainer'
 import { stringParamToInt } from '../../../utils/paramParseHelpers'
-import { AssessmentCategories } from '../../assessment/assessmentShape'
+import { AssessmentCategories, AssessmentStatuses } from '../../assessment/assessmentShape'
 import { controlButton } from '../../commons'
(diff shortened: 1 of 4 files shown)


src/components/academy/grading/index.tsx
```diff
@@ -25,6 +26,12 @@ type State = {
   columnDefs: ColDef[]
   filterValue: string
   groupFilterEnabled: boolean
+  missionFilter: boolean
```
rrtheonlyone: I feel its unnecessary to add so much clutter into the internal state for each filter. The whole reason behind using `ag-grd` instead of a pure table was to reduce this!

rrtheonlyone: I feel its unnecessary to add so much clutter into the internal state for each filter. The whole reason behind using `ag-grid` instead of a pure table was to reduce this!

src/components/academy/grading/index.tsx
```diff
@@ -181,6 +199,68 @@ class Grading extends React.Component<IGradingProps, State> {
               onChange={this.handleGroupsFilter}
             />
           </div>
+
```
rrtheonlyone: I understand where your coming from but isnt it better to use the filter within the table itself? The checkboxes are all doing the same thing (even if its more intuitive from a UI standpoint)

PR #419, opened 2018-12-29
Make Workspace testable and add a test for it

thomastanck:
There's a bug (with react-scripts-ts?) that causes yarn test to load a
different version of `re-resizable` compared to yarn start, so in order
to work around that I've added a commons component that just wraps
around re-resizable and makes it work no matter which version is loaded.
Ugly though.

### DIFF
--- a/src/components/commons/Resizable.tsx
+++ b/src/components/commons/Resizable.tsx
@@ -1,0 +1,7 @@
+import * as ResizableOriginal from 're-resizable'
+
+export { default as ResizableElement } from 're-resizable'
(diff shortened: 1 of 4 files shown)


ZiHawkEye:
Okay I'll take a look at it

martin-henz:
We have moved the repo from yarn to npm. Is this PR still relevant? Can you take a look, and if you think it's still relevant, rebase and find a suitable reviewer? If not, can you close it?

thomastanck:
Closing as we no longer use yarn

PR #421, opened 2019-01-08
Update js-slang

openorclose:
- Update libraries to use null as empty list
- Update imports


podocarp:
Using the npm js-slang 0.2.0 seems to cause a problem with importing `js-slang/interop`.
A local copy also throws the same error:
```
Module not found: Error: Can't resolve 'js-slang/dist/interop' in '[...]/cadet-frontend/src/utils'
```
I believe this is coming from Repl.tsx.

openorclose:
> Using the npm js-slang 0.2.0 seems to cause a problem with importing `js-slang/interop`.
> A local copy also throws the same error:
> 
> ```
> Module not found: Error: Can't resolve 'js-slang/dist/interop' in '[...]/cadet-frontend/src/utils'
> ```
> I believe this is coming from Repl.tsx.

I just did a fresh git clone and yarn of this branch, and can't seem to reproduce this issue. Could you try the same again and see if it persists?

src/utils/slangHelper.ts
```diff
@@ -34,13 +24,29 @@ declare global {
...
+ * e.g. [1, 2, 3] would be displayed as 1,2,3.
+ * An action is dispatched using the redux store reference
+ * within the global window object.
+ *
+ * @param value the value to be displayed
+ * @param workspaceLocation used to determine
+ *   which REPL the value shows up in.
+ */
+function rawDisplay(value: Value, workspaceLocation: any) {
+  display(String(value), workspaceLocation)
```
thomastanck: It should be the other way around: display should call rawDisplay. I'm not even sure this works correctly at the moment (seems like quotes will get escaped).

openorclose: Whoops yeah, they're flipped. I'll just swap them back. Yeah display will call stringify first and then call rawdisplay, escaping the strings.

openorclose:
@thomastanck can you reproduce the bug above?

podocarp:
I think the previous error was some dependency conflict, the merge from master probably solved it. All good now.

PR #422, opened 2019-01-10
Show grading status in assessment overview

jiachen247:


rrtheonlyone:
Nice work with the PR! 

However, I feel for a user, this icon is rather odd:

screenshot 2019-01-13 at 12 26 22 am

It is not very clear as to what it means (until you hover for the tooltip)

Shall we try something like this instead?

screenshot 2019-01-13 at 12 52 19 am

screenshot 2019-01-13 at 12 54 13 am

In order to do this, I made use of the `tag`:

```
const makeGradingStatus = (gradingStatus: string) => {
  let intent: Intent
  let label: string

  switch (gradingStatus) {
    case GradingStatuses.none:
      intent = Intent.PRIMARY
      label = "Ungraded"
      break

    case GradingStatuses.grading:
      intent = Intent.WARNING
      label = "Grading"
      break

    case GradingStatuses.graded:
      intent = Intent.SUCCESS
      label = "Graded"
      break

    default:
      intent = Intent.PRIMARY
      label = "Ungraded"
      break
  }



  return (
    <Tag 
      intent={intent} 
      round={true}
      interactive={true}
    >
      {label}
    </Tag>
  )
}
```

jiachen247:
As discussed, I think its cleaner to use icons instead and just add clarity using tooltips.

image

jiachen247:
updated icons to more consistent and better reflect the current state

image

Issue #420, opened 2018-12-31
More transparency when viewing results of assessments for students


jiachen247:
Currently only comments are being reflected to the students wrt grading.

Although breakdown of xp and grade are being passed to the frontend, they are not being reflected to the students which seems like a waste.

This comment tab only appears when a comment is added. If avengers were to grade without adding a comment, nothing would be shown to the student.

Looking at the timeline i doubt ill have time to work on the frontend this holiday!

I was thinking of a grading tab instead of a comment tab that we have currently.

This is purely a frontend fix. Does anyone want to take up this issue?

jiachen247:
Closed with #426

PR #425, opened 2019-01-10
Move max grade and max xp into question

jiachen247:
This is the accompanying frontend fix.

Now, max grade and xp will be available in the assessment workspace.
Which #420 will require.

To merge only after this backend pr has been merged.
source-academy/cadet#320


jiachen247:
@rrtheonlyone Opps i referenced the wrong issue.

PR #430, opened 2019-01-22
update .env.example template for clarity (Resolves #429)

### LINKED ISSUE #429: Parsing .env boolean values
Currently, in .env `REACT_APP_USE_BACKEND=` is left empty to set to false and setting any value here results in using the backend.

This is really counter intuitive, we should pattern match the value here to 1/true/ or 0/false to use or not use the backend respectively.

keep in mind that dotenv parses all valuse as strings
https://github.com/motdotla/dotenv/issues/51

And then update the readme  and template .env

jiachen247:
- updated .env.example to contain sensible defaults
- parsing of boolean value for REACT_APP_USE_BACKEND

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,4 @@
 REACT_APP_IVLE_KEY=your_ivle_key_here
 REACT_APP_VERSION=$npm_package_version
-REACT_APP_BACKEND_URL=http://something.com
-REACT_APP_USE_BACKEND=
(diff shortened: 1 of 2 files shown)


martin-henz:
Very low test coverage that will require a concerted effort. This particular decrease in test coverage seems spurious and should not hold back this PR.

PR #446, opened 2019-02-15
Use exactly enzyme 3.1.17

openorclose:
Compatibility issues with 3.1.18


openorclose:
It could be a npm/yarn version difference too, when deciding what version of enzyme to use. I'll try cloning again and see what ^3.1.17 gets

Issue #454, opened 2019-02-22
Data Visualiser: Display truncates beyond a certain size


lumos309:
See example [here](https://sourceacademy.nus.edu.sg/playground#chap=4&ext=NONE&prgrm=MYewdgzgLgBAhjAvDANgS2gCgIwBoYBM%2BAzPgCz4Cs%2BAbPgOz4Ac%2BAnPtgAwd4zZF9SfCn2p86fRnxZ92hTgEoA3ACgAJgCc4AdwD66LHGUqgA).

Issue #451, opened 2019-02-20
list visualiser: viewport not resizing in y-direction


martin-henz:
The viewport resizes in x-direction: When you re-size the browser window, the up-down scrollbar moves to the right edge of the window. However, the viewport does not re-size in y-direction: The left-right scrollbar is fixed at a particular position, and even overlaps with the REPL if the viewport is not tall enough.

lumos309:
The problem appears to originate from [this stylesheet file](https://github.com/source-academy/cadet-frontend/blob/master/src/styles/_playground.scss):

```
.kineticjs-content {
  /** 
   * Fixes a bug where the 1000x1000 canvas does not overflow
   * when it's container is too small.
   */
  width: 100% !important;
  height: 40vh !important; // 100% results in 0 height, 100vh is too big.
  overflow: auto;
}
```

Specifically, the "hard limit" of `height: 40vh` stops the visualiser panel from expanding beyond a certain height, even if the window is expanded.

Testing locally in Chrome and Edge, it seems that simply commenting out the entire block of code fixes the issue, and produced no problems with the overflow behaviour. Perhaps some native browser behaviour has changed since this fix was implemented?

martin-henz:
@remo5000 @ningyuansg : @lumos309 took a look at this and cannot reproduce the problem that you fixed with the snippet above. Do you remember details? Was this browser-specific?

ningyuansg:
https://github.com/source-academy/cadet-frontend/pull/268/commits/38d0bffe97dd502f7faedba7ee118fece68c9eb2

Is the commit adding that style. @remo5000 says that he will look into it.

remo5000:
IIRC, when I ported it over the canvas was fixed in size. I think I limited it so as to prevent weird overflowing issues. The best solution would be a resizing canvas (that works beyond the 1000x1000 limit). In addition, we would need to ensure that the visualised list is navigable. I'm all for removing it if the functionality is unaffected, but I can't reproduce any of the stuff right now due to having just wiped my computer @lumos309 @martin-henz

lumos309:
Update:

> changing `overflow` to `scroll`

as I mentioned above introduces an additional set of scroll bars within the visualiser panel- so essentially there are two levels of overflow handling, one within `kineticjs-content` and one within `side-content-text` (or one of its subclasses). Therefore it seems safe to remove the entire block of `kineticjs-content` styling, or alternatively change `overflow` to `none`.

[...]

PR #460, opened 2019-02-28
Fix for issues #451 & #454

lumos309:
[Issue #451](https://github.com/source-academy/cadet-frontend/issues/451): List viewport not resizing in y-direction.

Issue was caused by a previous bug fix that enabled overflow in the list visualiser panel, but limited the maximum height of the panel itself. Since overflow and scrolling now works properly even without the bug fix (tested on Chrome and Edge), it can be commented out to fix this current issue.


lumos309:
Fix for Issue #454 also: visualizer.js script now calculates the height of the list before using it to create the Kinetic Stage.

jiachen247:
Can you yarn format and push again please! @lumos309

lumos309:
> Can you yarn format and push again please! @lumos309

This is only for the .scss file and not the .js script, right?

public/externalLibs/visualizer/visualizer.js
```diff
@@ -2,16 +2,12 @@
   /**
    * Setup Stage
    */
+
```
jiachen247: what with the new line here?

public/externalLibs/visualizer/visualizer.js
```diff
@@ -698,6 +694,20 @@
     return this.image
   }
 
+  /**
+   *  Find the height of a drawing to determine canvas (stage) height
+   */ 
+  function findListHeight(xs) {
+      if ((!is_pair(xs) && !is_array(xs)) || is_empty_list(xs)) return 0;
```
jiachen247: I think itll be more readable and consistent if you use braces here.

lumos309:
@jiachen247 I made the changes you suggested and also edited the commenting style in the .scss file, and now it passes! `yarn format` didn't pick up on any problems, but I suspect it was an extra newline and/or using `//` in consecutive lines.

martin-henz:
Hmm, trying to build and run, but am getting this. Any idea what's happening?

TypeError: Cannot read property 'init' of undefined
ListVisualizer.componentDidMount
src/components/workspace/side-content/ListVisualizer.tsx:8
   5 | 
   6 | public componentDidMount() {
   7 |   if (this.$parent) {
>  8 |     (window as any).ListVisualizer.init(this.$parent);
   9 |   }
  10 | }
  11 | 
View compiled
▶ 13 stack frames were collapsed.

public/externalLibs/visualizer/visualizer.js
```diff
@@ -709,6 +719,11 @@
    *  Then shift it to the left end.
    */
   function draw(xs) {
+    stage = new Kinetic.Stage({
+      width: 1000,
```
jiachen247: Can I ask why width defaults to 1000? This creates some padding on the right that shouldnt be there.

image

lumos309: I'm not really sure how to go about finding the width. It seems like it would be tricky for lists that contains pointers to other lists...

martin-henz: I think you can find the width of the required window by visiting the lowest node on the right fringe of the tree. You start at the root and go right at each step until you reach a leaf. I think with the current tree drawing algorithm, you will find that that leaf is the node with the rightmost position.

lumos309: The problem is drawing like [these](https://sourceacademy.nus.edu.sg/playground#chap=4&ext=NONE&prgrm=MYewdgzgLgBAhjAvDANgS2gCnVgjAGhgCZCcpMBmASlI3IBZCBWGmABzjQCdMA2QgOxVWJGNQDcAKFCRYAIySo6mOIVUxcABkK4C8KlIAmXOAHcA%2BmUxyDQA), where there are both left and right nodes to account for, as well as multiple references to the same object. Nonetheless I think I may have a solution, which I'm working on now.

lumos309: Ok, I've sorted it out! Both width and height are now calculated dynamically.

martin-henz: > The problem is drawings like [these](https://sourceacademy.nus.edu.sg/playground#chap=4&ext=NONE&prgrm=MYewdgzgLgBAhjAvDANgS2gCnVgjAGhgCZCcpMBmASlI3IBZCBWGmABzjQCdMA2QgOxVWJGNQDcAKFCRYAIySo6mOIVUxcABkK4C8KlIAmXOAHcA%2BmUxyDQA), where there are both left and right nodes to account for, as well as multiple references to the same object. Nonetheless I think I may have a solution, which I'm working on now.

Wow, that's a weird example. Great that you worked it out.

martin-henz:
For this program
```
const a = list(list(1, 2, list(3), list(4, 5), pair(6, 7)), 2, 3);
const b = list(a, a, 10, 11, a);
draw_list(b);
```
I get:
```
Line 3: ReferenceError: is_empty_list is not defined
```
Note that we have replaced the function is_empty_list with is_null.

lumos309:
Oh right, I'd forgotten about that. Updated.

jiachen247:
Good job @lumos309!!

lumos309:
Thanks @jiachen247 ! I can delete the branch now right?

martin-henz:
Please don't delete the branch. I have merged it already with the master branch. So you can just checkout the master branch, and your changes should be in there. We keep merged branches in the repo, for future reference.

PR #466, opened 2019-03-01
Add auto-run for playground (#466)

martin-henz:
I suggest we merge this feature into master, so that we can get as much feedback as we can.


martin-henz:
@ccyccyccy can you take a look at the merge conflicts?

ccyccyccy:
The autorun functionality is only working for playground, and not working in the assessment workspace. Should I add it in before we actually merge?

Also, I don't know if we should preserve the state of autorun after the user closes the browser or should we let it default to off? Currently it remembers the user's choice and sets it when they come back.

ningyuansg:
> The autorun functionality is only working for playground, and not working in the assessment workspace. Should I add it in before we actually merge?

It's probably for the best that this feature is only available on the playground for now, so the students can catch bugs. We want to avoid bugs in the assessment workspace as far as possible, especially when the external libs are involved. Though, admittedly, I have not tested this feature out enough to judge how stable it is.

You can track auto-run for assessments workspace as a todo in the issues, or projects (whichever the active development team prefers).

This also gives time for discussion on if auto-run should be enabled for all assessments no matter what, or if some assessments should not have the auto-run feature. For example, perhaps in a particular assessment you want the students to practice visualising and predicting the output of their program as they write it. In this case, you may want to disable the option for auto-run to encourage the student to use their brains a little more.

> Also, I don't know if we should preserve the state of autorun after the user closes the browser or should we let it default to off? Currently it remembers the user's choice and sets it when they come back.

Intuitively, I think that auto-run should default to off whenever a workspace is freshly rendered.

1. I see some space for potential bugs if every source program that passes the validity test is evaluated automatically on load.
2. I think that auto-evaluating programs may discourage students from visualising the computational processes in their programs, since it is usually given to them live via auto-run. It is my *personal* opinion that this laziness should not be encouraged.

ningyuansg:
p.s. I also recommend the active dev team adopt [these specifications for good commit messages](https://chris.beams.io/posts/git-commit/). Since the *'Squash and merge'* buttons on github PRs generate the squash commit message from the PR title, this also implies good PR titles. In this case, it would be something like *"Add auto-run for playground (#466)"*.

martin-henz:
@ningyuansg that's an interesting discussion. 

You are saying that the auto-run feature may discourage students from visualizing (mentally) what happens when the program runs, and instead rely on their very eyes to do that "visualization".

My hunch is that the opposite is true for beginners: I think that providing an immediate connection between the program as it is being edited and the outcome and effects of the program will facilitate the formation of mental models. I do not have evidence for this, but I would love to get some.

I'm trying to push the Source Academy towards "immediate connection": Shortening the distance between the learner and the computational process that unfolds due to his actions. For a passionate case for "immediate connection", see Bret Victor's talk "Inventing on Principle", https://vimeo.com/36579366

I'm not 100% convinced that it will be beneficial, but I see an opportunity to move in this direction, and experiment. In the same vein, we will introduce an environment visualizer in the Source Academy, and a tool to visualize the substitution model. My objective at the moment is to develop these tools and get as much exposure as we can, in order to take a decision what to include and what not to include in the Source Academy. In the same spirit, I'm arguing for enabling the autorun-feature for missions.

PR #506, opened 2019-03-30
INSPECT

podocarp:
## Inspector
This requires the use of the `debugger` branch of js-slang to work. Clone both the frontend and the `debugger` slang to the same directory, then `npm install`. If that give some errors about slang, you can try removing the entries for slang in the lock file, then install again. The merge over there is still ongoing. Meanwhile, please try to break this.

The mental model we are using is: A breakpoint means that the interpreter will stop right before it. Whatever is highlighted is going to be evaluated next. If you meet any inconsistencies with this, also please raise it up for discussion.

## What you can do
- Set breakpoints by clicking on the gutter
- `debugger;` just like ECMAScript
- Inspect!
- Run stuff in the context of the paused program!

## Usage
Here's what happens: After you click run, if there the interpreter meets a breakpoint, the first thing you're going to notice is that the REPL feedbacks to you it hit a breakpoint, the line is highlighted, and one of the icons on the right pane is going to start blinking. If you click on the icon, it reveals the inspector. All the variables in every frame is exposed here. The REPL is also now in the context of where ever you are. So you can evaluate anything you would normally be able to in the REPL. It is all quite simple really.

## Note
Because we use a local version of `js-slang`, the CI just breaks all the time.


thomastanck:
Would you mind showing a diagram showing how the control flow of the feature looks like? Or at least, a list of important functions in which files and what they do. Thanks!

thomastanck:
By the way, you might want to fix the build and re-merge master.

martin-henz:
> Would you mind showing a diagram showing how the control flow of the feature looks like? Or at least, a list of important functions in which files and what they do. Thanks!

Yes, that would be good, eventually. For now, we can merge by adding the stuff below in the README of cadet-frontend.

(copied from above)
Inspector
This requires the use of the debugger branch of js-slang to work. Clone both the frontend and the debugger slang to the same directory. You would want to yarn build the slang you just obtained and then yarn && sudo yarn start in the frontend and it should just work. The merge over there is still ongoing. Meanwhile, please try to break this.

The mental model we are using is: A breakpoint means that the interpreter will stop right before it. Whatever is highlighted is going to be evaluated next. If you meet any inconsistencies with this, also please raise it up for discussion.

What you can do
Set breakpoints by clicking on the gutter
debugger; just like ECMAScript
Inspect!
Run stuff in the context of the paused program!
Usage
Here's what happens: After you click run, if there the interpreter meets a breakpoint, the first thing you're going to notice is that the REPL feedbacks to you it hit a breakpoint, the line is highlighted, and one of the icons on the right pane is going to start blinking. If you click on the icon, it reveals the inspector. All the variables in every frame is exposed here. The REPL is also now in the context of where ever you are. So you can evaluate anything you would normally be able to in the REPL. It is all quite simple really.

Note
Because we use a local version of js-slang, the CI just breaks all the time.

lumos309:
Can I request @jiachen247 to review the additions for the env model visualizer? Thanks!

martin-henz:
There seems to be a bug with inspecting parameters inside a breakpoint:

Screenshot 2019-04-13 at 5 49 31 PM

podocarp:
> There seems to be a bug with inspecting parameters inside a breakpoint:
> 
> Screenshot 2019-04-13 at 5 49 31 PM

Hmm I seem to be able to get it working. 
2019-04-13-204049_944x1043_scrot

Perhaps try cloning it brand new somewhere and see if it is still persists?

martin-henz:
The error appears only after you try to display the value of "x" in the REPL.

podocarp:
> The error appears only after you try to display the value of "x" in the REPL.

Fixed in latest commit along with the latest slang version.
Furthermore, you can now pause anywhere. Where the execution stops is entirely up to how fast you can click the button. For example, try this:

```
function a(x){
    for(let i = 0; i < 3; i=i+1){
        display(1);
        a(1);
        display(1);
    }
}

a(1);
```

This executes forever so it gives you plenty of time to click the pause button. Then you're given the same highlighting and the same inspector as though there's a breakpoint there.

Also, there is something quirky with this. The stack overflow detector does not actually terminate this code. If you remove the display functions you will see that it does. Might be a separate issue.

martin-henz:
> Also, there is something quirky with this. The stack overflow detector does not actually terminate this code. If you remove the display functions you will see that it does. Might be a separate issue.

Hmm, strange. So this runs with constant space consumption?

podocarp:
> > Also, there is something quirky with this. The stack overflow detector does not actually terminate this code. If you remove the display functions you will see that it does. Might be a separate issue.
> 
> Hmm, strange. So this runs with constant space consumption?

Looking at the chrome process manager shows that the memory consumption keeps increasing, so I would think otherwise.

martin-henz:
> > > Also, there is something quirky with this. The stack overflow detector does not actually terminate this code. If you remove the display functions you will see that it does. Might be a separate issue.
> > 
> > 
> > Hmm, strange. So this runs with constant space consumption?
> 
> Looking at the chrome process manager shows that the memory consumption keeps increasing, so I would think otherwise.

Interesting. So the memory goes into the JavaScript heap and the stack does not grow. Or the stack grows slowly compared to other stuff that's happening.

PR #519, opened 2019-04-10
Mission editing master

martin-henz:


martin-henz:
I cannot start the frontend in this branch. After npm install and npm start, I get:

Failed to compile.

/Users/henz/Documents/source-academy/cadet-frontend/src/mocks/context.ts
(38,24): Argument of type 'acorn.Node' is not assignable to parameter of type 'import("/Users/henz/Documents/source-academy/cadet-frontend/node_modules/@types/estree/index").Node'.
  Type 'Node' is not assignable to type 'ExportSpecifier'.
    Property 'exported' is missing in type 'Node'.

martin-henz:
Screenshot 2019-04-13 at 11 54 13 AM

There is a spurious closed bracket in the template editor.

martin-henz:
src/components/workspace/ControlBar.tsx
```
  private replControl() {
    // toggleEditMode can be null or undefined here, thus !=
    const toggleEditModeButton =
      this.props.toggleEditMode !== null ? (
```
In Playground, the toggleEditMode appears to be undefined, not null. Thus, != might be more appropriate. Otherwise this happens:

Screenshot 2019-04-13 at 12 03 00 PM

martin-henz:
The question template does not seem to be saved properly. 

How to reproduce the bug:
(1) create a new mission
(2) edit the template of Question 1
(3) save
(4) re-load the mission
Now the template is back to the original "//This is a mock solution template".

martin-henz:
This tool is an editor: It allows the user to open an existing mission (that is stored in a file), create a new mission, and save the current mission (to a file). The notion of "import" and "export" is not appropriate here. We should take a look at established editors such as Word, and adopt their terminology.

The options in "Mission Control" should be:
New Mission (creates a new mission without saving it anywhere)
Open... (prompts the user to navigate to a file in a finder, and open it)
Save (saves to the most recently opened or saved location, or prompts the user to give a location)
Save as... (prompts the user to give a location)

This scheme will allow us to eliminate the highly unusual choice of setting a Filename under "Other options". A filename is not an "option" but an essential concept for such an editor.

Screenshot 2019-04-13 at 12 50 01 PM

blackening:
> This scheme will allow us to eliminate the highly unusual choice of setting a Filename under "Other options". A filename is not an "option" but a central concept for such an editor.
> 
> Screenshot 2019-04-13 at 12 50 01 PM

Actually there's no real requirement for filename. After all, it will be uploaded to bitbucket manually, and the file name can be changed then.

martin-henz:
In addition to the usual features "New Mission", "Save", "Save As..." and "Open..." we should have a "Play" option in the "Mission Control" interface. This feature would load the mission without any of the editing features. It would allow us to test the client-side features of the mission before uploading it to any server.

martin-henz:
> > This scheme will allow us to eliminate the highly unusual choice of setting a Filename under "Other options". A filename is not an "option" but a central concept for such an editor.
> > Screenshot 2019-04-13 at 12 50 01 PM
> 
> Actually there's no real requirement for filename. After all, it will be uploaded to bitbucket manually, and the file name can be changed then.

Yes, let the file selector deal with the filename. that's the most natural way that the users will be familiar with.

martin-henz:
Can you guys put some example mission files in the repo "assessments" in folder "samples". Maybe you want to make clear by the choice of their filenames that they are meant as test files for the mission editor.

gerhean:
> This tool is an editor: It allows the user to open an existing mission (that is stored in a file), create a new mission, and save the current mission (to a file). The notion of "import" and "export" is not appropriate here. We should take a look at established editors such as Word, and adopt their terminology.
> 
> The options in "Mission Control" should be:
> New Mission (creates a new mission without saving it anywhere)
> Open... (prompts the user to navigate to a file in a finder, and open it)
> Save (saves the current mission to the most recently "Open..." or "Save as..." location, or prompts the user to give a location, if the current mission is a "New Mission")
> Save as... (prompts the user to give a location)
> 
> This scheme will allow us to eliminate the highly unusual choice of setting a Filename under "Other options". A filename is not an "option" but a central concept for such an editor.
> 
> Screenshot 2019-04-13 at 12 50 01 PM

It is not possible to force users to save at any particular location due to Javascript security concerns. Only the client, in their browser, set that they want to be prompted where to save their files.

martin-henz:
> The options in "Mission Control" should be:
> New Mission (creates a new mission without saving it anywhere)
> Open... (prompts the user to navigate to a file in a finder, and open it)
> Save (saves the current mission to the most recently "Open..." or "Save as..." location, or prompts the user to give a location, if the current mission is a "New Mission")
> Save as... (prompts the user to give a location)

I see. How about this, then:

The options in "Mission Control" should be:
New Mission (creates a new mission without saving it anywhere)
Open... (prompts the user to navigate to a file in a finder, and open it)
Save (saves the current mission using the most recently "Open..." or "Save as..." file name, or prompts the user to give a file name in an input field, if the current mission is a "New Mission")
Save as... (prompts the user to give a file name in an input field)
Play (opens the current mission, as a real mission, without editing)

gerhean:
> There is a spurious closed bracket in the template editor.

Fixed

> src/components/workspace/ControlBar.tsx
> 
> ```
>   private replControl() {
>     // toggleEditMode can be null or undefined here, thus !=
>     const toggleEditModeButton =
>       this.props.toggleEditMode !== null ? (
> ```
> In Playground, the toggleEditMode appears to be undefined, not null. Thus, != might be more appropriate. Otherwise this happens:

Fixed

gerhean:
> > There is a spurious closed bracket in the template editor.
> 
> Fixed
> 
> > src/components/workspace/ControlBar.tsx
> > ```
> >   private replControl() {
> >     // toggleEditMode can be null or undefined here, thus !=
> >     const toggleEditModeButton =
> >       this.props.toggleEditMode !== null ? (
> > ```
> > In Playground, the toggleEditMode appears to be undefined, not null. Thus, != might be more appropriate. Otherwise this happens:
> 
> Fixed

> The question template does not seem to be saved properly.
> 
> How to reproduce the bug:
> (1) create a new mission
> (2) edit the template of Question 1
> (3) save
> (4) re-load the mission
> Now the template is back to the original "//This is a mock solution template".

martin-henz:
We are almost there with this. 

As discussed, let's remove the buttons "Save" and "Reset".

Changes in the the template (and everywhere else) should have immediate effect.

Shuming's "Reset" button should therefore work as expected: It should copy the template from the right hand side to the left hand side.

Therefore, you can remove the button "Copy to Editor".

PR #520, opened 2019-04-10
Sound improvements

TheRealRavager:
-Removed distinction between sourcesound (function takes in time and returns amplitude) and sound (discretized sound samples stored as arrays).

-Edited all abstractions that rely on sound to now utilise sourcesound (limit unnecessary processing of arrays)

-Edited Play(sourcesound) function to utilise webaudio API and buffer the playback of sound to make it instantaneous


public/externalLibs/sound/sounds.js
```diff
@@ -111,33 +111,139 @@ function get_duration(sourcesound) {
...
+            return data;
+        }
+    }
 
-  _audio = raw_to_audio(data)
-  _audio.addEventListener('ended', stop)
-  _audio.play()
-  _playing = true
+    // Schedules playback of sounds (Once keeps track of whether first instance, first is first or second sound)
+    function ping_pong(wave, duration, elapsed_duration,
```
rrtheonlyone: This seems to be core of your new implementation. Could you elaborate more on what this function actually does? 

It is not super clear from your comments.

martin-henz:
I agree that the pingpong function needs to be explained better.

A suggestion: In the recursive calls within "if (first) ...", just reverse the roles of sound1 and sound2. That way, you can remove the parameter "first", and get rid of "if (first)...", and save the ugly repeated branch. This will make your program more maintainable.

martin-henz:
Another question (probably for the next version): What happens if discretize_from takes longer than playing the sound? I suspect we need to have some fall-back position: If this happens, maybe we alert the student, and (maybe) offer an alternative play function:
play_safe
perhaps, which would work as in the previous version?

public/externalLibs/sound/sounds.js
TheRealRavager: Hi! I've refactored my play() function in the latest commit. Removed loads of repetitive code. Let me know if its still unclear

public/externalLibs/sound/sounds.js
```diff
@@ -1,490 +1,501 @@
...
+    for (var i = 0; i < duration * FS; i++) {
+        vector.push(wave( i / FS));
+    }
 
-  return vector
+    return vector;
+}
+
+// Discretizes a sound to a sound starting from elapsed_duration, for sample_length seconds
+function discretize_from(wave, duration, elapsed_duration, sample_length, data) {
```
rrtheonlyone: Cant you make this into one for loop? (You jus need to change the terminating condition)

e.g.
```javascript
var completion = (elapsed_duration + sample_length > duration) ? duration * FS : (elapsed_duration + sample_length) * FS;
for (var i = elapsed_duration * FS; i < completion; i++) {
    data[i - elapsed_duration * FS] = wave(i / FS); 
}
return data;
```

Issue #521, opened 2019-04-10
Updating Dependencies


jiachen247:
This is a list of outdated deps. As seen some deps have new major releases. Someone should go over this list and see what is safe and not safe to bump!

```
[jiachen@jiachen cadet-frontend]$ npm outdated
Package                               Current         Wanted   Latest  Location
@blueprintjs/core                       2.3.1          2.3.1   3.15.1  cadet-frontend
@blueprintjs/icons                      3.6.0          3.8.0    3.8.0  cadet-frontend
@blueprintjs/select                     2.0.1          2.0.1    3.8.0  cadet-frontend
@types/acorn                            4.0.3          4.0.5    4.0.5  cadet-frontend
@types/classnames                       2.2.6          2.2.7    2.2.7  cadet-frontend
@types/common-tags                      1.4.0          1.8.0    1.8.0  cadet-frontend
@types/draft-js                       0.10.24        0.10.30  0.10.30  cadet-frontend
@types/enzyme                          3.1.17         3.1.17    3.9.1  cadet-frontend
@types/enzyme-adapter-react-16          1.0.3          1.0.5    1.0.5  cadet-frontend
@types/jest                            22.2.3         22.2.3  24.0.11  cadet-frontend
@types/lz-string                       1.3.32         1.3.33   1.3.33  cadet-frontend
@types/node                           11.13.0        11.13.4  11.13.4  cadet-frontend
@types/pixi.js                          4.8.0          4.8.7    4.8.7  cadet-frontend
@types/query-string                     5.1.0          5.1.0    6.3.0  cadet-frontend
@types/react                          16.4.11        16.8.13  16.8.13  cadet-frontend
@types/react-dom                       16.0.7         16.8.3   16.8.3  cadet-frontend
@types/react-redux                     5.0.21         5.0.21    7.0.6  cadet-frontend
@types/react-router                    4.0.30          4.4.5    4.4.5  cadet-frontend
@types/react-router-dom                 4.3.0          4.3.1    4.3.1  cadet-frontend
@types/react-router-redux              5.0.15         5.0.18   5.0.18  cadet-frontend
@types/react-test-renderer             16.0.2         16.8.1   16.8.1  cadet-frontend
@types/redux-mock-store                0.0.13         0.0.13    1.0.0  cadet-frontend
@types/showdown                         1.7.5          1.9.2    1.9.2  cadet-frontend
acorn                                  4.0.13         4.0.13    6.1.1  cadet-frontend
[...]

PR #488, opened 2019-03-21
Updated dependencies. Added CORS/windows scripts.

blackening:
-From primordial edition to dinosaur edition.
-Also added convenience scripts for CORS preflight and windows.

For this PR, anyone who tests it, simply put a comment that it works. Once we have 3-5 comments, just merge it.

A good reason to test this is if you git pull, yarn install and it explodes.
- e.g. "A rest parameter must be of an array type" from enzyme

----------------------------------
Changed this:
image

To this:
image

PR #522, opened 2019-04-10
CRITICAL FIX (rollback bumping deps)

jiachen247:
This PR rolls back all the dependencies bumped in #488 due to a compilation error

A new issue has been created to deal with the bumping of versions (#521)

We should also consider adding typescript compilation as a pre-hook before we are allowed to push to the remote.


openorclose:
I changed travis to do npm run build too.

PR #540, opened 2019-04-18
Video testing

JJiahui:


geshuming:
Can we import the p5 library via npm? I would rather not bloat the source code with the p5 library when you can install it via npm. https://www.npmjs.com/package/p5

JJiahui:
I have changed the p5 libraries to minified versions.
Some code in the p5 library needs to be edited, so we cannot import it via npm.

martin-henz:
Last time I used the video tool, there was an issue with my mac heating up even when I didn't use the video. Could it be that somehow recording continues to happen, after starting VIDEO? Is there a way to turn off the video processing when no one is watching?

JJiahui:
> 
> 
> Hmm, can you be more specific, what you had to change in p5.dom?
> 
> I'm concerned about the long run: Once we include a library literally like that, we are cut off from future improvements and bug fixes.

I deleted one line in p5.min.js:

this.createCanvas(
      this._defaultCanvasSize.width,
      this._defaultCanvasSize.height,
      'p2d');

Doing this stopped the problem of excessive ram usage.

martin-henz:
> > Hmm, can you be more specific, what you had to change in p5.dom?
> > I'm concerned about the long run: Once we include a library literally like that, we are cut off from future improvements and bug fixes.
> 
> I deleted one line in p5.min.js:
> 
> this.createCanvas(
> this._defaultCanvasSize.width,
> this._defaultCanvasSize.height,
> 'p2d');
> 
> Doing this stopped the problem of excessive ram usage.

Looks like a rather desperate way to solve the problem. Are you sure there is no other way to prevent excessive ram usage? Can the canvas size be set in other ways?

JJiahui:
The problem was that a default canvas would be created on every page, even if I didn't actually call the function "createCanvas".
I tried removing that line of code and it seemed to solve the problem.
I'm not sure how we would deal with this in the long run.

martin-henz:
> The problem was that a default canvas would be created on every page, even if I didn't actually call the function "createCanvas".
> I tried removing that line of code and it seemed to solve the problem.
> I'm not sure how we would deal with this in the long run.

@JJiahui I don't dispute that your change solves the problem at hand. What I'm saying is that the solution comes at a high cost: We lose connection to the p5.dom development effort and cut ourselves from future p5.dom improvements and bug fixes.

My hunch is that there should be a better way. Can you investigate more?

JJiahui:
I've investigated two ways, putting the p5 library in the source code and importing via npm. 
Both ways I found that merely loading/importing the library would cause excessive ram usage, without any additional code from me. 

This was what led me to think that the bug lies within the p5 library itself.

martin-henz:
Hmm, a cursory search leads me to thing that p5 has had cpu usage issues, which is not surprising per se. You may be able to find reports on the specific performance issue that you faced. Literally copying the library is a drastic measure. If you are convinced that this is a bug in p5.dom, can you file a bug report with p5? See what they say. Is p5.dom the only library that provides the service that you need?

martin-henz:
Just took a closer look at this: P5 is distributed under the LGPL, see
https://github.com/processing/p5.js/blob/master/license.txt

This means we cannot do what you propose: include a modified minified version of the library. We would need to apply LGPL explicitly to our modification. This would complicate the licensing of our entire system.

JJiahui:
As I have discovered a way to make it work without the p5 libraries, I have removed them entirely.

martin-henz:
Can you resolve the conflicts, @JJiahui ? Thanks!

geshuming:
image
I'm getting this error and no video is displayed.

martin-henz:
> image
> I'm getting this error and no video is displayed.

Hmm, works for me. Can you be more specific, @geshuming ?

martin-henz:
If I set the size to 1000x700 (I know: quite large), I get tons of errors like this in the console:

video_lib.js:199 [Violation] 'requestAnimationFrame' handler took 67ms
video_lib.js:199 [Violation] 'requestAnimationFrame' handler took 55ms
video_lib.js:199 [Violation] 'requestAnimationFrame' handler took 51ms

We may be able to catch these and issue a warning for the user.

Also maybe we should suggest some common screen dimensions? Is there a way to get the native pixel size of the camera?

geshuming:
Yup I'm using chrome latest version. Firefox latest version was also giving me the same error

martin-henz:
Did you give permission to use your camera?

martin-henz:
@JJiahui can we be more specific with error messages? What are possible error codes from navigator.mediaDevices.getUserMedia ? 

Here is a starting point: https://blog.addpipe.com/common-getusermedia-errors/

geshuming:
> Did you give permission to use your camera?

Yep permission is given. The error above is all i'm getting.

martin-henz:
My recent changes:

* changed default resolution to 400x300
* changed textinput to NumericInput
* added some icons from the list in https://evergreen.segment.com/components/icons/
* changed the pause button to a "snap picture" button
* changed apply_filter so that it updates the snapped picture

martin-henz:
A couple of remaining issues:

* Re-sizing should update the snapped picture, but currently, the picture goes black, and the user has to press "snap" again. @JJiahui can you take a look?
* The formatting of the buttons is poor: I just inserted some "nbsp;", but that's obviously not the right way. @alcen can you see if there is a better way.

alcen:
> A couple of remaining issues:
> 
> Re-sizing should update the snapped picture, but currently, the picture goes black, and the user has to press "snap" again. @JJiahui can you take a look?
> The formatting of the buttons is poor: I just inserted some "nbsp;", but that's obviously not the right way. @alcen can you see if there is a better way.

Ok, I'll look into it!

PR #556, opened 2019-05-09
Consolidate run and autorun buttons

### LINKED ISSUE #536: Improve UX of "Run" and "Autorun"
We should have a button

☐ Run

Clicking on "Run" gives you the usual running of programs. Clicking on ☐ leads to

☑ Auto

and autorun is turned on: Any edit that leads to a syntactically correct program will be run. The "Auto" would be not clickable.

Now, clicking on 

☑ 

toggles back to

☐ Run

jiayushe:
Fixes #536

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,6 @@
   },
   "devDependencies": {
     "@blueprintjs/core": "^2.1.1",
-    "@blueprintjs/docs-theme": "^3.2.0",
     "@blueprintjs/icons": "^3.6.0",
     "@blueprintjs/select": "^2.0.1",
     "@types/acorn": "^4.0.3",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
   },
   "devDependencies": {
     "@blueprintjs/core": "^2.1.1",
+    "@blueprintjs/docs-theme": "^3.2.0",
     "@blueprintjs/icons": "^3.6.0",
     "@blueprintjs/select": "^2.0.1",
     "@types/acorn": "^4.0.3",
--- a/src/components/workspace/ControlBar.tsx
+++ b/src/components/workspace/ControlBar.tsx
@@ -9,7 +9,6 @@ import {
   Text,
   Tooltip
 } from '@blueprintjs/core';
-import { handleBooleanChange } from '@blueprintjs/docs-theme';
 import { IconNames } from '@blueprintjs/icons';
 import { ItemRenderer, Select } from '@blueprintjs/select';
 import * as React from 'react';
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -78,6 +78,7 @@
   },
   "devDependencies": {
     "@blueprintjs/core": "^2.1.1",
+    "@blueprintjs/docs-theme": "^3.2.0",
```
jiachen247: sorry im a little confused! why are we adding a documentation dep in this pr?

src/components/workspace/ControlBar.tsx
```diff
@@ -251,15 +256,28 @@ class ControlBar extends React.PureComponent<ControlBarProps, { value: string }>
...
+      </div>
+    ) : (
+      undefined
+    );
+
     return (
       <div className="ControlBar_editor pt-button-group">
-        {this.props.isRunning ? stopButton : this.props.isDebugging ? resetButton : runButton}
+        {toggleAutorunButton}
+        {this.props.isEditorAutorun
```
jiachen247: can we put this into function!! too must business logic for ternary operators.

src/components/workspace/ControlBar.tsx
```diff
@@ -196,7 +197,11 @@ class ControlBar extends React.PureComponent<ControlBarProps, { value: string }>
           this.props.handleSetEditorSessionId!('');
         }
       };
-      xmlhttp.open('GET', 'https://' + LINKS.SHAREDB_SERVER + 'gists/' + this.state!.value, true);
+      xmlhttp.open(
+        'GET',
+        'https://' + LINKS.SHAREDB_SERVER + 'gists/' + this.state!.joinElemValue,
```
jiachen247: this should be externalized into another component leh we shouldn't be having logic in a dumb component such as the control bar.

martin-henz: The control bar turns out to be interesting. I think it should have the requested logic, in order to support run/autorun. I feel that run/stop/autorun can be succinctly supported in the requested feature. Let's merge this and see if we get feedback from users.

package.json
jiayushe: Yea it is redundant, we have removed it.

src/components/workspace/ControlBar.tsx
ccyccyccy: By externalized do you mean putting it into a helper file? If so we can do it in another PR.

jiachen247: yes I meant another file! if you think about it if the url changes lets say `gists` or `lastest` get changed to something else the control bar is the definitely wrong place to come to change things. As prof always says think in levels of abstraction!

jiachen247: I'll approve but at the rate we're going something is gonna break soon haha

PR #580, opened 2019-06-07
Frontend testing: Add tests for session.ts reducer

jon-chua:
Part of #581 

- added tests for session.ts reducer

### DIFF
--- a/src/reducers/__tests__/session.ts
+++ b/src/reducers/__tests__/session.ts
@@ -1,0 +1,387 @@
+import { reducer } from '../session';
+import {
+  IAction,
+  LOG_OUT,
+  SET_TOKENS,
+  SET_USER,
+  UPDATE_ASSESSMENT,
+  UPDATE_ASSESSMENT_OVERVIEWS,
+  UPDATE_GRADING,
+  UPDATE_GRADING_OVERVIEWS,
+  UPDATE_HISTORY_HELPERS
+} from '../../actions/actionTypes';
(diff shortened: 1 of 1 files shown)


src/reducers/__tests__/session.ts
```diff
@@ -0,0 +1,387 @@
...
+  const action: IAction = {
+    type: UPDATE_ASSESSMENT,
+    payload: assessmentTest1
+  };
+  const resultMap: Map<number, IAssessment> = reducer(defaultSession, action).assessments;
+
+  expect(resultMap.get(assessmentTest1.id)).toEqual(assessmentTest1);
+});
+
+test('UPDATE_ASSESSMENT works correctly in updating assessment', () => {
```
rrtheonlyone: We can add one more test here!

We have not tested whether the old data in the map is retained (we have only checked the latest data we add/update)

src/reducers/__tests__/session.ts
```diff
@@ -0,0 +1,387 @@
...
+      submissionId,
+      grading: gradingTest1
+    }
+  };
+
+  const gradingMap: Map<number, Grading> = reducer(defaultSession, action).gradings;
+  expect(gradingMap.get(submissionId)).toEqual(gradingTest1);
+});
+
+test('UPDATE_GRADING works correctly in updating gradings', () => {
```
rrtheonlyone: Similar to above^

Issue #581, opened 2019-06-07
Frontend Testing 


rrtheonlyone:
As our frontend code base grows, it is also important to have some level of tests to prevent regression.

We can start with some unit tests for our actions/reducers. This issue will track the progress of this project.

PR #582, opened 2019-06-07
Frontend-Testing : Test Academy Reducers

cheongsiuhong:
Included tests for LOG_OUT and SAVE_CANVAS reducers.
Part of #581


src/reducers/__tests__/academy.ts
```diff
@@ -0,0 +1,26 @@
...
+  };
+  const result: IAcademyState = reducer(defaultAcademy, action);
+
+  expect(result).toEqual(defaultAcademy);
+});
+
+test('SAVE_CANVAS works correctly on default academy', () => {
+  const action: IAction = {
+    type: SAVE_CANVAS,
+    payload: null
```
rrtheonlyone: We ideally dont want to be using `null` as the payload. The payload for this is of type `HTMLCanvasElement`. 

You can create one using this function below:
```
function createContext(width : number, height : number) : HTMLCanvasElement {
    let canvas = document.createElement('canvas');
    canvas.width = width;
    canvas.height = height;
    return canvas
}
```

cheongsiuhong: Changes have been made, thanks Rahul!

PR #585, opened 2019-06-08
Make REPL work in paused state

### LINKED ISSUE #558: REPL in paused state no longer works
The REPL used to work after the debugger paused. Right now there seems to be some issues with it:

- The only thing working now is to expose variables, for example just `i;` or `x;`. 
- Executing other statements like `f(x-10)` or `x=100` will end the program
- In for loops the loop counter is now seen as a constant and cannot be changed. The REPL prints an error message telling us this, and the program ends.

podocarp:
Fixes #558


podocarp:
Fix #586  as well

PR #593, opened 2019-06-09
[ENV] Visualizer Fixes

lumos309:
Deleted duplicate files and fixed case-sensitive imports.


geshuming:
May i know how the visualizer works? I cannot get anything to be displayed, with or without merging the current master. Is there a snippet of code that you can provide as an example?

lumos309:
> May i know how the visualizer works? I cannot get anything to be displayed, with or without merging the current master. Is there a snippet of code that you can provide as an example?

Did you set a breakpoint in the code editor before pressing run?

geshuming:
Would you like to fix some issues before merging? 

1) It will be great if we can see the environment just by running the code without a breakpoint.

2) The canvas size is fixed, which by default does not fit the website. Can the canvas size be made dynamic (i.e. fits the actual content?) 

canvas_example

3) The visualizer cannot render lists? 
broken_list

geshuming:
If not, we can just log these as issues for now.

martin-henz:
Regarding:

**Would you like to fix some issues before merging?

It will be great if we can see the environment just by running the code without a breakpoint.**

This is a feature, not a bug: We only take the environment that is accessible at the time of the breakpoint. This is consistent with the textbook. Frames that are not accessible are not displayed. We'll change the lecture material and all our homework to be consistent with this view.

martin-henz:
Regarding:

**The canvas size is fixed, which by default does not fit the website. Can the canvas size be made dynamic (i.e. fits the actual content?)**

Yes, please file an issue.

martin-henz:
Regarding:

**The visualizer cannot render lists?**

Yes, you can file an issue, and mark it as __postponed__

Issue #609, opened 2019-06-20
About, Team, Help pages


rrtheonlyone:
Now that we have grown as a team, it is important to recognise the efforts that all of us have put in.

We should work on an improved contributors page listing our teams' contributions for all our various projects (Github allows for this information to be retrieved). Any suggestions for this page can be put in this issue as well!

martin-henz:
I suggest that we change the DOWN button on the top right as follows:
* About: 

The Source Academy is an experimental environment for learning programming in the CS1101S module at the National University of Singapore.

This iteration of Source Academy, code-named Cadet, is available under the MIT License. You may find the source code for Cadet at our GitHub organisation, Source Academy.

* Team:

One page with the following:

Leadership
Backend: 
Frontend Team

Lee Ning Yuan • Vignesh Shankar • Rahul Rajesh

Backend Team
 • Liow Jia Chen

Source Team

Tan Chee Kun, Thomas

Hall of Fame:

Ng Tse Pei • Joey Yeo • Tan Yu Wei





* Help: similar to what we have

geshuming:
Looks good

sandydays:
From what I can see, there are two overarching issues we need to tackle:

1. How should a user be able to navigate to the Contributors page? The route to the contributors page via /contributors has been created, but where should a user to click to reach this page? We want to make it prominent. But there are few issues with how we are going to go about doing that:

a) Dropdown > About > Developers tab > 'here' word that when clicked links to the contributors page --> this is what I had implemented previously, but we don't know whether it would be prominent enough.

b) Another idea I had was very similar to what Prof Matrin has mentioned: adding a Team menu item under the dropdown, but this might be inconsistent since all the menu items when clicked lead to pop up but when Team is clicked it will take the user to a completely new page.

However, Prof, you have mentioned that clicking Team still takes us to a popup page with a brief overview of the leadership, hall of fame, etc. and then we have all contributors listed. This would solve the issue of any inconsistencies. However, to clarify, are All Contributors to be rendered in that popup or is there is a link that takes the user to a separate page where all contributors would be shown? 

Showing everything in the popup could get messy unless we show just the top contributors and then link the rest to another page. On the other hand, if we are not showing the contributors in the popup and just linking to another page, we need to consider whether this is prominent enough (I think it should be okay though). 

c) Another idea is to put it out as a button in the main navbar, but I feel that is not the right place for it to be in thematically. 

[...]

PR #612, opened 2019-06-20
Basic Outline of the Contributors Page #609

sandydays:
The basic outline of the page has been done, including division into sections of what the page could contain and fetching data from the github endpoint of all the contributors and rendering that.

### DIFF
--- a/src/components/Contributors.tsx
+++ b/src/components/Contributors.tsx
@@ -1,4 +1,3 @@
-import { Component } from 'react';
 import * as React from 'react';
 
 type Contributor = {
@@ -17,26 +16,23 @@ type Repo = {
 };
 
 type ContributorsState = {
-    ignoreRepos: string[];
-    ignoreContributors: string[];
     repos: Repo[];
     contributors: Contributor[][];
 };
 
-class Contributors extends Component<{}, ContributorsState> {
-
-    constructor(props: any) {
-        super(props);
+class Contributors extends React.Component<{}, ContributorsState> {
     
+    private readonly ignoreRepos: string[] = ["sicp", "assessments", "tools", "source-academy2"];
+    private readonly ignoreContributors: string[] = ["dependabot[bot]", "dependabot-preview[bot]"];
+
+    constructor(props: {}) {
+        super(props);
+
         this.state = {
-            ignoreRepos: ["sicp", "assessments", "tools", "source-academy2"],
-            ignoreContributors: ["dependabot[bot]", "dependabot-preview[bot]"],
             repos: [],
             contributors: []
         };
     }
-
-    
 
     public componentDidMount() {
 
@@ -45,7 +41,7 @@ class Contributors extends Component<{}, ContributorsState> {
             return results.json();
         })
         .then(repos => {
-            const { ignoreRepos } = this.state;
+            const { ignoreRepos } = this;
             const contributorLinks = repos
                 .filter((repo: any) => {
                     return !ignoreRepos.includes(repo.name);
@@ -79,7 +75,7 @@ class Contributors extends Component<{}, ContributorsState> {
             .then(contributorsByRepo => {
                 return Promise.all(
                     contributorsByRepo.map((contributors: any) => {
-                        const { ignoreContributors } = this.state;
+                        const { ignoreContributors } = this;
                         const contributorList = contributors
                             .filter((contributor: any) => {
                                 return !ignoreContributors.includes(contributor.login);
@@ -113,15 +109,19 @@ class Contributors extends Component<{}, ContributorsState> {
                         <div key={contributor.key}>
                             <img src={contributor.photo} alt="Image" />
                             <p><a href={contributor.githubPage} target="_blank">{contributor.githubName}</a></p>
-                            <p>Number of commits: {contributor.commits}</p>
+                            <p>Commits: {contributor.commits}</p>
                         </div>
                     );
                 });   
                 return (
-                    <div key={repo.key} className="repoDetails">
-                        <h3><u>{repo.name}</u></h3>
-                        <h5>{repo.description}</h5>
-                        <div className="contributorsByRepo">{arrayMapped}</div>
+                    <div key={repo.key} className="containerPermalink">
+                        <div className="repoDetailsPermalink">
+                            <h3>{repo.name}</h3>
+                            <h5>{repo.description}</h5>
+                        </div>
+                        <div className="inPermalink">
+                            {arrayMapped}
+                        </div>
                     </div>
                 );    
             })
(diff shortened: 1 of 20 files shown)


martin-henz:
Sandy, can you take a look at this Issue 
https://github.com/source-academy/cadet-frontend/issues/609
and compare. Where are the gaps? Let's reach a consensus first.

src/components/Contributors.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+interface IContributorsArrayState {
+    ignoreRepos: string[];
+    ignoreContributors: string[];
+    repos: IReposState[];
+    contributors: IContributorsState[];
+}
+
+class Contributors extends Component<{}, IContributorsArrayState> {
+
```
geshuming: States suggest the data is mutable during the user's interaction with the page. Can we classify the above structs as `type` instead, for example 
```typescript
type Contributor = {
    ignoreRepos: string[],
    ignoreContributors: string[],
    repos: ReposType[],
    contributors: ContributorsType[]
};

type Repo {
    key: number,
    name: string,
    description: string,
    link: string
};
```

and add the following as constants
```suggestion
private ignoreRepos = ["sicp", "sharedb-ace-backend", "assessments", "tools", "source-academy2"];
private ignoreContributors: ["dependabot[bot]", "dependabot-preview[bot]"],
private repos: Repo[],
private contributors: Contributor[]
```

src/components/Contributors.tsx
```diff
@@ -0,0 +1,167 @@
...
+        fetch('https://api.github.com/orgs/source-academy/repos')
+        .then(results => {
+            return results.json();
+        })
+        .then(repos => {
+            const { ignoreRepos } = this.state;
+            const contributorLinks = repos
+                .filter((repo: any) => {
+                    for (const repoName of ignoreRepos) {
```
geshuming: can be simplified, and try to avoid using the `any` term

```typescript
const contributorLinks = repos
  .filter(repo => !(repo.name in ignoreRepos)
```

src/components/Contributors.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    }) 
+                );
+            })
+            .then(contributorsByRepo => {
+                return Promise.all(
+                    contributorsByRepo.map((contributors: any) => {
+                        const { ignoreContributors } = this.state;
+                        const contributorList = contributors
+                            .filter((contributor: any) => {
+                                for (const contributorName of ignoreContributors) {
```
geshuming: same as above `filter` simplification

src/components/Contributors.tsx
openorclose: That's not how the `in` operator works. Use `array.includes(item)` to check if array contains the item.

geshuming: Yep my bad, i was mistaken

sandydays: I have simplified the code within filter in my latest commit.

src/components/Contributors.tsx
sandydays: I have changed the structs from interface to type as suggested in my latest commit. Yet to figure out how to change parts of the code if I add ignoreRepos, ignoreContributors, repos, and contributors as constants instead of managing them within the state.

sandydays:
Shuming mentioned:
> Try to avoid the type `any`, the reason we use typescript is so we know what we props and functions an object will have.

I think some places in the code where I cycle through the array and use the type `any` is because at that point I am fetching an object in an array from the api endpoint, where each object has too many key-value pairs + attributes of each object might also change. So, I decided to stick with `any` type in those places. Is that alright?

But you are right, there are some places where I think have used the `any` type but it can be replaced by the types I have already declared at the top of the code. So, will work on fixing those soon.

martin-henz:
**team vs contributors**: "team" emphasizes the fact the we are working closely together, and "contributors" emphasizes that people can take part easily. I'd vote for "contributors". That way it's more inviting to newcomers.

martin-henz:
**a little stretched...*: I agree. One way to handle is to have two columns, similar to the **Facilitators** of modules in LumiNUS. Also worth considering: we could put the name and number of commits to the right of the picture, instead of under it.

src/components/Contributors.tsx
geshuming: You can directly refer to the constants like `this.ignoreRepos` and `this.repos = ...`.

We shouldn't store them as states if they need not be.

geshuming:
> **a little stretched...*: I agree. One way to handle is to have two columns, similar to the Facilitators of modules in LumiNUS. Also worth considering: we could put the name and number of commits to the right of the picture, instead of under it.

[Yes we can use css-grid for this! ](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout)

sandydays:
I have tried to make the popup look better with the latest commit:

1. Made the dialog box wider as well as adjusted the image size of each contributor and the width of their name and number of commits. So, now the contributors are arranged row by row with 4 contributors per row. 
\
@martin-henz suggested that we split it into 2 columns, so that would mean 2 contributors per row? I felt that since there are so many, might as well have 4 per row to make it easier to see more contributors at first glance and have lesser scrolling. I don't know though, which one do you all think is better? 
\
Also, as @martin-henz mentioned, we could put the gitHub username and number of commits to the right of the photo which I don't mind as well. Do we go for this? 

2. Changed "Team" to "Contributors" where necessary as it was voted to be the better option.

With my minimal knowledge of css and sass, I have experimented quite a bit and decided certain things like how many pixels something should be, etc. I am not too sure about issues of how the same thing would appear on different browsers/devices, as well as if the way I have created different classes for different container elements is fine. Any suggestions on that as well later when cleaning code would be helpful as well!

sandydays:
@rrtheonlyone I have no idea why the `package-lock.json` file got changed when I made my first commit. I did `git add .` so I didn't realize at that point that the file had changed. I saw it and it seems two lines got deleted for some reason. Added back those two lines, ran the code, everything still worked fine. 

So, what do I do now with regards to that?

geshuming:
> @geshuming I used css display: inline-block instead of grid, but I can look into grid as well if you need me to use that!

Yep, we can use css-grid to display 4 contributors per row, and then have the dialog box width fit the content, instead of using a fixed width.

src/components/Contributors.tsx
sandydays: Yes agreed, I have made the necessary changes in the latest commit.

martin-henz:
> @sandydays I would recommend against using redux to handle contributor data. You are just complicating the data fetching process without any benefit of using the redux store.
> 
> I would recommend you do an initial fetch when the `Application` performs `componentDidMount()`, store the data as a constant and pass down the prop to the `About` component. No need for redux at all.

Agree. The "single source of truth" is GitHub, here.

PR #621, opened 2019-06-24
Remove exports from list.js file

### LINKED ISSUE #487: Why is there an export here?
https://github.com/source-academy/cadet-frontend/blob/c93fb65f26159d702e012632347cfc01f752b777/public/externalLibs/list.js#L352

rrtheonlyone:
Fixes #487 

Right now, when we run Source Academy and open the console, we will see this error:
Screenshot 2019-06-24 at 10 16 55 PM

This is because of the `exports` in the `list.js` file. This PR removes it as it is not needed.

### DIFF
--- a/public/externalLibs/list.js
+++ b/public/externalLibs/list.js
@@ -348,5 +348,3 @@ function set_tail(xs, x) {
     throw new Error('set_tail(xs,x) expects a pair as argument xs, but encountered ' + xs)
   }
 }
-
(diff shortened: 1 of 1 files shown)


martin-henz:
Yes, don't know where this came from in the first place. Good to fix it.

martin-henz:
Fixes: https://github.com/source-academy/cadet-frontend/issues/487

Issue #608, opened 2019-06-20
Making the Frontend more Developer-Friendly and Maintainable


rrtheonlyone:
As I was deploying, I realised there are some aspects which can perhaps be improved upon. This issue will serve to document some of these possible improvements and track their progress.

Any sugggestions or opinions can also be listed below. The aim is to make this project easy to setup and work with for newer developers.

Current Improvements that can be done (in my opinion):
- [ ] Documentation : improve `README.md` + seperate `/docs` folder for more information
- [ ] Upgrade outdated libraries 
- [ ] Check CI and make sure its well integrated
- [ ] `package.json` cleanup - Remove some of the unnecessary commands and bash scripts
- [x] Update links to Forum etc.
- [ ] Look into linting - `npm run format` looks like it only runs prettier

geshuming:
tslint should have default severity set to warning level so that compilation will not fail while running the dev build. Of course, tslint will be set to error level for prepush check and CI.

geshuming:
We can consider migrating to eslint as [tslint is getting deprecated](https://medium.com/palantir/tslint-in-2019-1a144c2317a9). 

[How to setup eslint for a react app in typescript](https://dev.to/benweiser/how-to-set-up-eslint-typescript-prettier-with-create-react-app-3675)

However, `typescript-eslint` currently supports `typescript@>=3.2.1 <3.6.0` .

jiachen247:
Whats the status of this? @rrtheonlyone @geshuming

chownces:
Closing this as the current CI and pre push hook addresses this.

PR #628, opened 2019-06-28
Clean up package.json

geshuming:
Removed a bunch of unused dependencies, and bumping js-slang to 0.2.2.

I don't really have a clue what to do with the scripts though. @rrtheonlyone would you like to check them out?

Part of #608


geshuming:
The coverage fix script seems like it is supposed to stash `*.d.ts` files and unstash them. I don't think this script is needed at all

martin-henz:
Not sure about the status of this issue. Is it worth resolving the conflicts?

martin-henz:
@alcen : any views on this? Should we try to see if there is anything useful in this PR?

alcen:
> 
> 
> @alcen : any views on this? Should we try to see if there is anything useful in this PR?

I think that the branch is too outdated already. Probably better to re-make these changes on the current master instead

martin-henz:
> > @alcen : any views on this? Should we try to see if there is anything useful in this PR?
> 
> I think that the branch is too outdated already. Probably better to re-make these changes on the current master instead

OK. Can you see what you can recover from this PR and put it in a fresh one?

PR #630, opened 2019-06-29
URLs updated

martin-henz:
URLs of piazza, textbook and js-slang doc updated


geshuming:
Hi prof, these commits are already merged into the master.

PR #633, opened 2019-06-29
Frontend Testing: Preparing for Saga Tests

cheongsiuhong:
Part of #581

Changes:

1.  Added redux-saga-test-plan@4.0.0-beta3
2.  Restructured ./src/sagas/index.ts into separate files to ease testing.
3.  Added a test for the playground saga.
    - Tests if the saga appropriately puts changeQueryString upon taking GENERATE_LZ_STRING 
This is more or less a simple test for further sagas testing reference.


package.json
```diff
@@ -69,6 +69,7 @@
     "redux": "^3.7.2",
     "redux-mock-store": "^1.5.1",
     "redux-saga": "^1.0.0",
+    "redux-saga-test-plan": "^4.0.0-beta.3",
```
geshuming: Can this be saved as dev dependencies?

`npm install --save-dev redux-saga-test-plan`

cheongsiuhong: Requested changes are made.

PR #635, opened 2019-07-01
starting to use jsdoc for our libraries

martin-henz:
How to use:

npm run jsdoc

The files are placed in a folder doc/jsdoc

The script jsdoc.sh controls which files are examined for jsdoc annotations. Currently, we use the sound library.

### DIFF
--- a/scripts/jsdoc.sh
+++ b/scripts/jsdoc.sh
@@ -1,0 +1,22 @@
+#! /usr/bin/env bash
+
+SOURCEFOLDERS="public/externalLibs/sound"
+JSDOC="node_modules/.bin/jsdoc"
+
+main() {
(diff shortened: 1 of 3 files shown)


martin-henz:
Addresses: https://github.com/source-academy/cadet-frontend/issues/636

scripts/jsdoc.sh
```diff
@@ -0,0 +1,22 @@
+#! /usr/bin/env bash
```
rrtheonlyone: Why is there a need for a bash script? I believe we can call `jsdoc -r -d doc/json public/externalLibs/*.js` from package.json itself.

martin-henz:
Moving this to js-slang

PR #640, opened 2019-07-01
Frontend Testing: Playground Sagas

cheongsiuhong:
Part of #581 

Changes:
1. Added types to ./src/sagas/playground.ts
2. Added another test to /src/sagas/__tests__/playground.ts
    - Now also tests if the put action contains the correct string with a dummy editor value.


src/sagas/__tests__/playground.ts
```diff
@@ -1,18 +1,66 @@
...
+} from '../../reducers/states';
 import playgroundSaga from '../playground';
 
-test('Playground puts changeQueryString action upon taking GENERATE_LZ_STRING', () => {
-  return (
-    expectSaga(playgroundSaga)
-      .withState(defaultState)
-      // The default query string generated by defaultState is undefined.
-      .put(actions.changeQueryString(undefined))
+describe('Playground saga tests with the default and a dummy case', () => {
```
rrtheonlyone: The describe is a noun and the test are verbs! So something like this:

```
Describe (noun): my beverage

Test (verb): is sour

Test (verb): is not flat

Test (verb): is carbonated
```

src/sagas/playground.ts
```diff
@@ -4,23 +4,32 @@ import { SagaIterator } from 'redux-saga';
...
-          chap: chapter,
-          ext: external
-        });
-  yield put(actions.changeQueryString(newQueryString));
+  const code: string | null = yield select(
+    (state: IState) => state.workspaces.playground.editorValue
+  );
+  if (code === null || code === '' || code === defaultEditorValue) {
+    yield put(actions.changeQueryString(undefined));
+  } else {
```
rrtheonlyone: remove the else block and do a return above

src/sagas/playground.ts
```diff
@@ -4,23 +4,32 @@ import { SagaIterator } from 'redux-saga';
...
-      : qs.stringify({
-          prgrm: compressToEncodedURIComponent(code),
-          chap: chapter,
-          ext: external
-        });
-  yield put(actions.changeQueryString(newQueryString));
+  const code: string | null = yield select(
+    (state: IState) => state.workspaces.playground.editorValue
+  );
+  if (code === null || code === '' || code === defaultEditorValue) {
```
rrtheonlyone: I believe you can do:

`if (!code || code === defaultEditorValue)`

src/sagas/playground.ts
```diff
@@ -4,23 +4,32 @@ import { SagaIterator } from 'redux-saga';
...
+    yield put(actions.changeQueryString(undefined));
```
rrtheonlyone: What does this yield do?

cheongsiuhong: It's needed to reset the queryString in case all the code in the editor is deleted.

src/sagas/__tests__/playground.ts
cheongsiuhong: Changes are made! Please check if the new test descriptions are appropriate

cheongsiuhong:
> Looking Good 
> 
> Could you add in one more case for when empty string is passed in as `editorValue`?

Done!

PR #648, opened 2019-07-02
Adding chat component

wardetu:
Integrated Chatkit to serve chats that will replace the current comment field in all assignments. Chatkit's documentation can be found [here](https://pusher.com/docs/chatkit). For Chatkit to work, it must be used with backend. 

What the chat looks like:
chatview

Changes made:
1. New folder `Chat` in `components`, where all Chatkit related files are stored
2. New file `ChatContainer.ts` in `containers`
3. New file `_chat.scss` in `styles`
4. Edited `.env-sample` to include option to disable Chatkit.
5. Add `jwt-decode` library for chat JWT decoding.

Todo: 
1. Rewrite `ChatApp.js` in TypeScript once this [PR](https://github.com/pusher/chatkit-client-js/pull/180) is merged

New issues will be created for the items on the todo list.


src/components/chat/Input.tsx
```diff
@@ -0,0 +1,62 @@
...
+  public render() {
+    const sendButtonOpts = {
+      className: 'chat-send-button',
+      fullWidth: true
+    };
+
+    return (
+      <div className="Input">
+        <div className="input-zone">
+          <textarea
```
geshuming: Possible to use blueprintjs Textarea? https://blueprintjs.com/docs/#core/components/text-inputs.text-area

package.json
```diff
@@ -43,10 +45,12 @@
     "flexboxgrid": "^6.3.1",
     "flexboxgrid-helpers": "^1.1.3",
     "js-slang": "^0.2.3",
+    "jwt-decode": "^2.2.0",
     "lodash": "^4.17.11",
     "lz-string": "^1.4.4",
     "moment": "^2.22.2",
-    "node-sass-chokidar": "^1.3.4",
+    "node-sass": "^4.12.0",
```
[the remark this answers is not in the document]
wardetu: I have to install `node-sass` every time I clone the frontend. Otherwise all the imports in `index.scss` can't be compiled to css.

Not sure if this just happens to my computer though.

src/components/chat/Input.tsx
wardetu: Yea, I've made the switch.

src/components/academy/grading/GradingWorkspace.tsx
```diff
@@ -270,6 +271,15 @@ class GradingWorkspace extends React.Component<GradingWorkspaceProps> {
         icon: IconNames.NINJA,
         body: <Markdown content={props.grading![questionId].question.content} />
       },
+      {
+        label: `Chat`,
+        icon: IconNames.CHAT,
+        body: USE_CHATKIT ? (
+          <ChatApp roomId={props.grading![questionId].grade.comment} />
```
geshuming: I think we should rename the `comment` field to something more meaningful. 

I recommended deprecating the comment field on the backend as well, because you might have unintended behaviors not accounted for. In this case, the action `SUBMIT_GRADING` is submitting an undefined `comment` to the backend, causing the backend to store "undefined" into the comment field on the database.

src/components/assessment/AssessmentWorkspace.tsx
```diff
@@ -21,6 +24,7 @@ import Workspace, { WorkspaceProps } from '../workspace';
 import { ControlBarProps } from '../workspace/ControlBar';
 import { SideContentProps } from '../workspace/side-content';
 import Autograder from '../workspace/side-content/Autograder';
+
```
geshuming: What are all these spaces? Can we remove them so that tslint can properly apply the rules, or group them in a way that it makes sense.

e.g.
```javascript
import ... from '../../utils/...';
import ... from '../../utils/...';
import ... from '../../utils/...';

import ... from '../workspace/...';
import ... from '../workspace/...';
import ... from '../workspace/...';
```

src/components/assessment/AssessmentWorkspace.tsx
```diff
@@ -327,6 +332,7 @@ class AssessmentWorkspace extends React.Component<
   }
 
   /** Pre-condition: IAssessment has been loaded */
+  // ADD CHAT COMPONENT
```
geshuming: Remove comment

package.json
```diff
@@ -106,6 +108,8 @@
     "@types/redux-mock-store": "^0.0.13",
     "@types/showdown": "^1.7.5",
     "@types/xml2js": "^0.4.3",
+    "@pusher/chatkit-client": "^1.5.0",
```
geshuming: This dep shouldn't be dev dependency, only the "@types" should be. 

Dev dependencies are dependencies that won't appear in production. "@types" are not needed because our typescript files will be compiled to js files.

src/components/academy/grading/GradingWorkspace.tsx
wardetu: Yea we leave the field as comment for now and see how the chat goes for this summer module.
Will change it afterwards.

flxffy: With respect to the deprecation of comment, Martin would like to have this tested on real users during the summer workshop (14 July). Hence, in the interest of time, we will not be deprecating comment in this pull request as that would require changes on both front and back end.

rrtheonlyone: So if I understand this correctly: Every time a TA grades the assignment, the `comment` field will be reset. Then, when they reopen the page, a new room is created and the field updated?

May I know what is the plan for this in the next month? Are we gonna drop `comments` entirely and have a new field? As @geshuming has pointed out, this field is currently ambiguous and it feels like a hacky solution to get the chat out quickly. 

Furthermore, with regards to the workshop, we can deploy a version of a branch also! (it does not have to merged into master necessarily). It is better to do it right than postpone it to a later stage (there is a risk of nobody fixing it and the issue causing further problems down the road).

Let me know your thoughts on this.

package.json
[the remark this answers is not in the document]
rrtheonlyone: @wardetu what happens if your try on our current master branch? Do you need to install `node-sass` also? Or is it only when you are working on your fork for this PR.

src/components/academy/grading/GradingWorkspace.tsx
flxffy: Shuming pointed this out in our backend PR. I will fix it once I get off work. As for deprecating `comment`, we can have it done over the next week or so. However, it should not be expected to be ready before the workshop.

rrtheonlyone: Okay thats fine! Do log an issue for deprecating `comment` either in `cadet` or here for documentation purposes.

We can proceed to move on with this PR.

package.json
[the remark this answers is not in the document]
wardetu: I always have to install. Prolly this is because I'm on Windows.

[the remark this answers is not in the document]
geshuming: Try removing it and then doing a clean reinstall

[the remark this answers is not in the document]
wardetu: It worked.
I have updated `package.json` accordingly.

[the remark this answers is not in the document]
geshuming: Whenever you update dependencies, please commit the `package-lock` as well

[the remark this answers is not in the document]
wardetu: Okay!

geshuming:
> A minor gripe is that it is a little slow to load. (however, that cant be helped)

This is probably because the api call is made when the chat component is mounted (i.e. when you open the chat tab)

A suggestion is to move the api calls to the parent component (AssessmentWorkspace), and pass down the results to the chat component

rrtheonlyone:
I think its better to have it isolated for now! Let us not couple it with the parent just yet (till we are sure this is the route we want to take in the long term). The API calls are done by the `chatManager` and it may be difficult to shift it out to the parent.

The app is still usable for an end-user and I feel its not a major hindrance.

src/components/chat/ChatApp.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+  } // for scrolling
+
+  scrollToBottom = () => {
+    this.messagesEndRef.current.scrollIntoView({ behavior: 'smooth' });
+  };
+
+  componentDidMount() {
+    // tslint:disable-next-line:no-console
+    console.log(this.props.roomId);
```
rrtheonlyone: why is this committed?

wardetu: This helps the chat auto scroll to bottom everytime it is updated.

Elsewise we need to manually scroll down.

wardetu: This helps the chat auto scroll to bottom everytime it is updated.

Elsewise we need to manually scroll down.

rrtheonlyone: the `console.log`?

wardetu: Ah that one.

Aihui was looking at this issue: updating grade causes comment to be stored as undefined. I think she was debugging.

rrtheonlyone: I will fix this no worries!

flxffy: I've removed it sorry about that!

rrtheonlyone: No problem :) Thanks for helping to clean this up

PR #651, opened 2019-07-02
accumulate -> reduce

martin-henz:
fixes https://github.com/source-academy/js-slang/issues/258


martin-henz:
Another look at it led me to decide to keep accumulate. Neither accumulate nor reduce indicate the direction of the operation. Usually, the left-to-right variant is the dominant one. Its somewhat more intuitive and also has a natural iterative implementation.

rreduce and lreduce would have been a nice choice. But sticking with accumulate is the best, given the dominance of accumulate (right-to-left) in the textbook.

PR #656, opened 2019-07-04
Add notifications

geshuming:
Notifications.

Core features:
- Notifications appear on academy navigation bar.
notificationbar

- Notifications appear on assessment overviews.
notificationAssessment

- Notifications appear on grading overviews
notificationSubmission

- Notifications can be acknowledged explicitly by hovering over the notification icon on assessment/grading overviews.

- Notifications can be acknowledged implicity by clicking on the button to navigating to the specific assessment/submission.

- Notifications are currently fetched every time a user navigates within the academy. In the future, this can be replaced by long polling/web sockets.


jiachen247:
Can we have a UI/UX person look at this? Anyone knows anyone haha

jiachen247:
Would suggest having a button that launches a popup with all the notifs! Which can be used to easily clear and manage the notifications!

geshuming:
Integrated with chatkit

src/components/notification/NotificationBadge.tsx
```diff
@@ -0,0 +1,78 @@
+import { Intent, Popover, PopoverInteractionKind, Position, Tag } from '@blueprintjs/core';
+import * as React from 'react';
+import { AcademyNotification, AcademyNotificationType } from './notificationShape';
+
+type OwnProps = {
+  className?: string;
+  enableHover?: boolean; // enable or disable hover popover option
+  large?: boolean; // enable to use large style
+  notifications: AcademyNotification[];
```
rrtheonlyone: Why is this an array?

src/components/notification/NotificationBadge.tsx
```diff
@@ -0,0 +1,78 @@
...
+      >
+        {makeNotificationMessage(notification.type)}
+      </Tag>
+    );
+  };
+
+  return (
+    <Popover
+      className={props.className}
+      content={makeNotificationTags(props.notifications)}
```
rrtheonlyone: If the content is formed from an array of notifications that you map over, wont it have many messages stacked?

How will that look? Why not just have one message and the number?

src/components/notification/NotificationBadge.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <div className="col">{notifications.map(makeNotificationTag)}</div>
+  );
+
+  const makeNotificationTag = (notification: AcademyNotification) => {
+    const onRemove = () => props.handleAcknowledgeNotification([notification.id]);
+
+    return (
+      <Tag
+        className="row"
+        style={{ backgroundColor: 'transparent', marginLeft: '0.4rem', marginRight: '0.4rem' }}
```
rrtheonlyone: Why is style here instead of in a stylesheet like the other styles?

src/components/notification/__tests__/NotificationBadge.tsx
```diff
@@ -0,0 +1,84 @@
...
+    assessment_id: 1,
+    assessment_type: 'Mission',
+    assesssment_title: 'The Secret to Streams',
+    submission_id: 1
+  }
+];
+
+const handleAcknowledgeNotification = (ids: number[]) => null;
+
+describe('Badge', () => {
```
rrtheonlyone: The snapshots for these tests don't even give any useful information. How are these tests going to be useful?

For example, I cannot even see the notifications data (since they are prob triggered on popover)

src/components/academy/NavigationBar.tsx
```diff
@@ -24,6 +31,12 @@ const NavigationBar: React.SFC<NavigationBarProps> = props => (
       >
         <Icon icon={IconNames.FLAME} />
         <div className="navbar-button-text hidden-xs">Missions</div>
+        <NotificationBadge
```
rrtheonlyone: You should consider exporting a simple badge in your notifications file that just displays a number. Or you could just put `<Tag></Tag>` here since all we want is the number. 

I find it pointless that your:
- filtering and passing in all your notifications to the component
- the component is running a map on your component and processing them

And eventually they wont even be shown since `enableHover` is false?

src/components/academy/grading/index.tsx
```diff
@@ -275,6 +308,24 @@ class Grading extends React.Component<IGradingProps, State> {
     }
     this.gridApi.exportDataAsCsv({ allColumns: true });
   };
+
+  private preProcessData = () =>
```
rrtheonlyone: How will this preprocessing help? Also what does it do?

Try not to use generic names like `preprocessData`! It does not allow an external reader to understand what is going on. (add a comment at the very least)

src/sagas/backend.ts
```diff
@@ -275,6 +276,80 @@ function* backendSaga(): SagaIterator {
...
+    const tokens = yield select((state: IState) => ({
+      accessToken: state.session.accessToken,
+      refreshToken: state.session.refreshToken
+    }));
+
+    if (!tokens.accessToken) {
+      return;
+    }
+
+    const resp: Response | null = yield request('notification', 'GET', {
```
rrtheonlyone: Why not follow convention and have a separate function to call the request? (similar to our other sagas)

src/sagas/backend.ts
```diff
@@ -275,6 +276,80 @@ function* backendSaga(): SagaIterator {
...
+      accessToken: tokens.accessToken,
+      refreshToken: tokens.refreshToken,
+      shouldAutoLogout: false
+    });
+    if (resp && resp.ok) {
+      const newNotifications: AcademyNotification[] = ((yield resp.json()) as any[]).map(n => {
```
rrtheonlyone: if you use a separate function and bind the result to a variable, you can avoid `yield resp.json()) as any[]`

src/sagas/backend.ts
```diff
@@ -275,6 +276,80 @@ function* backendSaga(): SagaIterator {
...
+        return {
+          id: n.id,
```
rrtheonlyone: Dont use `n`. Its better to be explicit (even if its a map)

src/sagas/backend.ts
```diff
@@ -275,6 +276,80 @@ function* backendSaga(): SagaIterator {
...
+    }));
+    const ids = (action as actionTypes.IAction).payload as number[];
+    const notifications: AcademyNotification[] = yield select(
+      (state: IState) => state.session.notifications
+    );
+    const newNotifications: AcademyNotification[] = notifications.filter(
+      notification => !ids.includes(notification.id)
+    );
+    yield put(actions.updateNotifications(newNotifications));
+    const resp: Response | null = yield request(`notification/acknowledge`, 'POST', {
```
rrtheonlyone: A separate function for this would be cleaner as well!

src/sagas/backend.ts
```diff
@@ -275,6 +276,80 @@ function* backendSaga(): SagaIterator {
...
+      yield put(actions.updateNotifications(notifications));
+    }
+  });
+
+  yield takeEvery(actionTypes.NOTIFY_CHATKIT_USERS, function*(action) {
+    const tokens = yield select((state: IState) => ({
+      accessToken: state.session.accessToken,
+      refreshToken: state.session.refreshToken
+    }));
+    yield request(`chat/notify`, 'POST', {
```
rrtheonlyone: Also this^

geshuming:
> Why is sorting necessary? I feel the sorting is introducing a ton of unneeded complexity?
We came to a conclusion that notifications will be most effective if the assessment/submissions with notifications are sorted to the top. If not, users may have to scroll to find assessments. 

How are the chatkit notifications triggered?
> There is a callback on message send that posts a request to the backend, creating a notification.

src/components/notification/NotificationHelpers.ts
```diff
@@ -0,0 +1,17 @@
+import { AcademyNotification } from './notificationShape';
```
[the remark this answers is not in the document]
geshuming: This is for filtering notifications, not the assessments/submissions. The frontend receives an array of all unread notifications from the backend, so the filtering here will push the notifications to the relevant places.

src/components/notification/NotificationBadge.tsx
geshuming: The badge will display the number of relevant notifications, hence the array.

src/components/notification/NotificationBadge.tsx
geshuming: I think it's good to differentiate the notifications, hence we can have multiple notifications per notification badge. I limited the popover for assessments/submissions only as those will have at most 4 associated notifications.

src/components/academy/NavigationBar.tsx
geshuming: Noted. I will disable processing of the tags if enableHover is false

src/components/notification/notificationShape.ts
```diff
@@ -0,0 +1,23 @@
+import { AssessmentCategory } from '../assessment/assessmentShape';
+
+export type AcademyNotification = {
```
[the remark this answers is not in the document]
geshuming: Renamed to `Notification`

src/components/notification/notificationShape.ts
```diff
@@ -0,0 +1,23 @@
...
+  assessment_id?: number;
+  assessment_type?: AssessmentCategory;
+  assesssment_title?: string;
+  id: number;
+  question_id?: number;
+  submission_id?: number;
+  type: AcademyNotificationType;
+};
+
+enum AcademyNotificationTypes {
```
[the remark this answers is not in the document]
geshuming: The backend passes the type as a string. In order to typecast from string to enum, we have to use `keyof typeof NotificationTypes`

src/components/academy/grading/index.tsx
geshuming: Renamed it to `sortSubmissions`

src/components/notification/__tests__/NotificationBadge.tsx
geshuming: I understand. I'm not sure how to test the popover so these are very basic tests

src/components/notification/NotificationBadge.tsx
geshuming: Moved to stylesheet

src/sagas/backend.ts
geshuming: Refactored accordingly

src/components/notification/notificationShape.ts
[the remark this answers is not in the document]
rrtheonlyone: Yup. I meant why not export this and use it in other areas. e.g. instead of 

```
switch (type) {
    case 'new':
```

you could just do `AcademyNotificationTypes.new`

rrtheonlyone:
Can you take me through the logic of your notifications? I believe I am still not clear on your implementation.

1. Why is it an array of notifications passed in when only one message is shown at the end of the day? (or are they stacked)

2. For things like navbar where only the number is going to be shown, why go through the effort of using this component? (and have more attributes on it like `enableHover`) etc.

geshuming:
> 1. Why is it an array of notifications passed in when only one message is shown at the end of the day? (or are they stacked)

Yep they can be stacked. Check out the closed mock assessment, where it has 2 notifications. 

> 2. For things like navbar where only the number is going to be shown, why go through the effort of using this component? (and have more attributes on it like `enableHover`) etc.

I thought that it would be better if we use the same component. If not, I think we can separate the component so we can have `BadgeWithPopover` and `Badge`, where `BadgeWithPopover` renders `Badge ` as well?

src/components/notification/NotificationBadge.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <Tag intent={Intent.DANGER} round={true} large={props.large}>
+        {props.notifications.length}
+      </Tag>
+    </Popover>
+  );
+};
+
+const makeNotificationMessage = (type: AcademyNotificationType) => {
+  switch (type) {
+    case 'new':
```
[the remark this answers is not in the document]
geshuming: Yep done

src/components/academy/NavigationBar.tsx
```diff
@@ -33,6 +46,12 @@ const NavigationBar: React.SFC<NavigationBarProps> = props => (
       >
         <Icon icon={IconNames.LIGHTBULB} />
         <div className="navbar-button-text hidden-xs">Quests</div>
+        <NotificationBadge
+          notifications={props.notifications.filter(
+            n => n.type !== 'submitted' && n.assessment_type && n.assessment_type === 'Sidequest'
```
geshuming: This is confusing, refactor to differentiate whether a notification belongs to an assessment or submission

src/components/academy/NavigationBar.tsx
```diff
@@ -42,6 +61,12 @@ const NavigationBar: React.SFC<NavigationBarProps> = props => (
       >
         <Icon icon={IconNames.PREDICTIVE_ANALYSIS} />
         <div className="navbar-button-text hidden-xs">Paths</div>
+        <NotificationBadge
+          notifications={props.notifications.filter(
+            n => n.type !== 'submitted' && n.assessment_type && n.assessment_type === 'Path'
```
geshuming: Same here

src/components/academy/NavigationBar.tsx
```diff
@@ -51,6 +76,12 @@ const NavigationBar: React.SFC<NavigationBarProps> = props => (
       >
         <Icon icon={IconNames.COMPARISON} />
         <div className="navbar-button-text hidden-xs">Contests</div>
+        <NotificationBadge
+          notifications={props.notifications.filter(
+            n => n.type !== 'submitted' && n.assessment_type && n.assessment_type === 'Contest'
```
geshuming: Same here

src/components/academy/NavigationBar.tsx
```diff
@@ -62,6 +93,10 @@ const NavigationBar: React.SFC<NavigationBarProps> = props => (
         >
           <Icon icon={IconNames.ENDORSED} />
           <div className="navbar-button-text hidden-xs">Grading</div>
+          <NotificationBadge
+            notifications={props.notifications.filter(n => n.type === 'submitted')}
```
geshuming: Same here

src/sagas/backend.ts
```diff
@@ -277,6 +278,56 @@ function* backendSaga(): SagaIterator {
...
+      refreshToken: state.session.refreshToken
+    }));
+    const ids = (action as actionTypes.IAction).payload as number[];
+    const notifications: Notification[] = yield select(
+      (state: IState) => state.session.notifications
+    );
+    const newNotifications: Notification[] = notifications.filter(
+      notification => !ids.includes(notification.id)
+    );
+    yield put(actions.updateNotifications(newNotifications));
```
geshuming: Comment here to indicate why notifications are updated before the server side is done

src/sagas/backend.ts
```diff
@@ -277,6 +278,56 @@ function* backendSaga(): SagaIterator {
...
+
+    const resp: Response | null = yield postAcknowledgeNotification(tokens, ids);
+
+    if (resp && resp.ok) {
+      return;
+    } else {
+      yield call(showWarningMessage, "Something went wrong, couldn't acknowledge");
+      yield put(actions.updateNotifications(notifications));
+    }
```
geshuming: Refactor for happy path

rrtheonlyone:
> I thought that it would be better if we use the same component. If not, I think we can separate the component so we can have BadgeWithPopover and Badge, where BadgeWithPopover renders Badge as well?

I think separating them is better. The component will then be smaller and easier to use also. It also doesnt feel clean to have an `enableHover` and an `undefined` content.

src/containers/notification/NotificationBadgeWithPopover.ts
```diff
@@ -0,0 +1,20 @@
...
+const mapDispatchToProps: MapDispatchToProps<DispatchProps, {}> = (dispatch: Dispatch<any>) =>
+  bindActionCreators(
+    {
+      handleAcknowledgeNotifications: acknowledgeNotifications
+    },
+    dispatch
+  );
+
+export default connect(
+  null,
```
rrtheonlyone: Do you need the null here?

src/sagas/backend.ts
```diff
@@ -277,6 +278,67 @@ function* backendSaga(): SagaIterator {
...
+    );
+    const newNotifications: Notification[] = notifications.filter(
+      notification => !ids.includes(notification.id)
+    );
+
+    /* 
+      For responsiveness, we update the store's state first, then
+      check if the server rejects the acknowledgement and undo the changes if necessary
+    */
+
```
rrtheonlyone: Here, I don't see the need to undo the changes. If the acknowledgement fails, when the notifications are fetched again from the backend, it will be there?

Furthermore, how slow is it really? Is it a noticeable difference?

src/sagas/backend.ts
```diff
@@ -277,6 +278,67 @@ function* backendSaga(): SagaIterator {
...
+    yield put(actions.updateNotifications(notifications));
+  });
+
+  yield takeEvery(actionTypes.NOTIFY_CHATKIT_USERS, function*(action) {
+    const tokens = yield select((state: IState) => ({
+      accessToken: state.session.accessToken,
+      refreshToken: state.session.refreshToken
+    }));
+
+    const assessmentId = (action as actionTypes.IAction).payload.assessmentId || null;
```
rrtheonlyone: why did you choose to change it to null here? Why not leave it as `undefined`?

I don't think we strictly adhere to this but even the [Typescript Styleguide](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined) advocates `undefined` over `null`.

src/sagas/backend.ts
```diff
@@ -574,6 +636,69 @@ async function postUnsubmit(submissionId: number, tokens: Tokens) {
...
+ */
+async function getNotifications(tokens: Tokens) {
+  const resp: Response | null = await request('notification', 'GET', {
+    accessToken: tokens.accessToken,
+    refreshToken: tokens.refreshToken,
+    shouldAutoLogout: false
+  });
+  let notifications: Notification[] = [];
+
+  if (resp && resp.ok) {
```
rrtheonlyone: You can use a guard clause here.

src/components/notification/__tests__/NotificationBadge.tsx
```diff
@@ -0,0 +1,67 @@
...
+];
+
+describe('Badge', () => {
+  test('renders properly for students', () => {
+    const tree = mount(<NotificationBadge number={studentNotifications.length} />);
+
+    expect(tree.debug()).toMatchSnapshot();
+  });
+
+  test('renders properly for avengers', () => {
```
rrtheonlyone: Why did you split by avengers and student? The component has no logic to differentiate between the roles?

I feel these tests are redundant (Blueprint have already probably tested their `Tag` component) There is no need for us to retest it here.

src/components/notification/NotificationHelpers.ts
```diff
@@ -0,0 +1,39 @@
+import { AssessmentCategories } from '../assessment/assessmentShape';
```
rrtheonlyone: Add tests for these functions to make sure the filtering behaviour is correct.

src/components/notification/NotificationHelpers.ts
```diff
@@ -0,0 +1,39 @@
...
+) {
+  return notifications.filter(
+    n =>
+      (opts.assessment_id === undefined || n.assessment_id === opts.assessment_id) &&
+      (opts.submission_id === undefined || n.submission_id === opts.submission_id)
+  );
+}
+
+/*
+  For filtering notifications on the academy navigation bar.
```
rrtheonlyone: Remove this line. Its a helper function so we dont really need to say where it is used in the function itself.

src/components/notification/NotificationHelpers.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Notification } from './notificationShape';
+
+type filterByIdOptions = {
+  assessment_id?: number;
+  submission_id?: number;
+};
+
+type filterByTypeOptions = AssessmentCategories | 'Grading';
+
+export function filterNotificationsById(
```
rrtheonlyone: Actually instead of going with this its simply cleaner to split into two functions which have a clear purpose:

`filterNotificationsByAssessment` and `filterNotificationsByAssessmentBySubmission`

src/components/notification/NotificationHelpers.ts
```diff
@@ -0,0 +1,39 @@
...
+*/
+export function filterNotificationsByType(
+  notifications: Notification[],
+  assessmentType: filterByTypeOptions
+) {
+  return notifications.filter(n => {
+    if (assessmentType === 'Grading') {
+      return n.submission_id !== undefined;
+    }
+    return n.submission_id === undefined && assessmentType === n.assessment_type;
```
rrtheonlyone: Why not `!n.submission_id && assessmentType === n.assessment_type`

src/components/notification/NotificationBadgeWithPopover.tsx
```diff
@@ -0,0 +1,74 @@
...
+      return 'Unknown notification';
+  }
+};
+
+const NotificationBadgeWithPopover: React.SFC<OwnProps & DispatchProps> = props => {
+  if (!props.notifications.length) {
+    return null;
+  }
+
+  const makeNotificationTags = (notifications: Notification[]) => (
```
rrtheonlyone: This parameter is redundant. Since your passing in `props.notifications` anyway, we can just use it here.

src/containers/academy/NavigationBar.ts
```diff
@@ -0,0 +1,10 @@
+import { connect, MapStateToProps } from 'react-redux';
```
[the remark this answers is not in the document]
geshuming: Changes in notifications state will trigger rerenders, which created bad UX when the state is connected to containers/academy/grading/index. 

I have a new idea, we can connect the notification state to the badge container, and pass a filter function as a prop to filter the notifications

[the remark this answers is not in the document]
rrtheonlyone: That does sound cleaner!

src/containers/notification/NotificationBadgeWithPopover.ts
geshuming: Yes. Without the null will invoke a different overloaded function.

geshuming:
> One thing I noticed you have done quite frequently it to use really long relative imports. You can always clean them up.

apparently jest could not resolve the absolute paths. Can you take a look at the configuration?

rrtheonlyone:
>apparently jest could not resolve the absolute paths. Can you take a look at the configuration?

Oh I see what you mean.

Okay I can take a look! Then let's leave it for now in this PR!

src/sagas/backend.ts
geshuming: I think it could become potentially noticeable. I guess I should revert it such that the notifications are updated on successful response?

src/components/notification/NotificationHelpers.ts
geshuming: The mocks will need to be updated. At first I thought there could be an issue when the id is 0 (which is falsey), but it wont be the case (database starts from 1)

src/sagas/backend.ts
rrtheonlyone: I just feel its messy to update and rollback when the response fails. I prefer either updating immediately (and let the notifications load again on fetch notification if updating fails) or updating on success.

geshuming: Ok, then it will update immediately

src/components/academy/__tests__/NavigationBar.tsx
```diff
@@ -5,7 +5,7 @@ import { Role } from '../../../reducers/states';
 import NavigationBar from '../NavigationBar';
```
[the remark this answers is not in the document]
geshuming: Not relevant anymore

src/components/assessment/__tests__/index.tsx
```diff
@@ -39,39 +43,47 @@ const mockPresentAssessmentForStudent: IAssessmentProps = {
```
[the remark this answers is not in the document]
geshuming: I'm not sure how to test this. Seems a bit difficult at this stage, maybe we can tackle this when we deal with testing

src/components/notification/__tests__/NotificationBadge.tsx
```diff
@@ -0,0 +1,61 @@
...
+    assessment_id: 1,
+    assessment_type: 'Mission',
+    assessment_title: 'The Secret to Streams'
+  }
+];
+
+const handleAcknowledgeNotifications = (ids: number[]) => null;
+
+describe('Badge', () => {
+  test('renders properly for students', () => {
```
rrtheonlyone: You can improve upon these tests! (once u add in the flags)

src/components/notification/__tests__/NotificationHelpers.ts
```diff
@@ -0,0 +1,124 @@
+import { AssessmentCategories } from '../../../components/assessment/assessmentShape';
+import * as NotificationHelpers from '../NotificationHelpers';
+import { Notification } from '../notificationShape';
+
+const notification1: Notification = {
```
rrtheonlyone: The testing code should also be readable. Dont use 1,2,3, to name your variables. Be explicit e.g. `newNotification` or `missionNotification`

src/components/notification/NotificationBadge.tsx
```diff
@@ -0,0 +1,90 @@
+import { Intent, Popover, PopoverInteractionKind, Position, Tag } from '@blueprintjs/core';
+import * as React from 'react';
+import { Notification, NotificationType, NotificationTypes } from './notificationShape';
+
+type OwnProps = {
+  className?: string;
+  disableHover?: boolean; // Set to true to disable popover content
+  large?: boolean; // Set to true to use large style
+  filterNotifications?: (notifications: Notification[]) => Notification[];
```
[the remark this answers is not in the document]
geshuming: The helper functions return a predicate function, which is then applied to the notification state in the component.

E.g. 
```javascript
<NotificationBadge filterNotifications={filterNotificationsByAssessment(assessment.id)} /> 
```

I think i will rename the prop to `filterPredicate` for clarity?

src/components/notification/NotificationBadge.tsx
```diff
@@ -0,0 +1,90 @@
...
+export type StateProps = {
+  notifications: Notification[];
+};
+
+export type DispatchProps = {
+  handleAcknowledgeNotifications: (notificationIds: number[]) => void;
+};
+
+const NotificationBadge: React.SFC<OwnProps & StateProps & DispatchProps> = props => {
+  const notifications = props.filterNotifications
```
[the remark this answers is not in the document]
geshuming: I think the ternary op here is quite clear though, 

If filter function exists, filter, else dont

src/components/notification/NotificationBadge.tsx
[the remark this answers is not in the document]
rrtheonlyone: You can use those helper functions inside here as well. You don't need to pass in a whole function from everywhere to this.

You can just have flag like `filterBySubmission=<submission_id>` or something. Then filter within the badge itself.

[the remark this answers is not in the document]
geshuming: I'm not sure how this will help simplify or clarify the filtering process.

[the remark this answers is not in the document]
rrtheonlyone: Do you understand the change I am proposing?

src/components/notification/NotificationBadge.tsx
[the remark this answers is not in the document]
rrtheonlyone: I agree it is clear. However, personally I would only advocate ternary for really obvious cases. It does not take that much longer to write an `if` statement. We can leave this for now.

src/components/notification/NotificationBadge.tsx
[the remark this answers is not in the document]
geshuming: The helper functions generate the required filter function depending on the props of other components. I think this makes it more readable?

If not, let me see if I get what you meant:
1. We'll add new flags to the NotificationBadge props
```javascript
type OwnProps = {
  className?: string;
  disableHover?: boolean; // Set to true to disable popover content
  large?: boolean; // Set to true to use large style
  filterBySubmission?: number;
  filterByType?: filterByTypeOptions;
  filterByAssessment?: number;
}
```

2. And then we'll filter the notifications.
```javascript
let notifications = props.notifications;

if (filterByAssessment) {
  notifications = filterNotificationsByAssessment(filterByAssessment)(notifications);
}

if (filterBySubmission) {
  notifications = filterNotificationsBySubmission(filterBySubmission)(notifications);
}

if (filterByType) {
  notifications = filterNotificationsByType(filterByType)(notifications);
}

...
```

Is that what you are suggesting?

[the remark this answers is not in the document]
rrtheonlyone: Yes that is correct. You can put it under a separate function `filterNotifications` in the component itself. 

Everything is now handled by the notification component itself and users of this component only need to worry about what number/type to pass in. 

In your current implementation, they have to write a separate function or reference the helper functions to use it. Its clunky and makes the component hard to use. Yours is more generic but considering you only need `filterNotificationsByAssessment` /  `filterNotificationsBySubmission` /  `filterNotificationsByType` passing in an entire function as props seems unnecessary.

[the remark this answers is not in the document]
rrtheonlyone: Actually, give me a little time, I need to think about it. I see where you are coming from as well!

Issue #616, opened 2019-06-21
DEBUG and ENV: should only be activated for Source §3,4


martin-henz:
The environment model is needed for understanding this, so we should limit the use to Source §3 or later.

One "easy" way may be to allow for gutter click only when Source §3 or §4 are active.

podocarp:
However, I feel like a debugger is useful regardless of where you are in the module. Perhaps we could enable the environment visualiser with 3 and 4 but leave the inspector as it is?

martin-henz:
The current INSPECT tool also uses the environment model. We are working on an inspector that operates on the substitution model. Yee Jian is working on it. We hope it can be ready in time.

The idea is to use the same entry point: A break point in the gutter of the Ace editor. If the user chooses Source §1 or §2, the substitution model inspector is used, and if the user chooses Source §3 or §4, the environment model inspector is used.

If a student is curious about the environment model before the recess, he can switch to Source §3 and play. There is nothing wrong with that. However, it's clear that its not covered in class yet. 

So yea, I differ from your view. I feel that the way to activate the INSPECT debugger is to choose Source §3 (or §4). Users who choose Source §1 or §2 should not see the debugger. (If Yee Jian succeeds, the user will see the SUBST inspector, and if he does not succeed, we will not have an inspector for Source §1 and §2.)

seanlowjk:
Fixed by #661

PR #661, opened 2019-07-04
Inspector update for #616

podocarp:
Simple change to disable the debugger until chapter 3 and 4.


martin-henz:
Travis complains. Not pretty enough.

martin-henz:
Once we switch to Source §1 or §2, the Inspector and Env Visualizer should be cleared. At the moment, they just sit there, with content that is not relevant any longer:

Screenshot 2019-07-05 at 6 17 03 AM

martin-henz:
Any other views on this, @ReignOfComputer and @geshuming ?

martin-henz:
Taking the telegram discussion a bit further, here are the options:

(A) Restrictive

So the correct solution would be to disable change of Source x and change of library while debugging is going on, by:
(1) removing the buttons
(2) making the buttons grey and unclickable, or
(3) letting the user click and choose, but then pop up an error saying stop inspector to switch.

(B) Permissive

Let the user click and choose Library and Source chapter, but for any selection, automatically leave INSPECT and ENV, and re-set REPL. I personally find this option attractive. Any change of Library and Source chapter already resets REPL.

martin-henz:
TL'DR: For any of the following actions, let's re-set REPL, ENV and INSPECT:
* Run
* change library
* change Source §x

PR #655, opened 2019-07-04
Contributors page new

sandydays:
Part of #609:

- Reorganized code from #612 - split fetching, types, rendering of list of contributors, and extra details about contributors into different files
- Contributors now displayed on a page with route `/contributors`, accessed from Contributors button on the Navbar
- Content of About popup in the dropdown updated
- Required snapshots updated

PR #668, opened 2019-07-05
Contributors page new 2

sandydays:
Part of #609:

- Very small change in text of the page
- Follows from now obsolete #655
- Can use this pr to discuss any other changes needed in code


geshuming:
Actually, can we get the contributors page to be consistent in style with the rest of source academy, aka white background box, on top of a spaceship background. And put in a bit more styling too

martin-henz:
Oh, I just noticed: We don't mention github on the Contributors page! I think there should be a link to the repo instead of just the name of the repo, and the text should mention the Github organisation Source Academy that hosts all the repositories.

sandydays:
> Actually, can we get the contributors page to be consistent in style with the rest of source academy, aka white background box, on top of a spaceship background. And put in a bit more styling too

I have addressed this in the latest commit. Still retained the same dark colours even though I pushed the whole page into a white card - any further suggestions on styling or if we want to change background and text colours please do let me know and I can make the changes.

martin-henz:
@sandydays : Can you post a screenshot of the new layout?

sandydays:
> @sandydays : Can you post a screenshot of the new layout?

This is how the page currently looks from the latest commit:

ezgif com-video-to-gif (1)

martin-henz:
I think that's an improvement.

Just curious: Currently, the content has a white background, which is good because it is consistent with the mission page. the top part of the page has a light grey field inside the white background, and dark grey letters on top. The lower part has black fields inside the white background, and dark grey fields inside those black fields. The lower part has white or light grey letters on top.

As a result, the lower part looks quite complicated, and the contrast black-on-white is very stark.

Have you considered the styling of the missions page: white background, white fields with a very thin border, and dark letters?

sandydays:
Yes, we have some options if we have a light background with dark text, the slight border comes with using the card from blueprintjs.

Option 1: White on white
ezgif com-video-to-gif_white on white

Option 2: Grey on white 
ezgif com-video-to-gif_grey_on_white

Option 3: White on grey
ezgif com-video-to-gif_white_on_grey

Personally, I prefer option 3 if we want to go in this direction, even if that means not being consistent with the missions white background. But that is my personal opinion and if you all have any other suggestions do let me know. The other alternative is dark background with light text as before.

geshuming:
I think option 2 looks good

martin-henz:
I'm indifferent between 2 and 3, and prefer these two over 1. Whatever we choose, can we apply the same to the Missions pages?

rrtheonlyone:
@martin-henz I dont think we should change all our mission pages! Shall we just adopt the same style of the missions to the contributors page instead?

I think option 2 looks good as well! Shall we move with that? @geshuming @alcen any other UI suggestions are welcome as well :)

I think I have some suggestions for the code itself. However, for starters lets fix the merge conflicts first @sandydays

martin-henz:
I'm ok with any of the options. They all look great! Regarding uniformity: The two use-cases are I guess different enough that we can apply slightly different styling. No need to shoehorn all into the same stylesheet. So yea, I'm ok with what you guys come up with.

alcen:
Personally I would have preferred option 3, followed by 2, then 1. I think the grey contrasts with the background slightly better than the white. But for consistency 2 is the best.


> The two use-cases are I guess different enough that we can apply slightly different styling.

I agree with this, I doubt students would spend much time on this page as compared to the missions. It will probably be viewed once in the semester and then forgotten about haha

sandydays:
>Personally I would have preferred option 3, followed by 2, then 1. I think the grey contrasts with the background slightly better than the white. But for consistency 2 is the best.

Yes, that was also why I was leaning towards option 3. At the moment, what has been merged is option 2, but can easily be changed to option 3. As for issues with consistency, I think this page looks quite similar to the missions page in lot of aspects so a slight variation of grey background shouldn't be much of an issue. 

Anyway, would be great if we can reach a consensus on this. Let me know what is the final call!

martin-henz:
The final call should be yours, I feel, after all the hard work!

PR #638, opened 2019-07-01
Improve UI of grading page

geshuming:
General improvements of the UI of ag-grid:

- Applied theme where there was none previously

- Shifted some columns for better grouping

- Auto-size the columns on window resize

Before:
currenttheme

After:
newtheme

Issue #351, opened 2018-09-04
Enhanced grading status


blackening:
Allow sorting/filtering by group and grading status.

This is for administrative purposes.

Pending https://github.com/source-academy/cadet/issues/260, https://github.com/source-academy/cadet/issues/261, https://github.com/source-academy/cadet/issues/262.

martin-henz:
All done.

PR #670, opened 2019-07-06
Grading overview page overhaul

Aulud:
## Grading overview page overhaul
Summary: Improve the UI of the grading overview page and provide more more granular information on each submission.

### Related PRs
Please review this PR in conjunction with the PR on cadet: https://github.com/source-academy/cadet/pull/408

### Changelog
  - Add `excluded` value for `GradingStatus` enum for assessments not (yet) requiring manual grading
  - Add `gradingStatus`, `questionCount` and `gradedCount` attributes to `GradingOverview` type
  - Edit `getGradingOverviews` in `backendSaga` to retrieve grading status details
  - Split `GradingHistory` component into separate (tooltip) components for displaying grade, XP and grading status details (addresses #351) for submissions
  - Correct ag-grid CSS stylesheet import and improve styling
  - Collate all control buttons above the datagrid
  - Fix unnecessary vertical scrollbar
  - Shuffle and resize columns in grid (from #638)
  - Auto-size columns on window resize (from #638)
  - Suppress blue border around selected cell and sub-components in datagrid (suggestion from #638)
  - Edit the grading navigation link to use a full-width button as with unsubmit button (as per @geshuming's suggestion - thanks!)
  - Update mocks

#### Before
grading-overview-old

#### After (verified after merging master #601)
grading-overview-new-3

#### Tooltips for XP, grade and grading status
Tooltip-preview

#### Grading status tooltip types
Grading-status-tooltip

### Amended grading navigation button (on hover)
grading-overview-new-edit

### Other browsers
Verified to work with Mozilla Firefox 67.0.4 (64-bit)

image

Last updated 9 Jul 2019, 11:15PM


martin-henz:
I discussed this with Keven just now. Maybe it's best if we postpone this PR until after the deployment scheduled for 7/7. Then we could take a closer look.

It's not needed for SWS3012, which will only have 40 students.

What do you all think?

src/components/academy/grading/GradingHistory.tsx
```diff
@@ -23,29 +26,59 @@ class GradingHistory extends React.Component<GradingHistoryProps, {}> {
   }
```
rrtheonlyone: It is better to split this component into three separate components. Since they are not related to each other, there is no reason to put them here and do a `if/else` chain to select one of them. 

It is also more scalable in future.

Aulud: Agreed; have split the component into three separate components `GradeTooltip`, `GradingStatusTooltip` and `XPTooltip`. Thanks!

src/styles/_academy.scss
```diff
@@ -200,34 +200,71 @@
...
 
-.ag-header-cell-label {
-  justify-content: center;
-  font-weight: bold;
-  padding-left: 15px;
+.ag-grid-controls {
+  display: flex;
+  justify-content: space-between;
+
+  .pt-icon {
```
geshuming: We just bumped to blueprint 3. Use `.#{$ns}.icon`

src/styles/_academy.scss
```diff
@@ -200,34 +200,71 @@
    */
   .ag-body-container {
     display: block;
+    /* Override ag-grid style to suppress unnecessary vertical scrollbar (bugfix) */
+    margin-bottom: 0;
   }
-  .ag-grid-export-button {
-    margin-top: 20px;
+
+  .pt-icon {
```
geshuming: Same here

src/styles/_academy.scss
Aulud: Have edited the stylesheet for this; thanks!

rrtheonlyone:
@martin-henz @geshuming can we come to a consensus on point 3 above? Do we want the grading to open in new tab or current window?

@Aulud have u taken a look at point 2? Do update the mock backend so its easier to test without running the backend!

geshuming:
I think we should open grading in current window. Some people may prefer doing grading 1 by 1, and others may prefer opening all the submissions. The latter can be achieved by Ctrl + Click

martin-henz:
> @martin-henz @geshuming can we come to a consensus on point 3 above? Do we want the grading to open in new tab or current window?
> 
> @Aulud have u taken a look at point 2? Do update the mock backend so its easier to test without running the backend!

I think @blackening has a workflow in mind for this. Eldric, you'd prefer if we open in new tab, right? I'm ok with it, even if the count is not right in the new tab. The workflow would be: open in new tab, do the grading, go back to the original and refresh, right?

I think if we write this down in the Avenger handbook, we'll be ok.

martin-henz:
> I think we should open grading in current window. Some people may prefer doing grading 1 by 1, and others may prefer opening all the submissions. The latter can be achieved by Ctrl + Click

Oh, good point. OK, same window then. @blackening can then show the Avengers to Ctrl + Click way of speed-grading :-)

blackening:
> > @martin-henz @geshuming can we come to a consensus on point 3 above? Do we want the grading to open in new tab or current window?
> > @Aulud have u taken a look at point 2? Do update the mock backend so its easier to test without running the backend!
> 
> I think @blackening has a workflow in mind for this. Eldric, you'd prefer if we open in new tab, right? I'm ok with it, even if the count is not right in the new tab. The workflow would be: open in new tab, do the grading, go back to the original and refresh, right?
> 
> I think if we write this down in the Avenger handbook, we'll be ok.

I think either way is fine, so long as the avenger is aware that opening in new tab doesn't sync across tabs.

In the current version of SA, i simply ctrl click all the items i want open. That opens them in the new tabs.

martin-henz:
Yep!

Aulud:
> @martin-henz @geshuming can we come to a consensus on point 3 above? Do we want the grading to open in new tab or current window?
> 
> @Aulud have u taken a look at point 2? Do update the mock backend so its easier to test without running the backend!

Hi @geshuming and @blackening, the button has been updated to load the current submission for grading in the current tab. CTRL+click opens the submission in another tab as expected.

Hi @rrtheonlyone, the present design of the mock backend relies upon loading mock information from a  `const` array or object defined within either `assessmentAPI.ts` or `gradingAPI.ts`. Notably,  `mockFetchGradingOverview` and `mockFetchGrading` return respectively the static array `mockGradingOverviews` and `mockGrading` defined in `gradingAPI.ts`. This contrasts with the Elixir backend where most of the information displayed in the grading overview are virtual (computed) fields, derived from the individual answers in each submission.

As a result of this design, the mock grading overviews rendered are completely independent of the mock submission (* the same mock submission is rendered for all three mock grading overviews). This makes it impossible for me to update the `takeEvery(actionTypes.SUBMIT_GRADING)` saga to effect a change to the mock grading overviews. Returning to the grading overviews page after submitting a grading triggers `handleFetchGradingOverviews`, which eventually invokes `mockFetchGradingOverview`, over-writing the value in the store with `mockGradingOverviews`.

I can verify that the `SUBMIT_GRADING` saga works as intended, by inspecting the value of the Redux store after submitting a grading. It would take a significant rewrite of the mocks to enable submitting mock grades to update the mock grading overviews (as with submitting mock answers to update the mock assessments) - which I propose should be left as an issue to be handled by a separate PR instead.

martin-henz:
> Hi @geshuming and @blackening, the button has been updated to load the current submission for grading in the current tab. CTRL+click opens the submission in another tab as expected.

Yes, that's good!

PR #677, opened 2019-07-07
binary trees needed for mission M6A

martin-henz:


martin-henz:
Sorry for slipping this in at the last minute. It's needed for Mission 6A

PR #693, opened 2019-07-09
Support password-protection and new practical type

geshuming:
We will have a new assessment type: practical. 

We will also support password protecting assessments, although it will be mainly practicals that support this.

New practical tab:
Capture

New password protected icon and tooltip:
Untitled

Password required on assessment fetch:
Capture2


martin-henz:
> We will have a new assessment type: practical.
> 
> We will also support password protecting assessments, although it will be mainly practicals that support this.
> 
> New practical tab:
> Capture
> 
> New password protected icon and tooltip:
> Untitled
> 
> Password required on assessment fetch:
> Capture2

Nice!

Yes, this should work for all kinds of assessments.

We shouldn't hard-wire any . behaviour for any kind of assessment. From the point of view of Cadet and Cadet Frontend, all assessments are the same. The students see the difference in the way they are set up by the teaching staff.

It should be super easy (eventually even without programming, maybe next year) to add and take away some kinds of assessment. 

This will make the Academy more flexible, more powerful, and at the same time simpler to develop, maintain and test.

jiachen247:
Closed with #878

Issue #689, opened 2019-07-08
Chat: Do some basic error catching for invalid tokens


geshuming:
![Capture](https://user-images.githubusercontent.com/43642522/60787230-ab6bb700-a18b-11e9-8837-2f0009fdd6ae.PNG)

martin-henz:
Can this crash the system? Where can invalid tokens come from?

geshuming:
Yes this crashes the frontend. Currently it happens because I did not supply a proper instance locator, but I think it can happen if a token is invalid, or if the chatkit server does not respond.

flxffy:
I have made a pull request #694 addressing this issue.

martin-henz:
I see. great!

martin-henz:
Addressed by https://github.com/source-academy/cadet-frontend/pull/694

PR #694, opened 2019-07-09
Chatkit TokenProvider try catch

flxffy:
Chatkit TokenProvider crashing developer view when token or url provided are invalid. Refer to issue #689


martin-henz:
Sorry cant review this. Have no clue.

src/components/chat/ChatApp.js
```diff
@@ -30,44 +31,52 @@ class ChatApp extends React.Component {
...
+            messageLimit: 100,
+            roomId: this.props.roomId
+          });
+        })
+        .then(currentRoom => {
+          this.setState({
+            connected: true,
+            currentRoom
+          });
+        });
```
geshuming: If I'm not wrong, you can actually use the [onRejected callback that the Promise provides](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch)

flxffy: I'll take a look later! Thanks

flxffy: I've added a `catch` which does nothing. I was wondering if there's any form of logging available in the frontend? Otherwise there's not much that can be done if chatkit refuses to connect

martin-henz: > I've added a `catch` which does nothing. I was wondering if there's any form of logging available in the frontend? Otherwise there's not much that can be done if chatkit refuses to connect

About logging: We have filed an Issue, but no action on this: https://github.com/source-academy/cadet-frontend/issues/169

geshuming: https://github.com/source-academy/cadet-frontend/blob/master/src/utils/notification.ts

 You can use these functions to show errors. The other way is to render an element showing an error

flxffy: Categorised the connection statuses - connecting, connected and failed to connect. Currently rendering a simple error message when connection fails.

src/components/chat/ChatApp.js
```diff
@@ -7,44 +7,43 @@ import MessageList from './MessageList.tsx';
...
-      instanceLocator: INSTANCE_LOCATOR,
-      tokenProvider: new TokenProvider({
-        headers: {
-          Authorization: `Bearer ${this.props.accessToken}`
-        },
-        url: `${BACKEND_URL}/v1/chat/token`
-      }),
-      userId: jwt_decode(this.props.accessToken).sub
-    });
+    var chatManager;
```
geshuming: Use `let` instead of `var`

flxffy: Fixed

flxffy:
Hi @geshuming, I believe you can squash and merge my PR, since there's only one fix. As for the error messages, I can't hide them since there are no options in the Chatkit documentation that allow it.

Issue #160, opened 2018-07-05
Don't use location as variable name


remo5000:
It clashes with window.location, which is accessible from anywhere as a global.

alcen:
Resolved!

PR #706, opened 2019-07-12
Refactor of code using variables named "location"

alcen:
This commit is meant as a patch for issue #160. `location` is already defined, used for `Window.location`, and should not be reused.

Variables that were named `location` were changed to either `wkspLocation` or `locat`. One function that used the variable `location` without defining it was refactored such that the correct reference was used.


alcen:
> Hi can we keep the name consistent everywhere, for example it has a different name here 

Ok, no problem. I have addressed this in the latest commit.
&nbsp;
> In my opinion the issue isn't really isn't an issue. We don't really use window.location

Well, I guess it makes the code more legible :/

martin-henz:
Yes, it's probably better like this. Nothing wrong with explicit names. The name "location" is indeed special in web programming with JavaScript: 
https://www.w3schools.com/jsref/obj_location.asp

So best avoid it.

src/components/academy/grading/GradingHistory.tsx
```diff
@@ -0,0 +1,81 @@
+import { Popover, PopoverInteractionKind, Position } from '@blueprintjs/core';
```
rrtheonlyone: Why is this file added back?

src/utils/history.ts
```diff
@@ -21,6 +21,6 @@ type lastLocation = string | null;
 
 export const history = createBrowserHistory();
 
-history.listen((location, action) => {
-  store.dispatch(updateHistoryHelpers(location.pathname));
+history.listen((locat, action) => {
```
rrtheonlyone: It is actually correct to use `location` here. Refer to https://github.com/ReactTraining/history#listening

src/components/academy/grading/GradingNavLink.tsx
```diff
@@ -0,0 +1,38 @@
+import { Classes, Icon } from '@blueprintjs/core';
```
rrtheonlyone: Same for this

alcen:
> I like this change! It indeed makes the naming more explicit and clear 👍
> Some comments:

Thanks for the review! Must have accidentally committed the removed files somewhere in the working tree, oops...

I have made the changes in the latest commit, do take a look!

PR #707, opened 2019-07-13
Integrate Sourcecast and Sourcereel components

jiayushe:
- Sourcecast workspace
image

- Sourcereel workspace
image

- Relevant to source-academy/cadet#418


ccyccyccy:
> This is a lot of code to look through. One thing I need to ask is do we really need to use redux? Is there really a need to persist the sourcecast state globally?

It's just good practice to keep data in the redux store isn't it? Maintaining a component level state is just going to confuse future maintainers with two different places for data storage, especially since we are going to need a lot of data from the redux store anyway.

martin-henz:
> > This is a lot of code to look through. One thing I need to ask is do we really need to use redux? Is there really a need to persist the sourcecast state globally?
> 
> It's just good practice to keep data in the redux store isn't it? Maintaining a component level state is just going to confuse future maintainers with two different places for data storage, especially since we are going to need a lot of data from the redux store anyway.

My hunch is that Chee Yuan is right. SourceReel and SourceCast are interacting with the playground in interesting ways. For example they activate the Run button. So I can imagine that it's a good idea to treat the state of SourceReel/Cast like any other globally available state and modify it explicitly.

geshuming:
> It's just good practice to keep data in the redux store isn't it? Maintaining a component level state is just going to confuse future maintainers with two different places for data storage, especially since we are going to need a lot of data from the redux store anyway.

Not everything should go on to the redux store. I actually think it is more confusing and harder to maintain because putting data in the redux store creates extra actions and reducers for each data, especially following our redux design pattern (actions, reducers, containers). 

Do consider the use cases of each data, and if it makes sense to persist them (for example, if someone is in the middle of recording a sourcecast, would that person really need to navigate to a different page and then navigate back to sourcecast, requiring that all that state is persisted?)

martin-henz:
> > It's just good practice to keep data in the redux store isn't it? Maintaining a component level state is just going to confuse future maintainers with two different places for data storage, especially since we are going to need a lot of data from the redux store anyway.
> 
> Not everything should go on to the redux store. I actually think it is more confusing and harder to maintain because putting data in the redux store creates extra actions and reducers for each data, especially following our redux design pattern (actions, reducers, containers).
> 
> Do consider the use cases of each data, and if it makes sense to persist them (for example, if someone is in the middle of recording a sourcecast, would that person really need to navigate to a different page and then navigate back to sourcecast, requiring that all that state is persisted?)

I see. I guess it's a trade-off then. I suppose it's not far-fetched that a user goes elsewhere during sourcecast viewing, and expects to resume the viewing where he left off.

jiayushe:
> > It's just good practice to keep data in the redux store isn't it? Maintaining a component level state is just going to confuse future maintainers with two different places for data storage, especially since we are going to need a lot of data from the redux store anyway.
> 
> Not everything should go on to the redux store. I actually think it is more confusing and harder to maintain because putting data in the redux store creates extra actions and reducers for each data, especially following our redux design pattern (actions, reducers, containers).
> 
> Do consider the use cases of each data, and if it makes sense to persist them (for example, if someone is in the middle of recording a sourcecast, would that person really need to navigate to a different page and then navigate back to sourcecast, requiring that all that state is persisted?)

Actually we are already keeping some data in local state. Those in redux are generally because they are needed to interact with other components. For example, the editor needs to know `recordingStatus`, `timeElapsedBeforePause` and `timeResumed` to record editor deltas and bind them with time when recording.

ccyccyccy:
> > It's just good practice to keep data in the redux store isn't it? Maintaining a component level state is just going to confuse future maintainers with two different places for data storage, especially since we are going to need a lot of data from the redux store anyway.
> 
> Not everything should go on to the redux store. I actually think it is more confusing and harder to maintain because putting data in the redux store creates extra actions and reducers for each data, especially following our redux design pattern (actions, reducers, containers).
> 
> Do consider the use cases of each data, and if it makes sense to persist them (for example, if someone is in the middle of recording a sourcecast, would that person really need to navigate to a different page and then navigate back to sourcecast, requiring that all that state is persisted?)

I don't think putting more action and reducers into the redux store would complicate things. Currently we are squeezing everything into the workspace reducer. We could totally create separate action and reducer files just for the sourcecast workspace which would then be imported from the main workspace reducer. It would be clear which actions or reducers are specific to sourcecast this way.

rrtheonlyone:
From the [Redux Documentation](https://redux.js.org/faq/organizing-state)

Some common rules of thumb for determining what kind of data should be put into Redux:

Do other parts of the application care about this data?
Do you need to be able to create further derived data based on this original data?
Is the same data being used to drive multiple components?
Is there value to you in being able to restore this state to a given point in time (ie, time travel debugging)?
Do you want to cache the data (ie, use what's in state if it's already there instead of re-requesting it)?
Do you want to keep this data consistent while hot-reloading UI components (which may lose their internal state when swapped)?

@jiayushe @ccyccyccy If any of the above applies to your use case, then perhaps a good idea to put it inside the `store`. If not, then it may be better to keep it within the local component state.

Also, its a good idea to rebase and clean up your commit history for large PRs like this. Its cleaner and much easier to review.

src/components/workspace/index.tsx
```diff
@@ -2,6 +2,9 @@ import Resizable, { ResizableProps, ResizeCallback } from 're-resizable';
 import * as React from 'react';
 import { Prompt } from 'react-router';
 
+import SourcecastControlbar, {
+  ISourcecastControlbarProps
+} from '../sourcecast/SourcecastControlbar';
```
geshuming: Same as above. Let's try to use composition.

src/containers/sourcecast/SourcecastContainer.ts
```diff
@@ -0,0 +1,117 @@
...
+      handleDebuggerResume: () => debuggerResume(location),
+      handleDebuggerReset: () => debuggerReset(location)
+    },
+    dispatch
+  );
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(Sourcecast);
```
geshuming: Is it possible to practice DRY here. We already have a connected `workspace`, and `Sourcecast` renders that `workspace`. Using composition, we should only add new mappings related to sourcecast here.

src/containers/sourcecast/SourcereelContainer.ts
```diff
@@ -0,0 +1,106 @@
...
+      handleDebuggerResume: () => debuggerResume(location),
+      handleDebuggerReset: () => debuggerReset(location)
+    },
+    dispatch
+  );
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(Sourcereel);
```
geshuming: Similarly

src/containers/sourcecast/SourcecastContainer.ts
ccyccyccy: > Is it possible to practice DRY here. We already have a connected `workspace`, and `Sourcecast` renders that `workspace`. Using composition, we should only add new mappings related to sourcecast here.

Not sure what you mean by we already have a connected `workspace`?
As far as I can tell `workspace` is just a component by itself and is not connected to the redux store.

src/components/workspace/index.tsx
ccyccyccy: Fixed by 9829b2e

src/components/workspace/Editor.tsx
```diff
@@ -8,51 +9,140 @@ import 'brace/mode/javascript';
...
+  KeyboardCommand
+} from '../sourcecast/sourcecastShape';
 import { checkSessionIdExists } from './collabEditing/helper';
 /**
  * @property editorValue - The string content of the react-ace editor
  * @property handleEditorChange  - A callback function
  *           for the react-ace editor's `onChange`
  * @property handleEvalEditor  - A callback function for evaluation
  *           of the editor's content, using `slang`
+ * @property editorReadonly - Used for sourcecast only
```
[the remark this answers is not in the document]
ccyccyccy: I'm not sure how to go about splitting this up as I am using a lot of APIs from `AceEditor`.
Should I wrap `Editor` with something like `SourcecastEditor` and reach into `Editor` with Refs?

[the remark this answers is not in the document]
geshuming: Let's duplicate `Editor` and apply your changes into `SourcecastEditor`. I think it would be easier to ensure that the current `Editor` works, than to add code to it. We will work out the refactoring later.

PR #713, opened 2019-07-14
Frontend: Refactor requests in backend saga

helloImHai:
Summary: 
Separate request into its own file and refactor it to use guard clauses.

Change log: 
- Separate requests into its own file.
- Changed imports of other files to reflect change in location of requests.
- Refactor requests to use guard clauses.


helloImHai:
Rebased

Issue #725, opened 2019-07-15
Strange label "skip story"


martin-henz:
<img width="806" alt="Screenshot 2019-07-15 at 9 03 59 PM" src="https://user-images.githubusercontent.com/30220768/61218438-9d4a0780-a744-11e9-83c0-194b55a4a936.png">

We should not say "skip story". Just remove that part and say: "Attempt"

alcen:
Handled by #727

PR #727, opened 2019-07-15
Change "Skip Story & Attempt" on buttons to just "Attempt"

alcen:
Simple change to remove "Skip Story" on the buttons displayed when trying to access an assessment for the first time. I took the opportunity to modify the icon used as well, from "step-forward" (similar to a 'skip' icon on videos) to "play".

only-attempt

This change is meant to resolve issue #725.


martin-henz:
I propose we merge this. @geshuming , what do you think?

There is no need to mention any "story" in the mission overview. We let students do the "obvious" thing and go straight to the mission.

The "stories" are what you see in the game. They should not appear elsewhere, in my view.

PR #740, opened 2019-07-18
Overhaul of SideContent component

alcen:
Currently, the Workspace SideContent component uses self-implemented tabs. In this commit, these are replaced with the BlueprintJS Tabs component to afford new functionality, and to allow for continued updates from their development team.

Key changes:
* Tabs and icons are now larger, with an animated underline that moves across tabs when the tabs are switched:
  tabs
* Tab panels are now all loaded into the DOM when the page is loaded; previously the tab panels will only be rendered by React when selected (this can be disabled if it is causing performance issues)
* Documentation for new properties of SideContent and SideContentTabs
* Ability to set default tab to be shown when the page is loaded
* Ability for tabs to be disabled:
  disabled
* Added loading screens for Data Visualiser, Inspector and Env Visualiser components as their React components are now being loaded faster than the corresponding libraries are
  loading
* Nested SideContentTabs components on Mission Control feature were removed or replaced with dropdown menus to clean up the interface

Please provide feedback about the look and feel of the new tabs. Do take a look and suggest any changes!


alcen:
We can, but the larger icons are easier to click on and work better with the animated line

martin-henz:
> We can, but the larger icons are easier to click on and work better with the animated line

I agree that the icons are on the big side. A bit smaller would be better. The line might not be necessary as long as you have the color highlighting.

alcen:
Thanks for the feedback, what about this? The tab icons are still slightly bigger than the others, but I think it's ok because the icons do not some with text like the other buttons on the UI.

tabs2

> The line might not be necessary as long as you have the color highlighting.

The line can be removed easily as well, it is entirely up to preference. The blue styles can be hidden if it stands out too much from the interface too

alcen:
The large buttons are far easier to use on touch devices, admittedly. Here is a screenshot of how the large buttons look on iOS Safari:

D47C6AA5-58FE-448D-B07D-039645543952

geshuming:
It will definitely be nicer for touch devices, but we don't currently plan to support that.

martin-henz:
> It will definitely be nicer for touch devices, but we don't currently plan to support that.

Actually, I think it's important to support touch devices. The Academy is almost usable on mobile devices. With a bit of effort, we can get there. 

I think this is important for our outreach. Sound processing is fun with phones. Also, next year we will have many more libraries that use sensors such as accelerometer and camera.

In the light of that, I support the large icons.

martin-henz:
> In my opinion, if we really want to design for mobile users, we should separate the style choices, similar to how this very github website has a Desktop view and a Mobile view.

I agree. Let's look into this in Sem 2 CP3108.

alcen:
> Actually, I think it's important to support touch devices. The Academy is almost usable on mobile devices. With a bit of effort, we can get there.
> 
> I think this is important for our outreach. Sound processing is fun with phones. Also, next year we will have many more libraries that use sensors such as accelerometer and camera.
> 
> In the light of that, I support the large icons.

Some context: Initially the icons were made to be as similar to the original as possible. I was unsure of larger icons at first too, but I decided to implement them after much comparison and toying around on various interfaces

Not too sure about the design decision to only include icons in the tab buttons though (and the titles as tooltips instead of being in-line with the icons), but I decided to go along with it for now

&nbsp;

> We can add left and right padding, the button does not have to be tall

The icons will get squished though, isn’t it?

&nbsp;

> In my opinion, if we really want to design for mobile users, we should separate the style choices, similar to how this very github website has a Desktop view and a Mobile view.

True, but many web apps also integrate their desktop and mobile designs together. I feel that such integration is a better solution than two separate websites with different interfaces

&nbsp;

> I agree. Let's look into this in Sem 2 CP3108.

Ok, so I guess the consensus is the smaller icons. Would the icons in the second comment I made be satisfactory, if not what changes could be made?

martin-henz:
> Thanks for the feedback, what about this? The tab icons are still slightly bigger than the others, but I think it's ok because the icons do not come with text like the other buttons on the UI.
> 
> tabs2
> 
> > The line might not be necessary as long as you have the color highlighting.
> 
> The line can be removed easily as well, it is entirely up to preference. The blue styles can be hidden if it stands out too much from the interface too

Yes, this is good. Let's go for it.

martin-henz:
Regarding mobile/touch: I've played around a bit, to check out sound recording. It's almost usable. A couple of issues:

* The re-sizing is very difficult, because you need to hit the line with your finger.
* the pop-up keyboard somehow has a large wasted white space on my Android phone.

alcen:
> Yes, seems good. Let's see if we get more feedback from the team.

Wow, thanks for the fast review. I was just about to comment that I had made the changes!

Anyway, for future reviewers: please refer to https://github.com/source-academy/cadet-frontend/pull/740#issuecomment-512699885 for the screenshot of the tabs in the current commit

src/components/workspace/side-content/ListVisualizer.tsx
```diff
@@ -1,24 +1,46 @@
...
+    if (this.$parent && element) {
+      // List Visualizer has been loaded into the DOM
+      element.init(this.$parent);
+      this.setState((state, props) => {
+        return { loading: false };
+      });
+    } else {
+      // Try again in 1 second
+      window.setTimeout(this.tryToLoad, 1000);
+    }
```
geshuming: Let's not implement `tryToLoad` or `NonIdealState` into these components. Unless we know that the init stage is failable, which I doubt so, we don't need to create "loading screens"

src/components/workspace/side-content/EnvVisualizer.tsx
```diff
@@ -1,24 +1,46 @@
...
+      // Env Visualizer has been loaded into the DOM
+      element.init(this.$parent);
+      this.setState((state, props) => {
+        return { loading: false };
+      });
+    } else {
+      // Try again in 1 second
+      window.setTimeout(this.tryToLoad, 1000);
+    }
+  };
```
geshuming: same here

alcen:
@geshuming The loading screen is meant as a fix for this error:

failable

Another alternative is to set the Playground side content not to load in all the tabs at once, but then this error would occur if you set the default loaded tab to the List Visualiser/Inspector/Env Visualiser

Not sure if there are plans to use this default tab feature, but I feel like there could be cases where assignments would want to show the List Visualiser/Inspector/Env Visualiser features outside of the playground and load them in by default

What do you think?

geshuming:
@alcen I see, so because the visualizer gets rendered first before the parent, it leads to the error. Is it possible to explore ways to delay the render of the visualizer?

martin-henz:
Yes, I feel that the missions are currently deprived of some of our features, such as the list visualiser. In practice, this means that students will leave the mission and go to the playground to have a better experience. Quite unfortunate!

Screenshot 2019-07-29 at 3 38 31 PM

The debugger tabs automatically appear when the program runs into a breakpoint. Can we do the same for the data visualiser? Currently, it remains hidden, even when you call draw_data().

alcen:
> @alcen I see, so because the visualizer gets rendered first before the parent, it leads to the error. Is it possible to explore ways to delay the render of the visualizer?

Possibly React state could be used to determine when exactly to initialise the List Visualiser/Inspector/Env Visualiser scripts, but the component would still display the loading screen while the scripts are not loaded yet

Other than that the easiest way to delay them would be to set the Blueprint tabs to only render React components when the tabs are switched (and hide those tabs from view by default), so there would be an artificial delay before someone gains access to those components. Before, this was how the tabs worked so such an error would have been rare, and probably there were less scripts to load (only List Visualiser) so they would have finished loading before the corresponding React component was mounted

Though, there could also be other methods of delaying React components from rendering, not too sure there

> Yes, I feel that the missions are currently deprived of some of our features, such as the list visualiser. In practice, this means that students will leave the mission and go to the playground to have a better experience. Quite unfortunate!
> 
> The debugger tabs automatically appear when the program runs into a breakpoint. Can we do the same for the data visualiser? Currently, it remains hidden, even when you call draw_data().

It will probably be next in the pipeline once the new tabs are implemented!

geshuming:
@alcen can you try forward ref to resolve the typeError? https://reactjs.org/docs/forwarding-refs.html

alcen:
> @alcen can you try forward ref to resolve the typeError? https://reactjs.org/docs/forwarding-refs.html
> nevermind, was just throwing ideas. I still think there should be a way to resolve this without try-catch

No problem. I feel like there should be a way as well

Perhaps the library loading can be done by React instead, then we can change the component's state when it is done causing a re-render?

PR #743, opened 2019-07-18
Patch issue with Grading in production

rrtheonlyone:
There was a regression bug in production which caused grading to fail. This was due to a mismatch of the fields passed into the backend API.

This bug has been patched. This PR updates the current master to be in sync with the deployed version.

#### Changes
- Removed comment from `SUBMIT_GRADING` action
- Send empty string as comment to backend

### DIFF
--- a/src/actions/__tests__/session.ts
+++ b/src/actions/__tests__/session.ts
@@ -139,15 +139,13 @@ test('submitAssessment generates correct action object', () => {
 test('submitGrading generates correct action object with default values', () => {
   const submissionId = 8;
   const questionId = 2;
-  const comment = 'test comment here';
 
-  const action = submitGrading(submissionId, questionId, comment);
+  const action = submitGrading(submissionId, questionId);
   expect(action).toEqual({
     type: actionTypes.SUBMIT_GRADING,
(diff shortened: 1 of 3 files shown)


rrtheonlyone:
This is what the action looked like:
```
export const submitGrading: ActionCreator<actionTypes.IAction> = (
  submissionId: number,
  questionId: number,
  comment: string,
  gradeAdjustment: number = 0,
  xpAdjustment: number = 0
) => ({
  type: actionTypes.SUBMIT_GRADING,
  payload: {
    submissionId,
    questionId,
    comment,
    gradeAdjustment,
    xpAdjustment
  }
});
```


However, the grading component passed this in:
```
this.props.handleGradingSave(
        this.props.submissionId,
        this.props.questionId,
        gradeAdjustmentInput,
        xpAdjustmentInput
      );
```

The `mapDispatchToProps` will not be able to map the parameter fields in as well based on the name. So instead, it just passes the first four into the first function. So comment takes the `gradeAdjustment` and xpAdjustment is always 0.

You can test this out by simply inspecting the HTTP request/response packets when making the POST request.

PR #767, opened 2019-07-28
Use a custom theme for editor

### LINKED ISSUE #761: UI: Improving matching parentheses
The highlighting of matching parentheses is *very* subtle: 
bild
It needs a better highlight, like possibly changing the background colour of the character. 

Also -- I suggest adding something like the Emacs rainbow-delimiters mode which colours the matching parentheses in the same colour for improved matching:
bild
Image above also shows the clearer highlighting of the outermost parentheses.

jiayushe:
- Use background highlighting for bracket matching
image

- Fix #761

### DIFF
--- a/src/components/workspace/source.js
+++ b/src/components/workspace/source.js
@@ -1,0 +1,119 @@
+ace.define("ace/theme/source",["require","exports","module","ace/lib/dom"], function(acequire, exports, module) {
+
+exports.isDark = true;
+exports.cssClass = "ace-source";
+exports.cssText = ".ace-source .ace_gutter {\
+background: #011e3a;\
+color: rgb(128,145,160)\
+}\
+.ace-source .ace_print-margin {\
+width: 1px;\
+background: #555555\
+}\
+.ace-source {\
+background-color: #002240;\
+color: #FFFFFF\
+}\
+.ace-source .ace_cursor {\
+color: #FFFFFF\
+}\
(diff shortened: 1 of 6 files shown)


src/components/workspace/source.js
```diff
@@ -0,0 +1,119 @@
+ace.define("ace/theme/source",["require","exports","module","ace/lib/dom"], function(acequire, exports, module) {
```
geshuming: Can we add a comment indicating that this theme originated from the cobalt theme that we have been using. As well as comment on what changes we made?

jiayushe: Okay done.

TobiasWrigstad:
I turned out I entered a dependency hell when upgrading node packages to be able to build cadet-frontend again. Will let you know ASAP.

TobiasWrigstad:
I would like to suggest that -- since we are making changes to the theme -- we change the font of the code from Ubuntu Mono to Inconsolata, to nicely unify with the SICP look.

jiayushe:
> I would like to suggest that -- since we are making changes to the theme -- we change the font of the code from Ubuntu Mono to Inconsolata, to nicely unify with the SICP look. (I mean to piggy back this change on this PR since we are theming.)

Hmm I checked the SICP website, the font that is being used seems to be `Droid Sans Mono`?

image

PR #773, opened 2019-07-29
SWS: Adding new string functions to runes library in Source Academy.

sandydays:
Temporary solution to allow students to get a string converted to a list of characters.


martin-henz:
Note that this PR is not meant to be merged into Master.

PR #778, opened 2019-07-30
Fix breaking changes due to #713

jiayushe:


jiayushe:
> OK, i assume this is for the sourcecast API calls?

Yea, it's needed to send binary data throught XHR.

PR #781, opened 2019-07-30
clean up libraries: removing streams and adding comment for lists sta…

martin-henz:
Adding comment for lists, stating the purpose to remain here.


public/externalLibs/streams/stream.js
```diff
@@ -1,289 +0,0 @@
-// stream.js: Supporting streams in the Scheme style, following
```
jiayushe: In \public\externalLibs\index.js line 39, the import statement for stream library should be removed.

martin-henz: Good catch. Addressed in https://github.com/source-academy/cadet-frontend/pull/781/commits/678bf320e77961e4604de9abde7aabfb5871e31d

martin-henz:
Checked that streams still work

PR #787, opened 2019-07-31
Add feature to allow manipulation of max execution time in frontend

### LINKED ISSUE #786: Max Execution Time causing issues when programming games
Max Execution Time is set to 1s by default. To increase it, the same code needs to be sent to the js interpreter twice. This caused issues when running programmes that emphasizes interactivity, e.g. games.

jiayushe:
- Bump js-slang to 0.2.5
- Add feature to allow manipulation of max execution time in frontend
- Fix #786


martin-henz:
How does this work?

Issue #739, opened 2019-07-18
Grading: Pressing "Save" should go to next question


martin-henz:
This is going to support the most common use of "Save": The grader is done with grading the question and goes to the next question.

geshuming:
We could look at having 2 buttons: `Save` and `Save and Continue`

martin-henz:
I think the overwhelmingly most common use case will be to "Save and Continue" so having a "Save" button would just clutter the interface. In the rare case where a grader wants to save and still look at the grades some more, he can press the back button.

I think the addition of the Chat has changed this use case. The chat is independent of the grade and does not require "save".

blackening:
There's a use case for multiple saves. In particular, when grading long questions like the practical exam, what i fill up in the comments box can be a lengthy essay. (involving 9 x 1 mark deductions) 

In such a situation, i save pretty often.

Save and move to next question is used more often in short questions.

martin-henz:
> There's a use case for multiple saves. In particular, when grading long questions like the practical exam, what i fill up in the comments box can be a lengthy essay. (involving 9 x 1 mark deductions)
> 
> In such a situation, i save pretty often.
> 
> Save and move to next question is used more often in short questions.

Our "comment box" is now the Chatkit chat, which is saved as soon as you hit return. So I suggest you just spit your comment into multiple lines in the chat. Note that the "save" button now just saves two numbers: the grade and the XP.

blackening:
This is very painful. Actually nearly unusable. 

The first time i put the comments in, it's a single pass full of things the students shouldn't see.

For example, a list of probable errors to check for, a list of known errors. No deductions are done during that pass. After that, the deductions are done in a batch after i determine the grading scheme. 

Sometimes, i grade first, then determine that a certain mistake is common enough and go back and comment on every submission that makes those mistakes.

martin-henz:
We decided that we bring back the primary comment box, and @alcen is looking into it.

Regarding the buttons, how about:
* Cancel: disregards edits in grades and comment box since the last save
* Save: saves current grades and comment and stays on the same question.
* Next (can also be called "Save & Next"): saves current grades and comment and moves to next question.

In the rare case where the staff wants to cancel recent edits and advance to the next question, two clicks are needed: "Cancel" followed by "Next". I think this is better than popping up a box in the most common "Next" use case: the staff wants to save and then go to the next question.

I guess a small detail, but since we are discussing it...

alcen:
[...]

PR #789, opened 2019-07-31
Comments feature for grading

alcen:
After some recent discussions, @martin-henz has decided that the comments feature should be re-implemented. This will ensure that grading is not too reliant on the Chatkit API.

Hence this commit will be focused on adding comments and comment editing to the grading system.

<br>

Key changes:

* The 'react-mde' Markdown editor (documentation [here](https://github.com/andrerpena/react-mde)) has been put back in the grading editor, and updated to the latest version v7.6.2 (was previously v5.6.0)
1editor

* The grading report shown to the student will now display comments:
2 acomments

* Editor preview uses the same component as the comments displayed to students - compare the screenshot below to the one above
2 z 5preview

* The 'Save' button has been replaced with 'Submit and Continue' - see issue #739

* New: a draft feature where graders can save drafts of their grade adjustments/XP adjustments/comments to their browser storage (without making the grading available for view for the student):
3drafts

<br>

For clarity, this is how the grading editor will work in this version:

1. Enter the grading overview page and click 'Edit' to enter the grading workspace

2. The default text loaded into the editor will be the current comment for the student's answer, if any

3. The website will look for a grading draft corresponding to that specific student's answer that is stored on the local system. If such a draft exists, a notification pops up and 'Load Draft' is highlighted until it is clicked:
4loaddraft

4. Proceed with grading. Once any of the values change, 'Save Draft' and 'Submit and Continue' will light up.

5. If the grade or XP are out of bounds, a notification will appear and you will be unable to save or submit:
5outofbounds

6. Select 'Save Draft' and the draft will be saved into the local browser storage. It can be loaded again with 'Load Draft'

7. If 'Load Draft' is clicked, and the saved values are different from the editor values, there will be a dialog box asking for confirmation:
6loadingerror

8. Select 'Submit and Continue' when done with the grading. The website will attempt to send the grading results to the backend, and when this is successful it will automatically navigate to the next page:
7continue

To accommodate these changes, the backend had to be modified to store comments in the database. Refer to this Pull Request for backend commits: source-academy/cadet#461

Please leave comments on either Pull Request to suggest changes!


alcen:
> I don't see the benefit of using localStorage to save a draft.

The use of local storage is meant to be temporary, allowing graders to pause their grading and resume where they left off later, without exposing the unfinished grades to students. 

On issue #739, @blackening mentions that graders may input temporary marking schemes before they actually begin to grade. This draft saving feature would help with that, serving as a convenient way to allow individual graders to store data that will only be available to themselves

It should function similarly to copying the grade out to a local text file and saving it, then copying it back into the editor once a grader is ready to finish the grading. The local storage saving makes it much easier as it tracks which grade belongs to which student and which question automatically, and loading fills in all the fields (grade, XP, comments) with a single click

<br>

> We should use the database as the source of truth. Using localStorage makes it difficult to discern whether the data in the database or localStorage should be used.
> 
> 
> Other graders might have written their own comments. This again creates an issue with the source of truth.

It should not cause an issue with the source of truth, as the current comment that the student can see is retrieved from the backend and always loaded by default. Only if there is a draft in the local storage will the 'Load Draft' option appear, prompting the grader to manually select if personal saved data is to be used

Possibly, one change to be made could be an option to append the local draft to the bottom of the document, to allow for ease of use when graders want to add on to something that is already graded

<br>

> I propose we remove the use of localStorage. Save should just save to the database.

Currently, the backend does not discern the difference between unfinished comments and comments that are ready for view, meaning that once something like a marking scheme is "saved" to the database students will be able to access it, which may confuse them or mislead them with the incorrect information

Having said that, a 'drafts' column could be added to the database, but there is no need to unnecessarily complicate the system or crowd the database with temporary data that only need to be accessed by a single user. Use of local storage resolves both of these issues, making sure that the backend system remains simple to understand for developers, and also that the drafts are only accessible by the creators.

One change could be to make it very clear that the draft is saved locally; so graders using publicly accessible computers will understand that if they do not submit the grading, there is no way to persist draft data to another grading session later on (on another computer)

blackening:
> > I don't see the benefit of using localStorage to save a draft.
> 
> The use of local storage is meant to be temporary, allowing graders to pause their grading and resume where they left off later, without exposing the unfinished grades to students.
> 
> On issue #739, @blackening mentions that graders may input temporary marking schemes before they actually begin to grade. This draft saving feature would help with that, serving as a convenient way to allow individual graders to store data that will only be available to themselves
> 
> It should function similarly to copying the grade out to a local text file and saving it, then copying it back into the editor once a grader is ready to finish the grading. The local storage saving makes it much easier as it tracks which grade belongs to which student and which question automatically, and loading fills in all the fields (grade, XP, comments) with a single click
> 

Ok, a little comment here. For the practical exam, i'm not sure how we will grade it.

On the other hand, what i'm sure is that it's very unlikely to be a single grader. 

Usually what happens in distributed grading environments is
- the grader will grade everything he's sure about.
- the grader will mark everything he isn't sure about.
- Someone will come by and figure out those which needs clarification/moderation.

In this case, using the local storage may be detrimental to the grading experience.
> 
> > We should use the database as the source of truth. Using localStorage makes it difficult to discern whether the data in the database or localStorage should be used.
> > Other graders might have written their own comments. This again creates an issue with the source of truth.
> 
> It should not cause an issue with the source of truth, as the current comment that the student can see is retrieved from the backend and always loaded by default. Only if there is a draft in the local storage will the 'Load Draft' option appear, prompting the grader to manually select if personal saved data is to be used
> 
> Possibly, one change to be made could be an option to append the local draft to the bottom of the document, to allow for ease of use when graders want to add on to something that is already graded
> 
> 
> > I propose we remove the use of localStorage. Save should just save to the database.
> 
> Currently, the backend does not discern the difference between unfinished comments and comments that are ready for view, meaning that once something like a marking scheme is "saved" to the database students will be able to access it, which may confuse them or mislead them with the incorrect information

Yes, this is a problem, i consider this as a future WIP.

> 
> Having said that, a 'drafts' column could be added to the database, but there is no need to unnecessarily complicate the system or crowd the database with temporary data that only need to be accessed by a single user. Use of local storage resolves both of these issues, making sure that the backend system remains simple to understand for developers, and also that the drafts are only accessible by the creators.
> 
> One change could be to make it very clear that the draft is saved locally; so graders using publicly accessible computers will understand that if they do not submit the grading, there is no way to persist draft data to another grading session later on (on another computer)

The main issue is that there's no finalize grading. It's an implicit process. WIP grading leaks out during the grading process.

For now, this is acceptable. So long as they don't get spammed with notifications when grading its fine.

alcen:
> Ok, a little comment here. For the practical exam, i'm not sure how we will grade it.
> On the other hand, what i'm sure is that it's very unlikely to be a single grader.
> Usually what happens in distributed grading environments is
> 
> the grader will grade everything he's sure about.
> the grader will highlight everything he isn't sure about*.
> Someone will come by and figure out those which needs clarification/moderation.
> 
> 
> Usually this is something which doesn't have a grading scheme yet. This will require the graders to reach a consensus on how the grade is to be assigned for a certain class of mistakes. E.g. how much do we deduct for infinite loop, whether something can ECF or not, etc.
> 
> In this case, using the local storage may be detrimental to the grading experience.

Okay, I see it may not be used much for grading the practical

For Missions and Quests however, how useful would this feature be? I don't see the harm of having this feature on the side, it should be useful for those assignments that can be graded by a single grader

blackening:
> > Ok, a little comment here. For the practical exam, i'm not sure how we will grade it.
> > On the other hand, what i'm sure is that it's very unlikely to be a single grader.
> > Usually what happens in distributed grading environments is
> > the grader will grade everything he's sure about.
> > the grader will highlight everything he isn't sure about*.
> > Someone will come by and figure out those which needs clarification/moderation.
> > Usually this is something which doesn't have a grading scheme yet. This will require the graders to reach a consensus on how the grade is to be assigned for a certain class of mistakes. E.g. how much do we deduct for infinite loop, whether something can ECF or not, etc.
> > In this case, using the local storage may be detrimental to the grading experience.
> 
> Okay, I see it may not be used much for grading the practical
> 
> For Missions and Quests however, how useful would this feature be? I don't see the harm of having this feature on the side, it should be useful for those assignments that can be graded by a single grader

This is fine. The majority use case is mostly single user. 

The only question is how much complexity is introduced vs how much value the system adds.

geshuming:
> The only question is how much complexity is introduced vs how much value the system adds.

I don't think there's much value. It's not a form that takes a long time to complete, nor is the data so big that it justifies saving it locally over saving it on the database.

martin-henz:
> > The only question is how much complexity is introduced vs how much value the system adds.
> 
> I don't think there's much value. It's not a form that takes a long time to complete, nor is the data so big that it justifies saving it locally over saving it on the database.

I agree. Let's take out local storage of drafts. Just save is fine. Once saved, the grader can go back, edit and save again, right?

alcen:
> > I don't think there's much value. It's not a form that takes a long time to complete, nor is the data so big that it justifies saving it locally over saving it on the database.
> 
> I agree. Let's take out local storage of drafts. Just save is fine. Once saved, the grader can go back, edit and save again, right?

The issue I have with just "Save" is that it is not just storing the grade in the database, but also making them available for students to view. For answers that are more difficult to grade, and that graders would like to reference other materials (like other students' answers) the "Save Draft" would be more useful, allowing the grader to type a comment halfway and then return to it

Ideally, the backend should be set up to receive the drafts, but this will make it much more complicated as there will have to be logic to clear the drafts from the system once the grading is finalised and available for view. Also, the database may be storing a lot of unnecessary data if the drafts were sent to it, as there could possibly be one draft per question, per student, per grader, leading to many permutations of stored unique drafts

Collaboration on the same draft could also be done, but if this is required the grading should probably be discussed in detail by all graders at a meeting, before the grading process begins

However, in any case making grades available for students to view should be more permanent, especially with the notification system. It feels really bad as a student to look at your grade one day and then realise it has completely changed on the next, and there is no indication of when you can be certain that the grade is final. Once the grades are made available for view by the students, changes should only be made on a case-by-case basis

<br>

Overall, I think it comes down to the style of grading for individual graders. For weaker students, graders may prefer lengthy, descriptive comments which may be more useful. These comments would take a longer time to create and so the "Save Draft" may come in handy. But for stronger students, they might many short comments like "Good job", which may even be handled in the chat instead of using the grading comments. Providing more options instead of less would make for a better experience in my opinion

Perhaps we should set up the backend to handle drafts, but this will require more development time. 
For now, I think the local storage is a passable solution.

blackening:
> Ideally, the backend should be set up to receive the drafts, but this will make it much more complicated as there will have to be logic to clear the drafts from the system once the grading is finalised and available for view. Also, the database may be storing a lot of unnecessary data if the drafts were sent to it, as there could possibly be one draft per question, per student, per grader, leading to many permutations of stored unique drafts

Just put a isGraded flag. It's not hard to implement. 

Writing test cases will take most of the time. 

Do not need to optimize unless you can show that it needs optimization. In this case it's 1 text field per question which is acceptable.

martin-henz:
Thanks, guys, for reasoning this out so carefully.

I understand that it would be nice to have a "save draft" feature. However, it would make the system more complicated and harder to maintain, and is a source of error. For now, I'm arguing for a simple and robust implementation. I'd rather go to the Avengers and tell them: (1) Be careful with Save: Students immediately see the comments.
(2) Don't use Save to save drafts of your comments.
(3) To grade all your 8 students, open them in a different tab, look at their solutions and compare. When you are happy with the grading scheme, enter your grades and comments, then save, then go to the next question.

blackening:
> (1) Be careful with Save: Students immediately see the comments.
> (2) Don't use Save to save drafts of your comments.

We can just not notify students until we are done grading. Unless students check immediately, which is unlikely (no email notifications), this is fine for the moment.

alcen:
> Just put a isGraded flag. It's not hard to implement.
> Writing test cases will take most of the time.
> Do not need to optimize unless you can show that it needs optimization. In this case it's 1 text field per question which is acceptable.

This sounds like a good suggestion! One question is, what should be shown in the grading editor when you want to edit grading that has already been submitted. At the top, instead of "Currently Grading Student_X", maybe it should show "Editing Grade of Student_X" instead?

<br>

> Thanks, guys, for reasoning this out so carefully.
> I understand that it would be nice to have a "save draft" feature. However, it would make the system more complicated and harder to maintain, and is a source of error. For now, I'm arguing for a simple and robust implementation. I'd rather go to the Avengers and tell them: (1) Be careful with Save: Students immediately see the comments.
> (2) Don't use Save to save drafts of your comments.
> (3) To grade all your 8 students, open them in a different tab, look at their solutions and compare. When you are happy with the grading scheme, enter your grades and comments, then save, then go to the next question.

Ok, I understand this point of view. I will remove the drafts feature for now.

Maybe we should log this as an issue, and then we can come back to it later on. What do you think?

<br>

> (1) Be careful with Save: Students immediately see the comments.
> (2) Don't use Save to save drafts of your comments.
> 
> We can just not notify students until we are done grading. Unless students check immediately, which is unlikely (no email notifications), this is fine for the moment.

Actually as a student I did check the Source Academy quite regularly. Some weeks there are a lot of assignments available and I would log in and check for updates often, including my grades

martin-henz:
> Maybe we should log this as an issue, and then we can come back to it later on. What do you think?

Yes, definitely log as an issue. We shall then label it as _postponed and re-visit after the semester, according to:
https://github.com/source-academy/cadet-frontend/wiki#policy-on-issues-and-prs

alcen:
Okay, I have pushed a new commit! Now instead of saving drafts, we have "Save Changes" and "Discard Changes"

newdraft

Of course, the "Save" now refers to actually submitting the grading entry into the database for students to see

martin-henz:
Looks great!

geshuming:
What does the discard change function do? Does it delete all unsaved changes, reverting to the previous saved data (from the database) or does it set it blank?

alcen:
> What does the discard change function do? Does it delete all unsaved changes, reverting to the previous saved data (from the database) or does it set it blank?

Yes, it reverts the state to the previous saved data from the database. It never sets it to blank unless the grading did not originally exist on the database

alcen:
There is a dialog box set up before the discard actually takes place, so graders should not accidentally throw away their work

PR #798, opened 2019-08-01
Version cadet

martin-henz:
fixes wrong version in js console startup message


src/utils/constants.ts
```diff
@@ -3,7 +3,7 @@ import * as dotenv from 'dotenv';
 dotenv.config();
 
 export const LUMINUS_CLIENT_ID = process.env.REACT_APP_LUMINUS_CLIENT_ID;
-export const VERSION = process.env.REACT_APP_VERSION;
+export const SOURCE_ACADEMY_VERSION = 'Cadet';
```
geshuming: Why are we replacing this with a constant? The log will read "Source Academy Cadet", which sounds weird.

martin-henz: "Cadet" is the name of the version. We could also write "Source Academy 3". 

The "Contributors" page currently says:

This page includes all developers who contributed to the Source Academy Cadet (2018), ...

Issue #607, opened 2019-06-20
Consolidate buttons as a prop to be passed down from workspace


geshuming:
[Right now, the control bar props is getting crowded with button-related props.](https://github.com/source-academy/cadet-frontend/blob/master/src/components/workspace/ControlBar.tsx#L30-L70)

Suggestion: Pass down buttons as a single prop. The workspace component will determine what type of buttons are required depending on the location (assessment, grading, playground), and pass down the required buttons.

For example, an assessment workspace will pass down `Run`, `Save`, `Reset` buttons, and playground workspace only pass down the `Run` button, and all the other buttons.

PR #801, opened 2019-08-01
Mission Autograder UI improvements, tests, run-all functionality

Aulud:
## Mission Autograder UI improvements, tests, run-all functionality
Summary: Improve the UI, UX and reliability of the Mission Autograder in assessment and grading environments.

### Changelog
- Enforce minimum width for Mission Autograder tab to ensure sufficient display area
- Reduce vertical footprint of Testcase and Autograder cards in the Mission Autograder
- Enable clicking on Testcase cards to run the testcase itself
- Add headers to Testcases and Autograder Results when a non-zero number of testcases or autograder results are rendered
- **EXPERIMENTAL**: Enable Mission Autograder to render `<canvas>` elements (containing WebGL graphics e.g. runes, curves) for testcases
   - Caveat: testcases for runes should be written as `show(<function call to compose rune>);` and those for curves should be written as `display_curve(<function call to compose curve>);`
- Pipe interpreter errors that arise during testcase execution through to 'Actual result' cell
- Remove redundant code from `EVAL_TESTCASE_SUCCESS` reducer since `consoleLogs` are not rendered in the 'Actual result' cell
- Enable 'Run' button in assessments (all users)/submissions (graders) to run all available testcases when the Mission Autograder is the active side-content tab
    - Caveat: the testcases will only run if the evaluation of the editor program succeeds without errors - this allows users to still interrupt execution of the editor program, whilst preventing common errors (e.g. syntax errors) from being replicated across all the testcases
- Add new `RESET_TESTCASE` action to clear testcase data (`results` and `errors`) prior to execution
- Reinstate `sideContentActiveTab` in IWorkspaceState, `UPDATE_ACTIVE_TAB` action, reducer and `handleActiveTabChange` handler

### Bugfixes
- Fix incorrect type of `result` in `ITestcase` interface
- Fix re-executing a testcase not re-applying the CSS fade animation
- Fix bug where the result 0 (number) renders `No Answer` in the 'Actual result' cell

### Mocks and Tests
- Add sample testcases to mock submission
- Add snapshot and DOM rendering tests for Mission Autograder component and sub-components
- Add tests for executing run-all testcase funtionality with the Autograder

#### Before (testcases section)
testcases-old

#### Before (autograder results section)
autograder-old

#### After
testcases-autograder-new

#### Feature demonstration (GIF)
mission-autograder-demo

Last updated 2 Aug 2019, 01:30AM


alcen:
I like the interface changes! One question though, why is the side content tabs handled by Redux again? Is there no way to keep it within React for the "Run All Tests" functionality

geshuming:
There are some ways but i suggested reverting to redux for now as it is slightly easier to handle considering how most of the logic is in redux anyway

alcen:
> There are some ways but i suggested reverting to redux for now as it is slightly easier to handle considering how most of the logic is in redux anyway

Ah, I see, but I feel that it isn't really necessary to do so just to change the "Run" action when the Autograder tab is loaded. What if I want to run it without the tests? There is not much indication that "Run" does different things when I am on different tabs

I suggest we handle this completely in React? We could pass down a function to the `SideContentTabs` to set the state of `GradingWorkspace` when the new tab is `SideContentType.autograder`, which would change the Run button. Or have another button for "Run All Tests"?

What do you think?

martin-henz:
> Or have another button for "Run All Tests"?

That sounds good.

Is the alternative to use the old "Run" button in order to run all tests? That wouldn't seem quite right.

blackening:
> Is the alternative to use the old "Run" button in order to run all tests? That wouldn't seem quite right.

It's contextual. Only does so when you have the tests panel open, otherwise runs normally. 

The default tab is the question description, which doesn't run tests.

This behavior should be rather intuitive to learn.

Aulud:
> > There are some ways but i suggested reverting to redux for now as it is slightly easier to handle considering how most of the logic is in redux anyway
> 
> Ah, I see, but I feel that it isn't really necessary to do so just to change the "Run" action when the Autograder tab is loaded. What if I want to run it without the tests? There is not much indication that "Run" does different things when I am on different tabs
> 
> I suggest we handle this completely in React? We could pass down a function to the `SideContentTabs` to set the state of `GradingWorkspace` when the new tab is `SideContentType.autograder`, which would change the Run button. Or have another button for "Run All Tests"?
> 
> What do you think?

The reason why I did not implement it as a standalone 'Run All Tests' button was to avoid the scenario where the editor's program itself contained errors. Attempting to run all testcases then would then replicate the same error across every testcase (dumping them to the REPL as multiple entries). The alternative here would be to wait for the result of each testcase before proceeding to the next, terminating execution of any subsequent testcases if one fails (a la Kattis style).

Regarding the run-all functionality, there were two ways I envisioned this was possible:

(1) Store the state of the side content tab in the relevant `IWorkspaceState` as `sideContentActiveTab` - this allows middleware (e.g. the Redux saga) to be able to retrieve the side-content tab ID (a string enum of type `SideContentType`) and branch actions accordingly - this has the side-effect of being easier to test and enabling other conditional logic depending on which tab is active. Note that rendering of the tabs is still handled by the Blueprint 3 `<Tabs>` component, and this `sideContentActiveTab` attribute is not mapped from the `IWorkspaceState` to the `IWorkspaceProps`. In effect, the `sideContentActiveTab` acts as a 'cache' for the current state of the active tab in each workspace, that is updated by dispatching the `UPDATE_ACTIVE_TAB` action. Additionally, the ability to run-all works with the `Shift-Enter` key combination for 'Run' without additional modification. The drawback is that it adds more attributes to the Redux store. This is how it is currently implemented, and the branching behaviour is tested for in the saga.

(2) Create a different action with accompanying action creator, reducer and saga to handle 'Run All', e.g. `EVAL_EDITOR_AND_TESTCASES`, and then trigger it from the 'Run All' button in the `<ControlBar>` component. This requires the `<ControlBar>` to receive an additional (probably optional) prop `activeTab` and a handler `updateLocalActiveTab` from its parent `<Workspace>` which will be read and invoked from within the control bar for conditional rendering of 'Run All'. This 'Run All' will only render when the `activeTab` prop exists (we set it in the `AssessmentWorkspace` and `GradingWorkspace`) and is equal to `SideContentType.autograder`. The advantage of this approach is that it cleans up the Redux store, and allows different workspaces to implement rendering of additional buttons on an as-needed basis. However, this needs to be handled carefully for each workspace and prevents any middleware from reading the current active tab. It would also complicate testing in that I would have to additionally deeply mount an entire `<Workspace>` component for each workspace that implements 'Run All', in order to trigger event handlers to assert the 'Run All' button indeed exists in the expected conditions.

@geshuming, @alcen, @blackening and @martin-henz: Any opinions regarding the above two options?

@alcen Regarding the change made to use the `SideContentType` enum instead of raw strings (for indexing the tabs) - I used this so we can collate all the equivalent `SideContent` types into `states.ts`. This allows us to change the id of all similar tabs (e.g. 'Grading' tab in `AssessmentWorkspace` and `GradingWorkspace`) centrally if needed in the future. Any feedback regarding this?

alcen:
I like the `SideContentType` change, it should make the tabs easier to reference and make the tab IDs more consistent across the board

But I still feel that we don't need to cache the active tab state in Redux. This may complicate matters as there are now multiple ways to carry out tab changing effects, through Saga and React. I feel it is more straightforward to set the state of the parent when the tabs are switched to "Autograder", to allow for side effects like changing of CSS styles

Regarding the second way you mentioned, we don't have to do this if we put the "Run All" button in the Autograder tab right? I don't think there is a need for it to exist on the control bar.

I think it will make much more sense to create another action for it as it is fundamentally different from the editor "Run" action. Since the "Run All" is only found in the tab, there won't be issues of the button polluting other components and we can ensure that this action is only available when the active tab is "Autograder". The "Run All" would function exactly the same as how the "Run" would function now (with the tab set to "Autograder")

blackening:
> Regarding the second way you mentioned, we don't have to do this if we put the "Run All" button in the Autograder tab right? I don't think there is a need for it to exist on the control bar.

> I think it will make much more sense to create another action for it as it is fundamentally different from the editor "Run" action. Since the "Run All" is only found in the tab, there won't be issues of the button polluting other components and we can ensure that this action is only available when the active tab is "Autograder". The "Run All" would function exactly the same as how the "Run" would function now (with the tab set to "Autograder")

This particular use case is for graders. When grading, we do some quick edits to minor errors, then we need to check if all the tests pass or not. Having to click is not really fun. Run has a default keyboard shortcut. Don't really think we should add another shortcut for run-all.

An example use case for this is where i go prune the student's unnecessary code and ensure its still working.

Aulud:
> But I still feel that we don't need to cache the active tab state in Redux. This may complicate matters as there are now multiple ways to carry out tab changing effects, through Saga and React. I feel it is more straightforward to set the state of the parent when the tabs are switched to "Autograder", to allow for side effects like changing of CSS styles

I assume by _tab-changing effects_ that it refers to conditional actions that are executed based on the current active tab?

Not really in favour still of adding an extra button to either the `<ControlBar>` or the Autograder component - we do have quite a lot of buttons already in the `<ControlBar>`. Instead, I propose doing a minor rewrite to conditionally render either a 'Run' or 'Run All' button in the `<ControlBar>` based on the value of the prop `activeTab` passed from its parent workspace, which will respectively dispatch either a `EVAL_EDITOR` or `EVAL_EDITOR_WITH_TESTCASES` action.

> I think it will make much more sense to create another action for it as it is fundamentally different from the editor "Run" action. Since the "Run All" is only found in the tab, there won't be issues of the button polluting other components and we can ensure that this action is only available when the active tab is "Autograder". The "Run All" would function exactly the same as how the "Run" would function now (with the tab set to "Autograder")

Can do regarding the `EVAL_EDITOR_WITH_TESTCASES` action - will make it dispatch `EVAL_EDITOR`, block waiting for the first result then proceed with `EVAL_TESTCASE` synchronously until the first failure or all testcases are completed.

> Regarding the second way you mentioned, we don't have to do this if we put the "Run All" button in the Autograder tab right? I don't think there is a need for it to exist on the control bar.

Given the above, I'm intending to preserve the 'Run All' functionality being triggered by the `Shift+Enter` key combination in the editor when the Autograder is open - both for convenience and that @blackening mentioned a potential use case with it during grading. For this, I would need to add an additional (optional) props `activeTab` and `handleEditorEvalAll` to `IEditorProps` in order for the editor to dispatch the correct action on `Shift+Enter` based on the active tab. This removes the need for tracking `sideContentActiveTab` in the Redux store.

alcen:
> Regarding the second way you mentioned, we don't have to do this if we put the "Run All" button in the Autograder tab right? I don't think there is a need for it to exist on the control bar.
> 
> 
> I think it will make much more sense to create another action for it as it is fundamentally different from the editor "Run" action. Since the "Run All" is only found in the tab, there won't be issues of the button polluting other components and we can ensure that this action is only available when the active tab is "Autograder". The "Run All" would function exactly the same as how the "Run" would function now (with the tab set to "Autograder")
> 
> This particular use case is for graders. When grading, we do some quick edits to minor errors, then we need to check if all the tests pass or not. Having to click is not really fun. Run has a default keyboard shortcut. Don't really think we should add another shortcut for run-all.
> An example use case for this is where i go prune the student's unnecessary code and ensure its still working.

Okay, no problem. I understand the use of the Run button is for graders' convenience.
For now, since we want to keep a singular Run button, I guess this solution will have to do currently?

I worry that it will set a precedent though, for the active tabs to be handled by the Redux store instead of the state where it should belong. A lot of refactoring will then have to be done to clean it up


> Given the above, I'm intending to preserve the 'Run All' functionality being triggered by the Shift+Enter key combination in the editor when the Autograder is open - both for convenience and that @blackening mentioned a potential use case with it during grading. For this, I would need to add an additional (optional) props activeTab and handleEditorEvalAll to IEditorProps in order for the editor to dispatch the correct action on Shift+Enter based on the active tab. This removes the need for tracking sideContentActiveTab in the Redux store.

Ok. Sounds like a good solution

I've thought of an alternative: Pass down the button to the control bar in a new prop that it will render in place of Run. Ideally we would want to resolve #607 before doing so, but this could be a stopgap measure and then in future all buttons will get passed down. We could pass this down based on the state in `GradingWorkspace`, and the changing of tabs to "Autograder" will set this state. Then `ControlBar` would not need to know which tab you are on? Would this work?

Aulud:
> > Regarding the second way you mentioned, we don't have to do this if we put the "Run All" button in the Autograder tab right? I don't think there is a need for it to exist on the control bar.
> > I think it will make much more sense to create another action for it as it is fundamentally different from the editor "Run" action. Since the "Run All" is only found in the tab, there won't be issues of the button polluting other components and we can ensure that this action is only available when the active tab is "Autograder". The "Run All" would function exactly the same as how the "Run" would function now (with the tab set to "Autograder")
> > This particular use case is for graders. When grading, we do some quick edits to minor errors, then we need to check if all the tests pass or not. Having to click is not really fun. Run has a default keyboard shortcut. Don't really think we should add another shortcut for run-all.
> > An example use case for this is where i go prune the student's unnecessary code and ensure its still working.
> 
> Okay, no problem. I understand the use of the Run button is for graders' convenience.
> For now, since we want to keep a singular Run button, I guess this solution will have to do currently?
> 
> I worry that it will set a precedent though, for the active tabs to be handled by the Redux store instead of the state where it should belong. A lot of refactoring will then have to be done to clean it up
> 
> > Given the above, I'm intending to preserve the 'Run All' functionality being triggered by the Shift+Enter key combination in the editor when the Autograder is open - both for convenience and that @blackening mentioned a potential use case with it during grading. For this, I would need to add an additional (optional) props activeTab and handleEditorEvalAll to IEditorProps in order for the editor to dispatch the correct action on Shift+Enter based on the active tab. This removes the need for tracking sideContentActiveTab in the Redux store.
> 
> Ok. Sounds like a good solution
> 
> I've thought of an alternative: Pass down the button to the control bar in a new prop that it will render in place of Run. Ideally we would want to resolve #607 before doing so, but this could be a stopgap measure and then in future all buttons will get passed down. We could pass this down based on the state in `GradingWorkspace`, and the changing of tabs to "Autograder" will set this state. Then `ControlBar` would not need to know which tab you are on? Would this work?

Nice! Creating the `<Button>` in the workspace with all the props it needs and passing it to the `<ControlBar>` is a good solution and allows `ControlBar` to be ignorant of the `activeTab`. Will create an issue to log this and refactor this myself when #607 is resolved - shouldn't be too big a task. Thanks!

alcen:
> Nice! Creating the <Button> in the workspace with all the props it needs and passing it to the <ControlBar> is a good solution and allows ControlBar to be ignorant of the activeTab. Will create an issue to log this and refactor this myself when #607 is resolved - shouldn't be too big a task. Thanks!

No problem! Happy to help

Aulud:
@geshuming Have fixed the merge conflicts.

Aulud:
> > Nice! Creating the  in the workspace with all the props it needs and passing it to the  is a good solution and allows ControlBar to be ignorant of the activeTab. Will create an issue to log this and refactor this myself when #607 is resolved - shouldn't be too big a task. Thanks!
> 
> No problem! Happy to help

@alcen @geshuming Have opened #812 to keep track of the refactor to-dos. Will take a look at the `ControlBar` changes before proceeding.

PR #807, opened 2019-08-04
Refactor for ControlBar component

### LINKED ISSUE #607: Consolidate buttons as a prop to be passed down from workspace

alcen:
`ControlBar` is rapidly becoming overloaded with props like `hasShareButton`, `hasSaveButton` etc. especially with the additions of new features. With this pull request, all the buttons have been made into their own component and dispatched to `ControlBar` instead

This is meant to resolve #607, and is mostly a change of code without any change in behaviour. However, I did take the liberty to make some small changes:
- The frontend will now clear the 'Save' button CSS properly on assessments if run without the backend in a development setting
- Toggle mode button for Mission Control is now capitalised, like the rest of the `ControlBar` buttons


alcen:
> Can you rename the directory to something other than 'controlBarButtons'. I don't think all the components are buttons

Ok, no problem. What do you suggest? Would 'controlBarComponents' work?

geshuming:
Sure. I think just 'controlBar' will do, let's keep it short.

alcen:
> Sure. I think just 'controlBar' will do, let's keep it short.

Ok. This sound better, I'll go with this!

src/components/missionControl/EditingWorkspace.tsx
```diff
@@ -151,7 +162,7 @@ class AssessmentWorkspace extends React.Component<AssessmentWorkspaceProps, ISta
     const questionId = this.formatedQuestionId();
     const question: IQuestion = this.state.assessment.questions[questionId];
     const workspaceProps: WorkspaceProps = {
-      controlBarProps: this.controlBarProps(this.props, questionId),
+      controlBarProps: this.controlBarProps(this.props, this.state, questionId),
```
geshuming: Why is the state being passed into the class function? Just reference it from within the class

alcen: There was actually a quirk with this function in which it took props as an argument, but the body referenced `this.props` instead of the parameter `props`

Not really sure what was originally intended but I fixed it to use the props that were passed into it, and then I decided that the state could be handled this way as well for consistency

Would it be a good idea to just not pass either props or state? Or is there a reason to pass only props

geshuming: I think the props shouldn't be passed to. The class function can reference the props and state using `this`

alcen:
> I feel like this is relatively well done. Thanks for taking the time to refactor the control bar and buttons.

No problem! As for the issues: I have resolved them for now

src/components/missionControl/EditingWorkspace.tsx
alcen: Ok, this can be done. I suppose it can be changed to be this way for `AssessmentWorkspace`, `GradingWorkspace` as well?

PR #811, opened 2019-08-06
Fix for "controlBar" import reference

alcen:
On some systems, the script handling imports does not redirect to "controlBar/index.tsx" for imports from "/controlBar". This change is meant as a fix for that error


geshuming:
Actually, why not just put the ControlBar.tsx inside the controlBar folder

alcen:
> Actually, why not just put the ControlBar.tsx inside the controlBar folder

It's a little strange, but ControlBar.tsx does not actually use anything found inside the controlBar folder

alcen:
How about renaming the folder to 'controlElements', and leaving ControlBar.tsx as it is? Would it be apparent that control elements are meant to be used in the control bar?

geshuming:
in the end everything in that folder only goes into the ControlBar component, it just makes sense to group them together

alcen:
> in the end everything in that folder only goes into the ControlBar component, it just makes sense to group them together

Ok yes, I agree. Will do this

alcen:
> You copied the ControlBar.tsx to the folder, not moved it.

Ok, the old one is deleted now

Issue #681, opened 2019-07-07
From 2018 team post-mortem: Profile page which can look at all the missions, all the grades, all the xp


martin-henz:
*Note from Martin:* I strongly support this. I think this is a low-hanging fruit, and one of the strongest reasons for having such a system: Give students a dashboard of their performance. The students should get reassurance that they are making progress in the module. So yes, I would rank this item as high as the removal of JSON from the database.

martin-henz:
I think this issue is still relevant. Any thoughts, anyone?

Aulud:
Addressed by #813, #854 and #898.

PR #813, opened 2019-08-06
Redesign Profile component to a full-fledged user profile page

Aulud:
## Redesign Profile component to a full-fledged user profile page
Summary: Redesign the current Profile component (a dialog prompt) to a full-fledged user profile page (side-panel from the left). Addresses #681.

### Changelog
- Update the Profile component from using bp3's `Dialog` component to use bp3's `Drawer` component - this opens a full-height sidebar overlaying the Academy from the left of the screen
   - The Profile component is still triggered from the dropdown at the top right of the Academy
- Amend the Profile component to compute the current and maximum grades and XP from `assessmentOverviews` of `ISessionState` of the Redux store, instead of displaying the `grade`, `xp`, `maxGrade` and `maxXp` attributes from `ISessionState`
   - This change renders the four above attributes redundant in the Redux store - they can be cleaned up along with the `SET_USER` action during refactoring
   - As a result, the Profile component is now entirely rendering derived data from `assessmentOverviews` - this improves the accuracy of the rendered total grade and XP, as the profile data agrees with the state of the assessment overviews and is updated whenever the assessment overviews are re-fetched
   - Further inspection of the `SET_USER` action suggests that it has never updated the `Xp`, `maxGrade` and `maxXp` attributes in the Redux store - this caused the `Profile` component to break during the past two semesters (where it would always render 0 for the grade and XP progress bars)
   - **Caveat**: Only closed assessments (with `status` of `AssessmentStatuses.submitted`) factor into the computation of total and max grade and XP, since open assessments are not yet graded
      - Additionally, **only Missions* factor into the computation of total and max Grade; see https://github.com/source-academy/cadet-frontend/pull/813#issuecomment-519070480
- Enable Profile component to trigger the `FETCH_ASSESSMENT_OVERVIEWS` action if `assessmentOverviews` are not loaded (e.g. upon just logging into the Academy, prior to loading the 'Missions' page)
- Render the proportion of total grade and XP obtained out of their respective current maximums as a fixed bp3 `Spinner`
- Render each closed assessment in a condensed card format using bp3's `Callout` component, displaying its `type` (as an icon), `title`, `grade`, `XP`, `maxGrade` and `maxXp`
   - The proportion of Grade and XP obtained is rendered as a fixed progress bar; Grade progress bar is **only rendered for Missions**, see https://github.com/source-academy/cadet-frontend/pull/813#issuecomment-519070480
   - Progress bars are coloured with custom shades of blue according to the table in https://github.com/source-academy/cadet-frontend/pull/813#issuecomment-519078600 - thanks @alcen!
   - **Caveat**: The corresponding progress bar is not rendered if both a value and its maximum value are 0, e.g. if `grade` is `0` and `maxGrade` is `0`, that assessment's card will not render its XP progress bar - this ensures bonus XP on assessments with `maxXp` of `0` still render the XP bar
- Allow clicking a closed assessment's card to directly open that assessment in a new browser tab, for quick user review
- Add a placeholder to render when there are no closed assessments

### Bugfixes
- Fix non-standard module import declaration in test file for 'Autograder' component

### Mocks and Tests
- Update mock assessments to render mock assessment cards with different grade and XP boundaries in the profile
- Update existing snapshots and add DOM rendering tests for the Profile component

#### Before (Profile dialog prompt)
profile-old

#### After (no closed assessments - placeholder rendered)
profile-new-empty

#### After (closed assessments exist - 1536 x 864 on a 15" laptop)
profile-new-populated-rev1

#### After (closed assessments exist and list of cards overflows - 1536 x 864 on a 15" laptop)
profile-new-populated-overflow-rev1

#### After (closed assessments exist - 1920 x 1080 on a 22" screen)
profile-new-populated-large-rev1

#### Feature demonstration (GIF - not updated to latest patch)
profile-page-new-feature-demo

Last updated 8 Aug 2019, 8:30PM

### DIFF
--- a/src/components/dropdown/Profile.tsx
+++ b/src/components/dropdown/Profile.tsx
@@ -1,14 +1,16 @@
 import {
+  Callout,
   Drawer,
   Intent,
   NonIdealState,
+  ProgressBar,
   Spinner
 } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import * as React from 'react';
 
 import { Role } from '../../reducers/states';
-import { AssessmentStatuses, IAssessmentOverview } from '../assessment/assessmentShape';
+import { AssessmentCategories, AssessmentCategory, AssessmentStatuses, IAssessmentOverview } from '../assessment/assessmentShape';
 
 type ProfileProps = OwnProps & StateProps;
 
@@ -52,6 +54,62 @@ class Profile extends React.Component<ProfileProps> {
                 : Intent.DANGER;
       };
 
+      const renderIcon = (category: AssessmentCategory) => {
+        switch (category) {
+          case AssessmentCategories.Mission:
+            return IconNames.FLAME;
+          case AssessmentCategories.Sidequest:
+            return IconNames.LIGHTBULB;
+          case AssessmentCategories.Path:
+            return IconNames.PREDICTIVE_ANALYSIS;
+          case AssessmentCategories.Contest:
+            return IconNames.COMPARISON;
+          default:
+            // For rendering hidden assessments not visible to the student
+            // e.g. studio participation marks
+            return IconNames.PULSE;
+          }
+      };
+
+      // Build condensed assessment cards from an array of assessments
+      const summaryCallouts = this.props.assessmentOverviews!
+        .filter( (item) => item.status === AssessmentStatuses.submitted )
+        .map( (item) => {
+          return (
+            <Callout
+                className='profile-summary-callout'
+                key={`${item.title}-${item.id}`}
+                icon={renderIcon(item.category)}
+                title={item.title}
+              >
+              {item.maxGrade <= 0 ? '' :
+                <div className='grade-details'>
+                  <div className='title'>Grade</div>
+                  <div className='value'>{item.grade} / {item.maxGrade}</div>
+                  <ProgressBar
+                    animate={false}
+                    className='value-bar'
+                    intent={parseFrac(item.grade / item.maxGrade)}
+                    stripes={false}
+                    value={item.grade / item.maxGrade} />
+                </div>
+              }
+              {item.maxXp <= 0 ? '' :
+                <div className='xp-details'>
+                  <div className='title'>XP</div>
+                  <div className='value'>{item.xp} / {item.maxXp}</div>
+                  <ProgressBar
+                    animate={false}
+                    className='value-bar'
+                    intent={parseFrac(item.xp / item.maxXp)}
+                    stripes={false}
+                    value={item.xp / item.maxXp} />
+                </div>
+              }
+            </Callout>
+          );
+        });
+
       // Compute the user's maximum total grade and XP from submitted assessments
       content = (
         <div className='profile-content'>
@@ -88,6 +146,9 @@ class Profile extends React.Component<ProfileProps> {
               </div>
               <div className="percentage">{((currentXp / maxXp) * 100).toFixed(2)}%</div>
             </div>
+          </div>
+          <div className="profile-callouts">
+              {summaryCallouts}
           </div>
         </div>
       );
--- a/src/components/dropdown/Profile.tsx
+++ b/src/components/dropdown/Profile.tsx
@@ -12,7 +12,7 @@ import * as React from 'react';
 import { Role } from '../../reducers/states';
 import { AssessmentCategories, AssessmentCategory, AssessmentStatuses, IAssessmentOverview } from '../assessment/assessmentShape';
 
-type ProfileProps = OwnProps & StateProps;
+type ProfileProps = DispatchProps & OwnProps & StateProps;
 
 export type StateProps = {
   assessmentOverviews?: IAssessmentOverview[];
@@ -25,7 +25,18 @@ export type OwnProps = {
   onClose: () => void;
 };
 
+export type DispatchProps = {
+  handleAssessmentOverviewFetch: () => void;
+};
+
 class Profile extends React.Component<ProfileProps> {
+  public componentDidMount() {
+    if (!this.props.assessmentOverviews) {
+      // If assessment overviews are not loaded, fetch them
+      this.props.handleAssessmentOverviewFetch();
+    }
+  }
+
   public render() {
     const isLoaded = this.props.name && this.props.role && this.props.assessmentOverviews;
     let content: JSX.Element;
--- a/src/components/dropdown/Profile.tsx
+++ b/src/components/dropdown/Profile.tsx
@@ -71,7 +71,14 @@ class Profile extends React.Component<ProfileProps, {}> {
         const [currentGrade, currentXp, maxGrade, maxXp] = this.props.assessmentOverviews!.reduce(
           (acc, item) =>
             item.status === AssessmentStatuses.submitted
-              ? [acc[0] + item.grade, acc[1] + item.xp, acc[2] + item.maxGrade, acc[3] + item.maxXp]
+              ? item.category === AssessmentCategories.Mission
+                ? [
+                    acc[0] + item.grade,
+                    acc[1] + item.xp,
+                    acc[2] + item.maxGrade,
+                    acc[3] + item.maxXp
+                  ]
+                : [acc[0], acc[1] + item.xp, acc[2], acc[3] + item.maxXp]
               : acc,
           [0, 0, 0, 0]
         );
--- a/src/components/dropdown/Profile.tsx
+++ b/src/components/dropdown/Profile.tsx
@@ -1,16 +1,15 @@
-import { Callout, Drawer, Intent, NonIdealState, ProgressBar, Spinner } from '@blueprintjs/core';
+import { Drawer, Intent, NonIdealState, Spinner } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import * as React from 'react';
 
-import { NavLink } from 'react-router-dom';
 import { Role } from '../../reducers/states';
-import { assessmentCategoryLink } from '../../utils/paramParseHelpers';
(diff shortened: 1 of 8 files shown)


Aulud:
> I think it looks good! One qualm that I have is about the use of colours, it might be discouraging to look at so many orange and red tones if you aren't doing well. Could I suggest the use of different shades of blue instead?

Using different shades of blue will require manual CSS styling of the bp3 `Spinner` and `ProgressBar` components - currently the colour of these components are rendered by setting their `Intent` prop according to the percentage of the grade/XP scored out of the maximum.

| Intent value | Percentage scored | CSS hex | Description (closest CSS colour) |
| :---: | :---: | :---: | :---: |
| `Intent.PRIMARY` | >=90% | #137CBD | Steel blue |
| `Intent.SUCCESS` | >=80%, <90% | #0F9960 | Sea green |
| `Intent.WARNING` | <=60%, <80% | #D9822B | Dark orange |
| `Intent.DANGER` | <60% | #DB3737 | Brown red |

Instead of using different shades of blue, may I propose dropping the bracket for percentages above 90% (dropping the first row in the table above), and retaining the other brackets with cut-off boundaries of 80% and 60%. Would this be more forgiving overall?

martin-henz:
Yes, > 85% would be steel blue
65 - 85% sea green
45 - 65% dark orange
< 45% brown red

This gives differentiation at the top and warning signals at a danger mark: 65% is acceptable but not great. Students below 65% should get an indication that they are below our expectation. Below 45% is really serious, and should be indicated as red.

alcen:
> Yes, > 85% would be steel blue
> 65 - 85% sea green
> 45 - 65% dark orange
> < 45% brown red
> This gives differentiation at the top and warning signals at a danger mark: 65% is acceptable but not great. Students below 65% should get an indication that they are below our expectation. Below 45% is really serious, and should be indicated as red.

Some thoughts I had on this:

I feel like if you are doing badly, and the grades matter to you, you will probably be very aware that you are not doing well and not need the system to remind you. Most report cards do not display colours, and the public education system (primary and secondary school: specifically the PSLE, O Level examinations) has even been moving away from showing definite percentages in favour of vague letter grades - this has changed a lot compared to when I was personally taking these tests!

Also, we have to keep in mind that the display currently counts both Missions and Quests together. Having a perfect score on all missions, but not attempting the quests at all will put you in the second bracket, which may not be a good indication of a student's actual ability

Overall, I think this suggested colour scheme is an improvement over the original, but I still think the colour doesn't add much value to the system. That said, I think green and blue should be switched, blue being a more "neutral" colour. A green colour feels more "successful" to me than blue, haha

Keeping this in mind, if we decide to keep the colours, I like the idea of red below 45%, blue for 45% - 80%, and green for >80%. Maybe a personalised message could even be added, like "Keep trying!" or "You're doing great!". Any thoughts?

martin-henz:
I agree. Let's avoid the green-orange-red cliche. Colors ok, but as suggested shades of the same color.

Aulud:
> > Yes, > 85% would be steel blue
> > 65 - 85% sea green
> > 45 - 65% dark orange
> > < 45% brown red
> > This gives differentiation at the top and warning signals at a danger mark: 65% is acceptable but not great. Students below 65% should get an indication that they are below our expectation. Below 45% is really serious, and should be indicated as red.
> 
> Some thoughts I had on this:
> 
> I feel like if you are doing badly, and the grades matter to you, you will probably be very aware that you are not doing well and not need the system to remind you. Most report cards do not display colours, and the public education system (primary and secondary school: specifically the PSLE, O Level examinations) has even been moving away from showing definite percentages in favour of vague letter grades - this has changed a lot compared to when I was personally taking these tests!
> 
> Also, we have to keep in mind that the display currently counts both Missions and Quests together. Having a perfect score on all missions, but not attempting the quests at all will put you in the second bracket, which may not be a good indication of a student's actual ability
> 
> Overall, I think this suggested colour scheme is an improvement over the original, but I still think the colour doesn't add much value to the system. That said, I think green and blue should be switched, blue being a more "neutral" colour. A green colour feels more "successful" to me than blue, haha
> 
> Keeping this in mind, if we decide to keep the colours, I like the idea of red below 45%, blue for 45% - 80%, and green for >80%. Maybe a personalised message could even be added, like "Keep trying!" or "You're doing great!". Any thoughts?

@alcen Any suggestions for the three shades of blue (as hex codes)? I'm not good at picking UI colours haha. Alternatively, we could leave all the `Spinner`s and `ProgressBar`s with the default dark blue colour from `Intent.PRIMARY` for now.

Regarding the note about Missions and Quests being factored into the Profile display - I have the following proposal.
- For computation of the total and maximum grade, **only Missions** will be counted - then the 'Grade' spinner serves the purpose of a _rough gauge_ of the overall proportion of the module's CA scored _thus far_
   - With this change, the 'Grade' progress bar will no longer be rendered for non-Missions
- For computation of the total and maximum XP, **all components** will be counted - then the 'XP' spinner serves the purpose of a _rough gauge_ of the student's progress towards the bonus 1MC awarded for extra achievement _thus far_
   - This includes Missions, Quests, Paths, Contests, bonus XP awarded from early submissions, tutorial participation XP (dummy assessments will be mounted for Avengers to enter their studio participation scores)
   - Not all sources of XP may be reflected here as some are awarded manually (e.g. contest winners, XP for contribution to the Academy's repositories)

Please provide feedback RE: the above. Thanks!

alcen:
> @alcen Any suggestions for the three shades of blue (as hex codes)? I'm not good at picking UI colours haha. Alternatively, we could leave all the Spinners and ProgressBars with the default dark blue colour from Intent.PRIMARY for now.

Ok. I suggest:

|              | Percentage   | Hex Code                            | Closest Colour |
|--------------|--------------|-------------------------------------|----------------|
| Highest Tier |  >=80%       | #137CBD (default of `INTENT.Primary`) | Steel blue     |
| Middle Tier  | >=45% & <80% | #579ECB                             | Deep sky blue  |
| Lowest Tier  | <45%         | #9AC0D8                             | Sky blue       |

&nbsp;
> Regarding the note about Missions and Quests being factored into the Profile display - I have the following proposal.
> 
> For computation of the total and maximum grade, only Missions will be counted - then the 'Grade' spinner serves the purpose of a rough gauge of the overall proportion of the module's CA scored thus far
> 
> With this change, the 'Grade' progress bar will no longer be rendered for non-Missions
> 
> 
> For computation of the total and maximum XP, all components will be counted - then the 'XP' spinner serves the purpose of a rough gauge of the student's progress towards the bonus 1MC awarded for extra achievement thus far
> 
> This includes Missions, Quests, Paths, Contests, bonus XP awarded from early submissions, tutorial participation XP (dummy assessments will be mounted for Avengers to enter their studio participation scores)
> Not all sources of XP may be reflected here as some are awarded manually (e.g. contest winners, XP for contribution to the Academy's repositories)
> 
> 
> 
> Please provide feedback RE: the above. Thanks!

This sounds good! I'm not very sure what the grade score of non-Missions are used for, but it seems like it wouldn't do any harm to hide it from the profile UI

martin-henz:
> This sounds good! I'm not very sure what the grade score of non-Missions are used for, but it seems like it wouldn't do any harm to hide it from the profile UI

Yes, hide. Much clearer if it's just the missions. That way this score corresponds to the CA component for missions.

Aulud:
Hi all, have patched in the above changes from https://github.com/source-academy/cadet-frontend/pull/813#issuecomment-519070480 and https://github.com/source-academy/cadet-frontend/pull/813#issuecomment-519078600 - please refer to the updated screenshots in the initial post. Thanks!

Aulud:
> This sounds good! I'm not very sure what the grade score of non-Missions are used for, but it seems like it wouldn't do any harm to hide it from the profile UI

The total grade for non-Missions is actually not used anywhere haha - it's a by-product of how the assessment XML, XP and autograder systems are currently set up.

As an aside regarding the assessment and grading system:
- During authoring of an assessment XML file, one can elect to specify one or more public or private testcases for each question
- Each testcase has a mandatory `grade` attribute (see [here](https://github.com/source-academy/assessments/wiki/Assessments) for more details of the XML format), allowing assessment authors to specify the relative weighing between testcases
   - Presumably, more challenging testcases or testcases for edge cases would have a higher relative `score`
   - `score` can also be set to `0` if the question or assessment is not meant to be graded, e.g. for Paths
- By default, the `maxgrade` attribute for each question is computed from the sum of the `grade`s of all its testcases
   - This value may be set by the author to be greater than the sum, if it is intended that manual review assign the remaining points in excess of the sum
- On an assessment level, its `max_grade` (virtual) field is also computed from the sum of the `max_grade`s of its questionis
- After a student's submission is finalised (either intentionally with 'Finalise Submission' or auto-submission on the deadline), the autograder runs all public and private testcases for each submission and returns a `Summary` object containing the `totalScore` for that question
- **The backend then assigns the XP awarded for a question based on the proportion of grade awarded by the autograder**

In my opinion, the above part in bold couples parts of the Grade and XP systems together, and this design decision is probably worth taking a second look at in the future.

> 👍 All good! This is a great change, students will enjoy the ability to see all their scores at one go. I don't see any issues with the code.
> 
> Something I am not 100% sure about: How come we need to check if `this.props.assessmentOverviews` is defined? I'm assuming that Redux doesn't load it in immediately

Yup! This handles an edge case where the `assessmentOverviews` have not been loaded into the Redux store, causing the Profile component to eternally render the `NonIdealState` spinner (alternatively the Profile component could always just trigger `handleAssessmentOverviewFetch`). This occurs primarily when a user has just logged into the Academy and navigates to either the Story page or the Playground first - since the `assessmentOverviews` were previously only fetched by `assessmentWorkspace` when it is loaded (e.g. by navigating to 'Missions').

PR #808, opened 2019-08-05
Conditional tab rendering in Playground

alcen:
In this commit, tabs in the Playground are now tied to their respective chapters. For example, Source §1 now only shows the "Introduction" tab.

A screenshot of what happens when 'Source §2' is selected in the Playground:

condiiton-tabs

This change is done in line with the future addition of the Substituter feature, which is only enabled for Source §2 and below - see #803

Note: disabling of tabs based on chapters is not done for Sourcecast and Sourcereel as I am not sure how it would affect the recording. If the developers for these features would advise how to do so, that would be great!

Note 2: this Pull Request was originally full of different unrelated changes for Playground side content. The changes have been since rebased into their own PRs:

| Description                               | Pull Request No.    |
|---------------------------------------|------|
| Assessments default tab fix     | #819 |
| Conditional tab rendering       | #808 |
| Default tab text                | #820 |
| List visualiser scrolling fix   | #817 |
| List visualiser orange flashing | #818 |

&nbsp;

Thanks to @lumos309 for a bug fix for a Playground crash when the editor was trying to find the non-existent Inspector/Env Visualiser tabs!

PR #818, opened 2019-08-09
Orange flashing for data visualiser when used

alcen:
In the commit, the Data Visualiser was modified such that the tab shows orange flashing when used, in line with the Inspector and Env Visualiser features, as well as the upcoming Substituter.

Note: this pull request was previously split from #808. You may view the comments history related to this pull request there


public/externalLibs/visualizer/visualizer.js
```diff
@@ -871,6 +871,10 @@
    *  Then shift it to the left end.
    */
   function draw(xs) {
+    // Blink icon
+    const icon = document.getElementById('data_visualiser-icon');
+    icon.classList.add('side-content-tab-alert');
```
geshuming: Do a typecheck here.

```suggestion
if (icon) {
  icon.classList.add('side-content-tab-alert');
}
```

Issue #283, opened 2018-08-10
Data visualiser should only appear after a call to draw_list


ning-y:
In order of preference,

1. The list visualiser appears in the repl output in the same way that a rune does, or
2. The list visualiser tab does not render (including the button to switch to it) until a call to `draw_list`.
    - Calling draw_list also automatically switches the tab to the list visualiser.

martin-henz:
Alternatively: The data visualiser should be always there (from Source §2 onwards), and show a default help text before draw_data is called the first time:

The data visualiser visualises data structures. It is activated by calling the function draw_data(the_data), where the_data is that data structure that you want to visualise.

angelsl:
This seems to no longer be a problem.

PR #820, opened 2019-08-09
Default texts for data visualiser, inspector, env visualiser

alcen:
In this commit, there are now default texts shown for Data Visualiser, Inspector and Env Visualiser. This text disappears when the respective components are used for the first time

These default texts would resolve issue #283

A screenshot of what happens when 'Data Visualizer' is selected in the Playground:

data-viz-default-text

Note: this pull request was previously split from #808. You may view the comments history related to this pull request there


alcen:
> I was wondering if we can keep the default text even when the component is being used.

We can, but I don't think there is a need to? It will only crowd the interface

martin-henz:
> > I was wondering if we can keep the default text even when the component is being used.
> 
> We can, but I don't think there is a need to? It will only crowd the interface

Yes, the text has useful info, but I agree with @alcen : It would take up screen space and would add the need to scroll frequently. At the moment, it's nicely satisfying the curiosity of the user who clicks around to see what's there.

martin-henz:
@alcen can you look at the conflicts?

alcen:
> @alcen can you look at the conflicts?

Ok, seems like this has resolved it!

PR #826, opened 2019-08-10
Fix default tab of ungraded assessments

alcen:
In pull request #819, the default tab of graded assessments was set to the grading results. However, I accidentally set the tab of ungraded assessments to 'Mission Briefing' instead of the original 'Task 1/2/...'.

This commit fixes the unintended change!


Aulud:
Small fix; LGTM.

PR #841, opened 2019-08-13
Use AWS S3 for file storage

jiayushe:
- Relevant to source-academy/cadet#468


jiayushe:
> What's this?

The url generated is now pointing to S3, so no need to prepend our backend url any more.

jiayushe:
Hi @geshuming, can we merge in this PR? The backend PR is ready.

PR #849, opened 2019-08-14
Add display waveform function

TheRealRavager:
Adds display waveform function which accepts a sound and displays it using the curves library


martin-henz:
Yes, very nice function. I'm happy with it and have seen it in action. 

A suggestion for discussion: How about using a relative number as parameter: number_of_points_per_second instead of the absolute number_of_points ?

That way the user can relate the number directly to the sample rate used in audio processing. For example, common values would be as given here:
https://en.wikipedia.org/wiki/Sampling_(signal_processing)#Sampling_rate

TheRealRavager:
> Yes, very nice function. I'm happy with it and have seen it in action.
> 
> A suggestion for discussion: How about using a relative number as parameter: number_of_points_per_second instead of the absolute number_of_points ?
> 
> That way the user can relate the number directly to the sample rate used in audio processing. For example, common values would be as given here:
> https://en.wikipedia.org/wiki/Sampling_(signal_processing)#Sampling_rate

Agreed, that would be more meaningful for students

PR #863, opened 2019-08-15
Fix scroll bar not clickable when Mission briefing is open

alcen:
A fix for an innate BlueprintJS issue: the Dialog component does not allow you to click on the scroll bar. Currently mouse scrolling, arrow keys, or Page Up/Page Down keys have to be used to navigate the Mission briefing, which uses the Blueprint Dialog

Also, I added some comments to explain the CSS fixes

### DIFF
--- a/src/styles/_commons.scss
+++ b/src/styles/_commons.scss
@@ -33,6 +33,26 @@
   }
 }
 
+/*
+  Force Blueprint Dialog to scroll on focus
+
+  Note: there should exist a better fix, as the Blueprint Dialog behaves
+  properly on the official demo/documentation website
+  https://blueprintjs.com/docs/#core/components/dialog
+*/
 .#{$ns}-dialog-container {
   pointer-events: all;
+}
+
+/*
+  Temporary fix for this issue: https://github.com/palantir/blueprint/issues/1008
+
+  (when Blueprint Dialog is open, browser scroll bar is not clickable)
+*/
+.#{$ns}-overlay-backdrop {
+  display: none;
+}
+
+.#{$ns}-overlay-open {
+  background-color: rgba(16, 22, 26, 0.7);
 }


alcen:
> Haha it seems the Dialog is broken in the same way on BpJS' official documentation - just that it's not immediately obvious as their demonstration sets the Dialog to close if a user clicks outside of it.
> This bug seems to be originating from their mode of use of Portal in the Dialog implementation. Given this is an issue that isn't yet resolved by BpJS, we will have to stick with the temporary fix from the issue on BpJS' repo (it works in Chrome and Firefox so I guess it's good to go).

Yes haha, but only the unclickable scroll bar seems to be a bug with Blueprint. The issue with the dialog not being scrollable when not in focus (by clicking on the overlay) is not present on their demo, but we haven't found another fix for it!!

Aulud:
> > Haha it seems the Dialog is broken in the same way on BpJS' official documentation - just that it's not immediately obvious as their demonstration sets the Dialog to close if a user clicks outside of it.
> > This bug seems to be originating from their mode of use of Portal in the Dialog implementation. Given this is an issue that isn't yet resolved by BpJS, we will have to stick with the temporary fix from the issue on BpJS' repo (it works in Chrome and Firefox so I guess it's good to go).
> 
> Yes haha, but only the unclickable scroll bar seems to be a bug with Blueprint. The issue with the dialog not being scrollable when not in focus (by clicking on the overlay) is not present on their demo, but we haven't found another fix for it!!

Eh the scrollbar's broken for me for their Dialog demo, regardless of which options I pick haha

alcen:
> Haha it seems the Dialog is broken in the same way on BpJS' official documentation - just that it's not immediately obvious as their demonstration sets the Dialog to close if a user clicks outside of it.
> This bug seems to be originating from their mode of use of Portal in the Dialog implementation. Given this is an issue that isn't yet resolved by BpJS, we will have to stick with the temporary fix from the issue on BpJS' repo (it works in Chrome and Firefox so I guess it's good to go).
> 
> Yes haha, but only the unclickable scroll bar seems to be a bug with Blueprint. The issue with the dialog not being scrollable when not in focus (by clicking on the overlay) is not present on their demo, but we haven't found another fix for it!!
> 
> Eh the scrollbar's broken for me for their Dialog demo, regardless of which options I pick haha

As in, if you click outside and try to scroll with the arrow keys/scroll wheel/Page Up/Down keys, it works (but not on the Source Academy without the `pointer-events: all` fix)

Aulud:
> > Haha it seems the Dialog is broken in the same way on BpJS' official documentation - just that it's not immediately obvious as their demonstration sets the Dialog to close if a user clicks outside of it.
> > This bug seems to be originating from their mode of use of Portal in the Dialog implementation. Given this is an issue that isn't yet resolved by BpJS, we will have to stick with the temporary fix from the issue on BpJS' repo (it works in Chrome and Firefox so I guess it's good to go).
> > Yes haha, but only the unclickable scroll bar seems to be a bug with Blueprint. The issue with the dialog not being scrollable when not in focus (by clicking on the overlay) is not present on their demo, but we haven't found another fix for it!!
> > Eh the scrollbar's broken for me for their Dialog demo, regardless of which options I pick haha
> 
> As in, if you click outside and try to scroll with the arrow keys/Page Up/Down keys, it works
> Also, if you click on the text and use the scroll wheel, it works (but these don't really work well on the Source Academy without the `pointer-events: all` fix)

Oops! Okay, gotcha

PR #865, opened 2019-08-15
try-catch updateContext to prevent non-existent getElement

blackening:
This breaks the mission control.


seanlowjk:
There seems to be merge conflicts at the moment. 

And I'm not sure what was of the final status quo with regards to the inspector. 

Should we rewrite it instead while refactoring the codebase?

blackening:
It's old enough that this is stale. Please update to new inspector. If it breaks the mission editor again I'll just submit another patch.

PR #868, opened 2019-08-16
Allow all source chapters but default to 1

openorclose:


.env.example
```diff
@@ -3,3 +3,4 @@ REACT_APP_VERSION=$npm_package_version
 REACT_APP_BACKEND_URL=http://localhost:4001
 REACT_APP_USE_BACKEND=TRUE
 REACT_APP_CHATKIT_INSTANCE_LOCATOR=instance_locator_here_otherwise_empty_string
+DEFAULT_SOURCE_CHAPTER=1
```
geshuming: Can we just leave this in the constants file, don't put it in the .env

The .env file should be used for secrets

PR #873, opened 2019-08-16
Fix 'Save & Continue' functionality in grading

alcen:
This commit fixes the "Save & Continue" button in grading to consistently bring you to the correct page.

Previously, the function updated the URL with the wrong value. Now, the function will simply try to get the current URL, and add 1 to the question number

### DIFF
--- a/src/mocks/backend.ts
+++ b/src/mocks/backend.ts
@@ -18,6 +18,7 @@ import { store } from '../createStore';
 import { IState, Role } from '../reducers/states';
 import { history } from '../utils/history';
 import { showSuccessMessage, showWarningMessage } from '../utils/notification';
+import { stringParamToInt } from '../utils/paramParseHelpers';
 import { mockAssessmentOverviews, mockAssessments } from './assessmentAPI';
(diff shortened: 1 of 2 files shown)


alcen:
> through componentDidUpdate of <GradingWorkspace> anyway.
> const currentQuestion = yield select(
>   (state: IState) => state.workspaces.grading.currentQuestion
> );
> yield history.push(`/academy/grading/${submissionId}/${currentQuestion + 1}`);

OK! No problem, will try this instead

Aulud:
> OK! No problem, will try this instead

I just realised I didn't really explain my preference here haha. Whilst I can't say I like the solution to use `currentQuestion` from the Redux store, Redux Saga here is a middleware and as such should only observe and interface with the store, not the browser window. This minimises additional coupling between the different layers of components (structural, presentational and behavioural), plus it avoids the use of regex.

Fun out-of-topic reading: regex is powerful, but also pretty dangerous if not written correctly
https://blog.cloudflare.com/details-of-the-cloudflare-outage-on-july-2-2019/

> Yes, this was the intention. If the saving fails, the comment would not be discarded

Okay, nice! LGTM, so I'll merge it in.

PR #874, opened 2019-08-16
Fix Blueprint Toast having dark background

alcen:
A small fix for the dark background present on these notifications:

bp3-overlay-open-fix

This is a consequence of the workaround introduced in #863. Apparently the notifications, which use the Blueprint Toast component, use the `bp3-overlay-open` CSS style as well

### DIFF
--- a/src/styles/_commons.scss
+++ b/src/styles/_commons.scss
@@ -55,4 +55,8 @@
 
 .#{$ns}-overlay-open {
   background-color: rgba(16, 22, 26, 0.7);
+
+  &.#{$ns}-toast-container {
+    background-color: unset;
+  }
 }


alcen:
> This is why I didn't really like the previous fix. Regression issues may occur. Do we know what else might use overlay-open?

It is also used by other popovers like the tooltips for "Run" for the editor, but these elements have 0 width and are invisible. The fix is not meant to be pretty anyway, since it is fundamentally an issue with the components we are using

geshuming:
How about we set a max height to the dialog container and set the internal display to use scroll?

alcen:
> How about we set a max height to the dialog container and set the internal display to use scroll?

It seems to just create multiple scroll bars:

scroll

alcen:
> Anyway, the temporary fix for this issue from palantir/blueprint#1008 seems to additionally break the ability to close the Profile (or any other component with the prop canOutsideClickClose set to true) by clicking outside it - I'm not sure keeping the temporary fix is the best solution until bp3 fixes their Overlay component for good.

Uh, this seems to already be broken on the current deployed version (without the temporary fix). When you click "About", there is no way to close it without pressing "Esc". when previously clicking outside would do so

> @alcen Hmm, I'm impartial to @geshuming's suggestion but that's a really odd behaviour.

Yes, I agree, not sure why it happens though? Maybe because Blueprint uses the `overlay-open` CSS on the body element? Idk really

Aulud:
@geshuming Have not found any other regression apart from the loss of functionality from `onClick` events on the overlay (e.g. Profile closing on click-outside). The loss in functionality is not severe - so I think we should log an issue to relook these changes when bp3 improves its implementation of the `Overlay` component.

PR #876, opened 2019-08-16
Show previous grader name/grading time in grading editor

alcen:
This commit is meant to make it easier for graders: by showing the previous grader's name and time graded in the interface, graders can avoid conflicts that arise when several graders work on the same set of submitted answers, and makes the grading editor reflect what students actually see in their grading results

gradername

This information is currently sent to the frontend, but it is not used in any way


alcen:
> I'll put this PR on hold. Ideally we should not use ! to indicate that a property is not null. I would also like to reduce the use of multiple falsey types i.e. undefined | null. I think we should stick to undefined, and only use null if absolutely necessary to differentiate between something that does not exist and something that does not need values

"null" is currently assigned by the backend though, so that will have to be changed in that case

Aulud:
@geshuming Instead of preventing the backend from building `:nil` (`null`) for the fields `grader` and `gradedAt` for an ungraded question, I have edited `getAssessment` in `requests.ts` to delete those fields if they are null instead. Is this a better alternative until `requests.ts` can be refactored to explicitly cast `resp` returned from the backend to a new object satisfying the frontend interfaces?

alcen:
> @geshuming Instead of preventing the backend from building :nil (null) for the fields grader and gradedAt for an ungraded question, I have edited getAssessment in requests.ts to delete those fields if they are null instead (hence the existence of these fields is indicated by presence of a corresponding value). Is this a better alternative until requests.ts can be refactored to explicitly cast resp returned from the backend to new objects satisfying the frontend interfaces?
> Have also updated getGrading in requests.ts to build these fields (in the new object returned as a GradingQuestion) iff they are non-null.

Thanks for the change. I think this workaround is a better option

PR #878, opened 2019-08-18
PE Infrastructure

jiachen247:
- add support for password protected assessment
- add practical tab

accompanying pr on the backend https://github.com/source-academy/cadet/pull/396
based on https://github.com/source-academy/cadet-frontend/pull/693

### DIFF
--- a/src/components/assessment/index.tsx
+++ b/src/components/assessment/index.tsx
@@ -298,7 +298,7 @@ class Assessment extends React.Component<IAssessmentProps, State> {
     switch (overview.status) {
       case AssessmentStatuses.not_attempted:
         icon = IconNames.PLAY;
-        label = 'Attempt';
+        label = overview.private ? 'Unlock' : 'Attempt';
         break;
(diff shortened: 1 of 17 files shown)


src/components/assessment/index.tsx
```diff
@@ -320,9 +320,15 @@ class Assessment extends React.Component<IAssessmentProps, State> {
...
-          overview.category
-        )}/${overview.id.toString()}/${DEFAULT_QUESTION_ID}`}
+        to={{
+          pathname: `/academy/${assessmentCategoryLink(
+            overview.category
+          )}/${overview.id.toString()}/${DEFAULT_QUESTION_ID}`,
+          search:
+            overview.private && overview.status === AssessmentStatuses.not_attempted
+              ? `isPasswordRequired=true`
+              : ''
```
geshuming: Using url query params is a really bad way of checking if an assessment requires a password.

I still think we should just send a null password, get the backend to check if the assessment requires a password, and return an error code if there is a need.

Issue #804, opened 2019-08-02
[Bug] Side content tab underline breaks on browser window resize


Aulud:
## [Bug] Side content tab underline breaks on browser window resize
UI-related bug.

### Bug description
The `SideContent` component was updated in PR #740 to make use of BlueprintJS 3's `Tabs` component. This includes an addition of an animated blue underline that moves between tabs when tabs are switched.

When the side-content width is modified by dragging the central divider, the position of the underline decouples from that of the selected tab. However, upon releasing the `<Resizable>` divider, the underline recomputes its position and snaps back under the currently selected tab.

Additionally, all of
   - resizing the browser window,
   - minimising it from a maximised state, and
   - opening the browser's developer console (F12)

also triggers this decoupling, afterwhich the underline fails to recompute its position, remaining frozen in an incorrect position until the side-content width is modified again or a tab is selected.

This bug seems to occur in all available workspaces (assessment, grading, playground, mission editor).

### Reproduction steps

1. Open the mock front-end (as of commit #796) in either Chrome or Firefox.
2. Proceed to the Playground.
3. Attempt to resize the central divider - this should demonstrate the decoupling of the animated underline.
4. Undock the browser window from a maximised state, then attempt to resize it by dragging its borders - this should demonstrate the frozen underline. Alternatively, press F12 to bring up the Developer console.

### Development environment
Browser: Google Chrome Version 75.0.3770.142 (Official Build) (64-bit)
Browser: Firefox Quantum 68.0.1 (64-bit)

Last updated 3 Aug 2019, 00:15AM

alcen:
Unfortunately this is a side effect of centering the tabs. The calculations for the position of the line come from the left margin

We could align the tabs to the left and it would solve this issue. Or we can disable the line altogether

martin-henz:
Hmm not sure. The bug is real, but minor. The underline is nice, and the centering looks good.

alcen:
> Hmm not sure. The bug is real, but minor. The underline is nice, and the centering looks good.

I think we could live with this for now? It is a core issue/feature request with Blueprint, that it does not allow tabs to be centered. Currently it is done by overriding their CSS styles with our own

martin-henz:
ok

martin-henz:
OK, let's live with it.

Aulud:
Closed as the issue is resolved by #879.

Aulud:
Closed as the issue is resolved by #879.

Aulud:
[...]

PR #879, opened 2019-08-18
Fix decoupling of active side-content tab and blue underline

Aulud:
## [Bugfix] Fix decoupling of active side-content tab and blue underline
Resolves issue #804.

### Bug description
Refer to issue #804 for full bug description.

### Changelog
- Applied CSS styling to some containers in bp3's `<Tabs>` to fix the decoupling between the active side-content tab and blue underline
   - Fixes both apply to resizing of the central divider and resizing of the browser window

### Demos
fixed-side-content-tab-blue-underline

Last updated 18 Aug 2019, 11:45PM

### DIFF
--- a/src/styles/_workspace.scss
+++ b/src/styles/_workspace.scss
@@ -318,8 +318,21 @@ $code-color-error: #ff4444;
     justify-content: center;
     display: flex;
 
-    .#{$ns}-tab-list {
-      justify-content: center;
+    /*
+      Fixes an innate issue with bp3's Tabs component where the active tab underline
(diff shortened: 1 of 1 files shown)


Aulud:
@geshuming This fix should not introduce any regression in any other part of the Academy as
   - bp3's `<Tabs>` is only used once in the Academy, in the `SideContent` React component that is used in `<Workspace>`, and
   - the CSS changes applied are scoped only to `<Tabs>` within the `<div>` with class `side-content-tabs`.

PR #882, opened 2019-08-19
Revert "Bump version to 1.3.3"

geshuming:
Reverts source-academy/cadet-frontend#881


Aulud:
Revert effected as GitHub editor removed latter ~10k LOC from `package-lock.json`.

PR #903, opened 2019-08-28
Inspector updates 

### LINKED ISSUE #843: Inspect displaying built in functions
Try pausing within this simple sorting algorithm. The inspector is exposed to built-in functions and somehow their implementations are visible as well.
```
function sort(M){
    function swap(i,j){
        const temp = M[i];
        M[i] = M[j];
        M[j] = temp;
    }
    
    let pLo = -1;
    let pHi = 15;
    let p = 0;
    while(p < pHi) {
        const curr = M[p];
        if (curr === 0){
            swap(p, pLo+1);
            pLo = pLo + 1;
            p = p + 1;
        }
        else if (curr === 2){
            swap(p, pHi-1);
            pHi = pHi - 1;
        }
        else {
            p = p + 1;
        }
    }
    
    return M;
}

sort([0,2,0,1,1,2,1,2,1,0,1,0,0,1,1]);
```

podocarp:
Updated and sorted all the builtins and we use the one in envvisualizer as a central source for this.

Fix #843


geshuming:
I'm getting weird visualizer images

image

geshuming:
@Aulud Can you test with all sorts of code? I'll check again what code I used to test it

geshuming:
This is the code snippet I used to test.

```javascript
let a = list(1,2,3,4,5);

for (let i = 0; i < 5; i = i + 1) {
    a = map(x=>x+1, a);
}

a;
```

This snippet seems to break it as well
```javascript
function fact(n) {
  return n === 0 ? 1 : n * fact(n - 1);
}
fact(5);
```

podocarp:
If you try the visualizer on the currently deployed version, it does not display. This is not a regression, we should just merge this small fix and open an issue for the visualizer.

podocarp:
The visualizer just tries to display too many things and clips. This is not a regression.

public/externalLibs/env_visualizer/visualizer.js
```diff
@@ -13,111 +15,10 @@
     height: 1000,
     container: container
   });
-  const frameFontSetting = '14px Roboto Mono, Courier New';
+  const frameFontSetting = '14px RobotGo Mono, Courier New';
```
geshuming: What is robotgo font?

Aulud: Hmm the `RobotGo Mono` doesn't exist. Unintended typo, @podocarp?

podocarp: That's odd. Don't recall trying to change that. I'll fix it.

martin-henz:
Let's focus on the INSPECTOR for this PR. From that point of view @geshuming do you think we can merge this? @TobiasWrigstad and I would like to test it. We can of course test it using the branch, but we would get a lot more eyeballs on it if the fixed version was in production.

The current version in production is barely usable.

.prettierrc
```diff
@@ -2,5 +2,4 @@
   "semi": true,
   "singleQuote": true,
   "printWidth": 100,
-  "parser": "typescript"
```
geshuming: Why is this removed? As far as I'm concerned, the prettier only looks at typescript files due to another config, and not this config.

podocarp: I wanted to prettify scss files with prettier. It's redundant either way. prettier knows what parser to use. I reverted it nevertheless.

PR #926, opened 2019-09-08
Update monospaced fonts used in Editor and REPL components

Aulud:
## Update monospaced fonts used in Editor components
Summary: Changes font settings for all code editor components and the REPL input.

### Changelog
- Remove non-monospaced fallback `Droid Sans` font from all editor components
   - This fixes a bug where some Linux distributions fail to load the monospaced fonts and fallback incorrectly to `Droid Sans` instead of the system-mapped `monospace` font
- Set the `Inconsolata` font used in to the SICP web edition as the default font for all editor components, REPL input, REPL output and substituter
   - `Consolas` then `monospace` are set as fallback fonts
   - Addresses https://github.com/source-academy/cadet-frontend/pull/767#issuecomment-515885192
- Increase font size for the above edited components to counteract the smaller point size of the `Inconsolata` font

### Mocks and Tests
- Update snapshots for the `Editor` component

#### Broken non-monospaced fallback font
broken-fallback-font

#### Before-after font comparison
TBD.

Last updated 8 Sept 2019, 11:45PM

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -14,7 +14,7 @@
     -->
   <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
   <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico?pr225" />
-  <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Droid+Sans|Droid+Sans+Mono|Droid+Serif" />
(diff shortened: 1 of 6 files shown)


Aulud:
Hi Prof @TobiasWrigstad, as an update, I've checked the web edition of SICP and it does indeed use `Inconsolata` as the default font for programs in the text. I have spoken to @jiayushe and we agreed that the previous style was in error.

pretty-print-code-blocks
pretty-print-code-piece

PR #930, opened 2019-09-11
Fix data visualizer resizing

lumos309:
Stops the visualizer from limiting the height of the display by creating a second, nested display area with its own set of scroll bars.

This solution is still not perfect, as if the drawing is too tall the user has to scroll to the bottom of the drawing to access the resizing bar and view the REPL.

### DIFF
--- a/src/styles/_playground.scss
+++ b/src/styles/_playground.scss
@@ -7,6 +7,4 @@
 
 .kineticjs-content {
   width: 100% !important;
-  // height: 40vh !important; // 100% results in 0 height, 100vh is too big.
-  // overflow: scroll;
 }


src/styles/_playground.scss
```diff
@@ -7,6 +7,6 @@
 
 .kineticjs-content {
   width: 100% !important;
-  height: 40vh !important; // 100% results in 0 height, 100vh is too big.
-  overflow: scroll;
+  // height: 40vh !important; // 100% results in 0 height, 100vh is too big.
+  // overflow: scroll;
```
geshuming: Don't comment out code unless there's really some need to? (e.g. there are plans to reimplement this)

lumos309: I'm not entirely sure whether those attributes will be needed again in the future, but I'll delete them for now.

PR #931, opened 2019-09-11
Clean up sound library

urieltan:
Fix for sound tone matrix quest


martin-henz:
> This just feels weird as a PR. I don't know what exactly this PR aims to fix. Can we just describe this as a clean-up?

The two most important things that it does are:
* establish the play_concurrently function used in Quest "Magical Tone Matrix"
* update get_matrix() function to new empty list: [] -> null

PR #940, opened 2019-09-19
Fix assessment fetch not prompting password

geshuming:
By setting the request option `shouldAutoLogout` to false, we should be able to obtain the 403 response and prompt user for password


jiachen247:
This if anything shoud be a temp can we log an issue to make sure we come back and fix this asap haha

PR #951, opened 2019-10-02
Swap sin and cos for unit circle

### LINKED ISSUE #914: 2020: CURVES: unit circle non-standard
The usual mathematical definition of the unit circle has it start on the right and go anticlockwise. The unit circle of the CURVES libary starts at the top and goes clockwise! The roles of sine and cosine need to be flipped.

JingYenLoh:
Fixes #914


martin-henz:
This needs careful review of the missions. I'm adding an Issue in the missions repo.

PR #958, opened 2019-10-10
[ENV] Fix various bugs, improve arrow layout

lumos309:
* Fixed anonymous functions not pointing back to program environment when appropriate
* Fixed arrows for multiple names in same frame pointing to same object not appearing
* Reduce overlapping object arrows

Todo: 
Refine implementations for points 2 and 3


martin-henz:
This was a critical bug. Ive checked the offending example.

More testing needed later.

PR #974, opened 2019-10-25
Bump js-slang

openorclose:
https://github.com/source-academy/js-slang/pull/362

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10202,9 +10202,9 @@
       "integrity": "sha512-hm2nYpDrwoO/OzBhdcqs/XGT6XjSuSSCVEpia+Kl2J6x4CYt5hISlVL/AYU1khoDXv0AQVgxtdJySb9gjAn56Q=="
     },
     "js-slang": {
-      "version": "0.3.4",
-      "resolved": "https://registry.npmjs.org/js-slang/-/js-slang-0.3.4.tgz",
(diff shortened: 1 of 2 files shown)


martin-henz:
Is merge deployed already? 

I'm asking because I might want to squeeze in a change in the parse function of Source §4, if you could wait a while.

openorclose:
> Is merge deployed already?
> 
> I'm asking because I might want to squeeze in a change in the parse function of Source §4, if you could wait a while.

Yeah I don't think there's a deploy yet

martin-henz:
OK. Want to check with you: I don't like the verbosity of the output of the parse function. It has all this debugging info: line and position numbers, which distract the students and which add clutter. In the course we don't use this info. It's of course needed for any serious work, but I think that would go beyond the purpose of this module. So I'm thinking of cutting out all that stuff from the parser. Any views?

openorclose:
> OK. Want to check with you: I don't like the verbosity of the output of the parse function. It has all this debugging info: line and position numbers, which distract the students and which add clutter. In the course we don't use this info. It's of course needed for any serious work, but I think that would go beyond the purpose of this module. So I'm thinking of cutting out all that stuff from the parser. Any views?

Yeah that sounds good, we can have `parse` that doesn't have the line and location, and `parse_verbose` that works like how the current `parse` works.

martin-henz:
That's a good idea! Can you set that up? Would save me a lot of time!

PR #980, opened 2019-11-03
Set default source chapter to 4

chrisgzf:


martin-henz:
No harm in this. Best set to Source §4 for the rest of the semester. I'll propose a different approach in an Issue for next year.

PR #982, opened 2019-11-04
Mission editor fix globals not loading

blackening:
This tries to parse the globals when the mission is loaded into the mission editor.

If there are any failures, they will be shown inside the browser console.


seanlowjk:
currently i'm not sure how to test it !

Are there any situations where this error occur? Thanks ><

PR #1000, opened 2020-02-06
Create a vanilla source mode based on brace/mode/javascript and enable autocompletion

BobbyZhouZijian:
1. Import javascript mode from "src/components/workspace/editorMode/source-mode.js instead of "brace/mode/javascript."

2. Enable autocompletion inside Editor.tsx at setOptions


martin-henz:
You call the file:
workspace/editorMode/source-mode.js
I think in Ace, the mode files just carry the name of the language, not "-mode", so I think we should rename it to 
workspace/editorMode/source.js

PR #1004, opened 2020-02-20
fix issues with colouring and add keywords of all source chapters

BobbyZhouZijian:
1. change the keys of KeyWordMapper back to string type.
2. change the imported highlight rules for SourceMode to SourceHighlightRules
3. add function names for source chapter1-4

### DIFF
--- a/src/components/workspace/editorMode/source.js
+++ b/src/components/workspace/editorMode/source.js
@@ -63,148 +63,63 @@ function _DocCommentHighlightRules(acequire, exports, module) {
 function _SourceHighlightRules(acequire, exports, module) {
     "use strict";
     
-    //parent classes to be inherited
     const oop = acequire("../lib/oop");
     const DocCommentHighlightRules = acequire("./doc_comment_highlight_rules").DocCommentHighlightRules;
     const TextHighlightRules = acequire("./text_highlight_rules").TextHighlightRules;
-    
-    // TODO: Unicode escape sequences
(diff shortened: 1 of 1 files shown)


src/components/workspace/editorMode/source.js
```diff
@@ -63,63 +63,99 @@ function _DocCommentHighlightRules(acequire, exports, module) {
 function _SourceHighlightRules(acequire, exports, module) {
     "use strict";
     
+    //parent classes to be inherited
     const oop = acequire("../lib/oop");
     const DocCommentHighlightRules = acequire("./doc_comment_highlight_rules").DocCommentHighlightRules;
     const TextHighlightRules = acequire("./text_highlight_rules").TextHighlightRules;
-    const identifierRegex = "[a-zA-Z\\$_\u00a1-\uffff]-[a-zA-Z\\d\\$_\u00a1-\uffff]*";
-
+
```
martin-henz: Avoid repetition of names. The Source languages form a strict hierarchy, so a predeclared name in Source §x is also a predeclared name in Source §(x+i), i > 0. 

You should exploit this by concatenating the names:
const chapter2 = chapter1 + "...";
etc

BobbyZhouZijian: i see. i will make changes to it

BobbyZhouZijian:
move the source mode and theme to  js-slang

PR #1022, opened 2020-03-19
Add support for Source 3 Concurrent

jonchan51:
Adds support for Source 3 Concurrent on cadet-frontend. 
Features that are not supported have been hidden from view when Source 3.4 is selected.
This includes the following:
- Eval Button
- REPL input
- Inspector
- Environment Visualizer

Breakpoint insertion has not been removed, as we might be adding support for them in the future.

Deployed version available [here](https://jonchan51.github.io/playground)


src/components/Playground.tsx
```diff
@@ -232,6 +240,21 @@ class Playground extends React.Component<IPlaygroundProps, PlaygroundState> {
...
     );
 
+    const playgroundIntroductionTab: SideContentTab = {
+      label: 'Introduction',
+      iconName: IconNames.COMPASS,
+      body: (
+        <Markdown
+          content={
+            INTRODUCTION +
+            (this.props.sourceVariant === 'concurrent' ? CONCURRENT_SOURCE_INTRODUCTION : '')
```
martin-henz, line 250: Nice to see a scheme how we can evolve these intro sections and make them specific to the particular Source language.

PR #1027, opened 2020-03-24
update testcases due to blueprint upgrade

martin-henz:


martin-henz:
Tried to upgrade blueprint, and re-generated the testcases.

PR #1051, opened 2020-04-07
Autocomplete feature

tuesmiddt:
- Enable autocompletion prompt
- Prompt names in scope including built-ins and external libraries
- Display documentation where possible

TODO:
- Disable prompt when declaring function parameters.

### DIFF
--- a/src/sagas/workspaces.ts
+++ b/src/sagas/workspaces.ts
@@ -107,16 +107,35 @@ export default function* workspaceSaga(): SagaIterator {
       (state: IState) => (state.workspaces[workspaceLocation] as IWorkspaceState).context
     );
 
-    const code: string = yield select(
-      (state: IState) => (state.workspaces[workspaceLocation] as IWorkspaceState).editorValue!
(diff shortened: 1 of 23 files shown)


src/sagas/workspaces.ts
```diff
@@ -96,6 +98,61 @@ export default function* workspaceSaga(): SagaIterator {
...
+    );
+
+    const editorSuggestions = editorNames.map((name: any) => ({
+      caption: name.name,
+      value: name.name,
+      meta: name.meta,
+      score: 1000 // Prioritize suggestions from code
+    }));
+
+    const builtinSuggestions = Documentation.builtins[context.chapter] || [];
```
jonchan51, line 141: Hi, was looking at the frontend and realised that the variants aren't supported yet.
image
I took a look and suggest something along the lines of the above. Not sure if there are any consequences of doing that right now though, and whether other variants are able to do something like this.
The inserted code in the screenshot works for Source 3 Concurrent at least.

PR #1065, opened 2020-04-13
Encode chapter and variant separately in Share URL

anubh-v:
Currently, when we create the Share URL for the Non-deterministic and Concurrent variants, we encode the chapter and variant information together in one string.

For example, if the playground was using `Source 3 Non-Det`, then then Share URL will look
like this `https://source-academy.github.io/playground#chap=3_Non_Det&exec=1000&ext=NONE&prgrm=IYWwRgFAjAlA3EA`

Note the string `#chap=3_Non_Det` in the URL.

Instead of this, we can encode the chapter and variant separately.
So, the url will look like: `https://source-academy.github.io/playground#chap=3&exec=1000&ext=NONE&prgrm=IYWwRgFAjAlA3EA&variant=non-det`

This PR reduces the steps needed to add a new language variant.
The Share URL will automatically support any language variants added into the `sourceLangauges` array in `src/reducers/states.ts`.

**Note: The `variant` field in the Share URL is optional**
If the URL does not have a `variant`, the playground will use the `default` variant.
This ensures the Share URL is backward compatible

### DIFF
--- a/src/components/Application.tsx
+++ b/src/components/Application.tsx
@@ -9,7 +9,7 @@ import Material from '../containers/material/MaterialContainer';
 import MissionControlContainer from '../containers/missionControl';
 import Playground from '../containers/PlaygroundContainer';
 import Sourcecast from '../containers/sourcecast/SourcecastContainer';
-import { languageURLNames, Role } from '../reducers/states';
+import { Role, sourceLanguages } from '../reducers/states';
 import { stringParamToInt } from '../utils/paramParseHelpers';
 import { ExternalLibraryName, ExternalLibraryNames } from './assessment/assessmentShape';
 import Contributors from './contributors';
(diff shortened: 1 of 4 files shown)


anubh-v:
If the URL has a chapter - variant combination that is not supported, then the variant will be set to `default`.

One example is if the URL has chapter = 4, but variant = `non-det`

martin-henz:
> If the URL has a chapter - variant combination that is not supported, then the variant will be set to `default`.
> 
> One example is if the URL has chapter = 4, but variant = `non-det`

Not sure if that's the right choice. I see four options:
(1) change variant to "default"
(2) increase chapter until a matching variant exists
(3) decrease chapter until a matching variant exists
(4) reject program

I think in any case (except perhaps 2) we should give a warning. 

A clear and simple solution would be simply (4) with an error message.

If you want to be fancy, you could:

* Follow (2), if not possible:
* Follow (3) and give warning, if not possible:
* Follow (1) with a warning, if not possible:
* Follow (4) with an error message.

PR #1070, opened 2020-04-14
fix minor look-up bug

BobbyZhouZijian:
see https://github.com/source-academy/js-slang/issues/543


BobbyZhouZijian:
hmmmm I didnt really write a test for the frontend part. I can try if I can come up with one for the frontend

PR #1086, opened 2020-04-22
Add boolean flag to activate wasm

### LINKED ISSUE #1091: Stepper: should be disabled in Source 1 wasm
Currently, the stepper there creates the impression that it's implemented using wasm, but it isn't.

Screenshot 2020-04-25 at 12 52 30 PM

J0:
Resolves #1091.


martin-henz:
Changed to WIP because results are not shown in REPL.

btzy:
WebAssembly will work on GitHub Pages and most decent servers.  The webpack-dev-server that is in cadet-frontend doesn't support it currently.  But if you do `npm run build` and use the production version, it will work.

PR #1087, opened 2020-04-24
support ext display and enable const look-ups

BobbyZhouZijian:
Add support to display highlights for different external libraries. Also enable lookups for constants.


martin-henz:
Looks good!

PR #1088, opened 2020-04-24
More precise navigation in library documentation lookup

martin-henz:


BobbyZhouZijian:
Tested and works. Didn't really notice that we have separate libraries for different external libs.

PR #1103, opened 2020-05-01
Fix stepper list/pair bug + update ace mode import

zhaojj2209:
Fixed bug that causes frontend to crash due to lists/pairs being wrongly plugged into the SubstVisualizer component.

Updated ace mode import for Stepper display.

Edit: SubstVisualizer now takes in IStepperPropContents[] instead of Array<[string, string, string]>, Playground now checks for IStepperPropContents using isStepperOutput function before plugging in output into SubstVisualizer (waiting for js-slang bump)


src/components/Playground.tsx
```diff
@@ -418,7 +418,13 @@ class Playground extends React.Component<IPlaygroundProps, PlaygroundState> {
 
   private processArrayOutput = (output: InterpreterOutput[]) => {
     const editorOutput = output[0];
-    if (editorOutput && editorOutput.type === 'result' && editorOutput.value instanceof Array) {
+    if (
+      editorOutput &&
+      editorOutput.type === 'result' &&
+      editorOutput.value instanceof Array &&
+      editorOutput.value[0] instanceof Array &&
+      typeof editorOutput.value[0][0] === 'string'
```
martin-henz, line 426: This is not obvious. Can you add a comment what problem this fixes and how it fixes it?

src/components/workspace/side-content/SubstVisualizer.tsx
```diff
@@ -143,8 +143,9 @@ class SubstVisualizer extends React.Component<ISubstVisualizerProps, ISubstVisua
     const lastStepValue = this.props.content.length;
     const contIndex = value <= lastStepValue ? value - 1 : 0;
     const pathified = this.props.content[contIndex];
-    const redexed = pathified[0];
-    const redex = pathified[1].split('\n');
+    const redexed = pathified[0] + '';
+    const unsplitRedex = pathified[1] + '';
```
martin-henz, line 147: Same here: a comment would be helpful

src/components/Playground.tsx
zhaojj2209: Since the stepper input is of type Array<[string, string, string]>, the above checks the editor output value for the following:
1. Is an array
2. First element is an array
3. First element of that array is a string

Will be adding a commit that includes this comment in the code!

PR #1113, opened 2020-05-11
Type hints

openorclose:
Show type errors when error occurs


martin-henz:
One small thing before we merge: Can we handle the case where the error is empty, and avoid the display of the string "Hints:"?

Screenshot 2020-05-14 at 11 48 30 AM

PR #1122, opened 2020-05-17
Fix #1005

### LINKED ISSUE #1005: INSPECT: type error displayed in intro dialog
Screenshot 2020-02-25 at 12 06 27 AM

podocarp:
For some reason the inspector is now trying to draw even before the program is evaluated. The error is now gone, and future errors will be logged in console instead of printed on the inspector.


podocarp:
Did not realise default went over to sa_2021, I'll simply cherry pick to another branch

PR #1130, opened 2020-05-20
Add URL Shortening Microservice

### LINKED ISSUE #1003: Share: URLs are very long
A suggestion was to make use of a free URL-shortening API to shorten these URLs.

rrtheonlyone:
Fixes #1003 

## Introduction

Adds an in-house micro-service to shorten long urls from the playground. The URL shortener is [YOURLs](https://yourls.org/). Currently, this is deployed to a custom AWS instance for testing purposes.

The playground share button now looks like this:
Screenshot 2020-05-20 at 8 26 06 AM

GIF of how it works with the meta-circular evaluator programme:
test

You may notice that the domain of the shortener is currently http://ec2-18-140-69-106.ap-southeast-1.compute.amazonaws.com (this is my own AWS server). This should be swapped out for a custom domain from NUS! 

## How it works 

Nothing fancy. The YOURLs php service is deployed on a custom AWS instance. The frontend makes a API call to the service to retrieve a short url. The urls are stored in MariaDB MySQL database.

There is also an admin dashboard available to show the urls in the database.

Screenshot 2020-05-20 at 8 55 05 AM

## Changelog

Small frontend changes namely:
- playground component and container updated
- sharebutton component updated
- new saga added

Remaining TODOs:
- [x] improve error handling
- [x] update authentication 
- [x] basic tests
- [x] discuss on deployment of actual AWS instance with domain


rrtheonlyone:
Requesting preliminary UI / feature review from @martin-henz and @blackening

martin-henz:
Looks good. There is a complication though. We have been playing with the idea of multiple deployments of the Source Academy, for different purposes. Currently we have:
* https://sourceacademy.nus.edu.sg/ for CS1101S, and
* https://source-academy.github.io/playground for the general public, including readers of SICP.

With this feature, we would establish a de-facto default playground: the URL used by the share feature.

If we want to be a bit more fancy, we could let each deployment generate its own URLs. So if you share a program in
https://source-academy.github.io/, the URL would go to
https://source-academy.github.io/...
and if you share a program in 
https://sourceacademy.nus.edu.sg/ , the URL would go to
https://sourceacademy.nus.edu.sg/...

I think for this, it would be acceptable, if the short names are globally distinct so that we can still use a single URL shortening service.

rrtheonlyone:
Yup we discussed on this and reached a consensus!

This share service will behave similar to tinyurl/bit.ly. So regardless of where you call it from (staging/prod), it will give the same domain. We are looking to make it `share.sourceacademy.nus.edu.sg/<alias>`

martin-henz:
Yes, we have a consensus on this: same short urls, and the actual long links take the user back to deployment where the link was generated.

martin-henz:
A small issue that I raised: The color of the popup in case the link exists already should be green and not orange.

rrtheonlyone:
@martin-henz yup this has been fixed already, try it out at https://source-gpu.herokuapp.com/playground

PR #1144, opened 2020-05-23
Allow specifying credentials for development

jiayushe:
- Allow specifying credentials as env vars for development
- Make it easier for development purpose


angelsl:
Eh, I thought of adding a button on the login screen that just asks for the tokens instead. Easier than specifying it in the `.env` IMO.

But currently you can just manually dispatch a token update to the Redux store.

jiayushe:
@angelsl 
> Eh, I thought of adding a button on the login screen that just asks for the tokens instead. Easier than specifying it in the `.env` IMO.

But we don't really need that in production? It will complicate the interface I think.

> But currently you can just manually dispatch a token update to the Redux store.

Yea I know but you need to do that everytime you start the dev server. It is easier and cleaner to just keep that as env vars.

angelsl:
> But we don't really need that in production? It will complicate the interface I think.

It wouldn't be shown in production, of course.

> Yea I know but you need to do that everytime you start the dev server. It is easier and cleaner to just keep that as env vars.

You don't need to. Just make a token that lasts long enough and dispatch it into the Redux store; it'll be persisted in the browser.

The thing is sometimes you want to change token to test something. If it's an environment variable, you'd have to restart the dev server, which takes super long.

angelsl:
It'd probably be even easier if you had a frontend button to set the tokens.

Let me come up with something since I'm in the midst of overhauling the authentication anyway.

angelsl:
Closing this; the new authentication changes make this unnecessary.

PR #1157, opened 2020-05-25
Fix autorun

### LINKED ISSUE #1059: Playground: Auto-run not working
The auto-run button on the top left is not working.

When auto-run is activated, the playground is supposed to run the program in the editor whenever it passes the parser.

### LINKED ISSUE #1159: Code editor crashes attempting to de-reference null
## [Bug] Code editor crashes attempting to de-reference null
Typing error.

### Bug description
The variable highlighting feature of the code editor relies upon asynchronous calls to `setTimeout` to perform variable highlighting, as seen here.

https://github.com/source-academy/cadet-frontend/blob/33c7704a279910fb26f01586776480f345fde65e/src/components/workspace/Editor.tsx#L406-L416

When the user navigates between pages containing code editors (e.g. when viewing questions in an assessment), the value of `(this.AceEditor.current as any)` is momentarily `null`. Due to the asynchronous nature of `setTimeout`, this result in the following error when the asynchronous call attempts to access the `editor` property of `(this.AceEditor.current as any)`.

image

The implementation of new features in the editor is badly done - removing the `any` typecast shows that the type of `this.AceEditor.current` is nullable, i.e. `AceEditor | null`.

image

Suppressing compiler warnings for nullable types by typecasting the offending properties to type `any`, so as to prevent the compiler from performing type inference, is **extremely bad practice**. This also includes ignoring the potential for `null` values by using `this.AceEditor.current!`, unless it has been checked as a pre-condition.

### Replication instructions
1. Configure the SA front-end to use a back-end, by setting `USE_BACKEND` to `TRUE` in `.env`, then set up the back-end
2. Open any assessment (Mission, Quest, etc.)
3. Click the editor to focus it
4. Then navigate to the next question and wait a few seconds - **the error occurs here**

Last updated 26 May 2020, 3:15AM

jiayushe:
- Fix #1059 
- Remove `as any` and fix #1159

### DIFF
--- a/src/components/sourcecast/SourcecastEditor.tsx
+++ b/src/components/sourcecast/SourcecastEditor.tsx
@@ -1,6 +1,6 @@
 import { isEqual } from 'lodash';
 import * as React from 'react';
-import AceEditor, { IAnnotation } from 'react-ace';
+import AceEditor from 'react-ace';
 import { HotKeys } from 'react-hotkeys';
 
 import 'ace-builds/src-noconflict/ext-searchbox';
@@ -45,7 +45,6 @@ class SourcecastEditor extends React.PureComponent<ISourcecastEditorProps, {}> {
   public ShareAce: any;
   public AceEditor: React.RefObject<AceEditor>;
   private onChangeMethod: (newCode: string, delta: ICodeDelta) => void;
-  private onValidateMethod: (annotations: IAnnotation[]) => void;
   private onCursorChange: (selecction: any) => void;
   private onSelectionChange: (selection: any) => void;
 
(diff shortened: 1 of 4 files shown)


angelsl:
There's not much difference between `as any` and `!`; it's still gonna blow up if `AceEditor` is in fact null or undefined.

See #1159

jiayushe:
> There's not much difference between `as any` and `!`; it's still gonna blow up if `AceEditor` is in fact null or undefined.
> 
> See #1159

Yea I know. This pr is not a fix for that. I just saw that everyone was copy-pasting `as any` blindly (and possibly have no idea what they are doing) and decided to change it to `!` so next time they will know what they are doing and handle it properly.

PR #1169, opened 2020-05-27
Story Simulator component (WIP)

sciffany:
This component allows you to load XMLs and play them in your browser.

image


martin-henz:
A small suggestion: The term "upload" I find a bit misleading here, because the files remain on the client. How about we say "load" instead of "upload"?

Issue #1147, opened 2020-05-24
UI: Modal Blueprint components do not close correctly


Aulud:
## [Bug] Modal Blueprint components do not close correctly
UI-related bug.

### Bug description
The front-end makes extensive use of interactive Blueprint modal components, e.g. [Dialog](https://blueprintjs.com/docs/#core/components/dialog), [Drawer](https://blueprintjs.com/docs/#core/components/drawer). By default, these components are set to close the modal overlay when the user:
- presses the Close button in the component's header 
- presses the Escape key
- **clicks outside of the overlay element**

However, due to hotfixes introduced in #863 and #874, the above bolded user action no longer correctly closes the modal components. This prevents users from closing modal components that do not explicitly provide button(s) to close the component.

Last updated 24 May 2020, 9:15PM

PR #1183, opened 2020-05-28
Fix modal Blueprint components not closing correctly

Aulud:
## UI/UX: Fix modal Blueprint components not closing correctly
Summary: Fixes issue #1147 introduced by PR #874, by introducing a better fix for the original issue addressed in PR #863 that led to PR #874. **This PR breaks compatibility with Internet Explorer 11**.

### Description
Modal Blueprint components such as `Dialog` or `Alert` are used extensively across the Academy to render content atop the base application, such as for user prompts (e.g. confirmation prompts) and notices (e.g. assessment briefings, warning messages). These components rely upon the low-level Blueprint `Overlay` component to be mounted and rendered atop the base page's DOM tree.

Specifically, the assessment briefing displayed to a student when they first view an assessment uses the `Dialog` component. Typically, the length of the assessment briefing is sufficient for the `Dialog` component to exceed the height of the browser window, which causes a vertical scrollbar to be rendered (see below example).

image

This vertical scrollbar renders at the level of the parent `Overlay` component `<div>`, which lies beneath both the overlay backdrop `<div>` and the full-width `Dialog` component `<div>`. As a result, this vertical scrollbar is blocked by the `Dialog` component `<div>` and can not be used to scroll the dialog. This is in fact a bug in Blueprint itself, tracked by issue https://github.com/palantir/blueprint/issues/1008. Whilst a reliable fix for the bug exists, it would break IE11 compatibility and thus will not be implemented by Blueprint.

PR #863 thus applied the hotfix suggested by https://github.com/palantir/blueprint/issues/1008#issuecomment-403207576 from the Blueprint issue, which disables the overlay backdrop (responsible for the gray tint over the rest of the Academy) and applies the gray tint directly to the `Overlay` component `<div>` itself. However, this has two undesirable side-effects:
- it causes all other overlay components to render with the gray background tint, for which another hotfix was introduced in #874
- it prevents the user from closing any modal components by clicking outside of the overlay element, as the overlay backdrop cannot catch the `click` mouse event to close the modal component

This hotfix, whilst sufficient during the teaching semester, is not a reliable long-term solution - even necessitating the PR #1120 to address further issues with `Dialog` prompts. I propose an alternative fix instead, which makes use of the [`sticky`](https://developer.mozilla.org/en-US/docs/Web/CSS/position) value for the CSS `position` property, **which is not supported by IE11**. This should be fine, since the Academy officially only supports the Firefox and Chrome browsers.

The fix is as follows:
- the overlay backdrop is tethered to the top of the page (viewport) with sticky positioning, and its width and height artificially set to 100% to span the full height and width of the page
  - this ensures the overlay backdrop continues to apply the gray tint over the Academy whilst the user scrolls the dialog
  - this allows the overlay backdrop to detect when it is clicked and close its associated modal component
- the full-width `Dialog` `<div>` is absolutely positioned with an offset of 0 from the top of the page (since no ancestor element of it is positioned)
  - the parent `Overlay` `<div>` continues to render the scrollbar due to the height of the content in the dialog
  - the full-width Dialog `<div>` however is removed from the document flow, preventing it from blocking the scrollbar

#### Potential for Regressions
The main concern with this fix is the possibility of introducing regressions in any Blueprint components which indirectly use an `Overlay`. The ones in use in the Academy are:
- `Alert` - unsubmission confirmation prompt in Grading page
- `Dialog` - assessment briefings and most user prompts, e.g editor reset prompt
- `Drawer` - Profile component
- `Popover` - notifications badge or the drop-down in the header bar
- `Toast` - for showing warning or error messages, e.g. to provide feedback to students in Paths
- `Tooltip` - hover-based prompts, e.g. grade and XP breakdown tooltips in Grading page, info tooltip for Run button in Playground

**I have verified that this fix resolves all the issues raised in #863, #874, #1120 and #1147 and does not result in unintended changes to the behaviour of these overlay components used elsewhere in the Academy.**

### Changelog
- Remove unnecessary props from the Blueprint `Drawer` component used in the Profile component
- Reverts the hotfixes applied in PRs #863, #874
  - This fixes modal Blueprint components not closing when the user clicks outside of the overlay element
- Applies a new CSS fix to the `Overlay` and `Dialog` components to allow the user to scroll long dialogs with the browser scrollbar and the mouse scrollwheel
  - This preserves the user's ability to scroll the dialog with the arrow keys or Page Up/Page Down keys
- Update mock assessment overviews, assessments and notifications and corresponding snapshots
  - A mock Mission *Symphony of the Winds* was added to use a long mission briefing to demonstrate the fix
- Fix the assessment betcha and editor reset dialogs not closing when the close button in the dialog's header is clicked due to a missing `onClose` prop

### Verification instructions
Please test on both **Firefox** (v70 or later) and **Chrome** (v75 or later):
1. Configure the SA front-end to use the mock backend by setting `USE_BACKEND` to `FALSE` in `.env`
2. Login to SA and navigate to *Missions*, then select the mock mission *Symphony of the Winds*
3. Attempt to scroll the mission briefing dialog in multiple ways - all of these should succeed:
    - Using the browser scrollbar
    - Using the mouse scrollwheel, with the mouse focused on or off the dialog
    - Using the arrow or Page Up/Page Down keys, with the mouse focused on or off the dialog
    - Using swiping touch gestures if using a tablet or touchscreen device
4. Now verify that the two Blueprint components `Dialog` and `Drawer` correctly close by default when the user clicks outside of the overlay element - suggested ones:
    - `Dialog`: *About* or *Help* dialogs from the dropdown menu, assessment betcha or editor reset dialogs
    - `Drawer`: *Profile* component from the dropdown menu

Additionally, you may want to manually verify that no regressions have occurred in any of the overlay Blueprint components used in the Academy - please refer to the list above.

Last updated 30 May 2020, 5:30AM

### DIFF
--- a/src/components/assessment/AssessmentWorkspace.tsx
+++ b/src/components/assessment/AssessmentWorkspace.tsx
@@ -175,6 +175,7 @@ class AssessmentWorkspace extends React.Component<
         />
       );
     }
+
     const overlay = (
       <Dialog className="assessment-briefing" isOpen={this.state.showOverlay}>
         <Card>
@@ -189,12 +190,14 @@ class AssessmentWorkspace extends React.Component<
       </Dialog>
     );
 
+    const closeOverlay = () => this.setState({ showResetTemplateOverlay: false });
     const resetTemplateOverlay = (
       <Dialog
         className="assessment-reset"
         icon={IconNames.ERROR}
         isCloseButtonShown={true}
         isOpen={this.state.showResetTemplateOverlay}
+        onClose={closeOverlay}
         title="Confirmation: Reset editor?"
       >
         <div className={Classes.DIALOG_BODY}>
@@ -203,19 +206,14 @@ class AssessmentWorkspace extends React.Component<
         </div>
         <div className={Classes.DIALOG_FOOTER}>
           <ButtonGroup>
-            {controlButton(
-              'Cancel',
-              null,
-              () => this.setState({ showResetTemplateOverlay: false }),
-              {
-                minimal: false
-              }
-            )}
+            {controlButton('Cancel', null, closeOverlay, {
+              minimal: false
+            })}
             {controlButton(
               'Confirm',
               null,
               () => {
-                this.setState({ showResetTemplateOverlay: false });
+                closeOverlay();
                 this.props.handleEditorValueChange(
                   (this.props.assessment!.questions[questionId] as IProgrammingQuestion)
                     .solutionTemplate
@@ -228,6 +226,7 @@ class AssessmentWorkspace extends React.Component<
         </div>
       </Dialog>
     );
+
     /* If questionId is out of bounds, set it to the max. */
     const questionId =
       this.props.questionId >= this.props.assessment.questions.length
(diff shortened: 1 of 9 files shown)


seanlowjk:
Kudos!

PR #1194, opened 2020-06-01
Add Clarity for Assessment Source Version 

seanlowjk:
### Description

This PR Fixes the issue of Source Chapter not being displayed for various Assessments in Source Academy. .

The main Implementation behind this is to add a simple `isClickable` flag in order to ensure that students **are not** allowed to change the Source Chapter of Assessments. 

Fixes #953

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### How to Test

1. Login to Source Academy as any user. 
2. Go to an Assessment 
3. Ensure the Source Chapter matches with the `ChapterSelect` button 
4. `ChapterSelect` **should not** be clickable. 

### Checklist

Please delete options that are not relevant.

- [X] I have tested this code
- [ ] I have updated the documentation

PR #1199, opened 2020-06-01
Implement Source Version indication for Assessments

### LINKED ISSUE #953: Reflect source level in assessments
Currently theres really no way for students and even avengers to know which source level is being used for the different assessments.

seanlowjk:
### Description

This PR Fixes the issue of Source Chapter not being displayed for various Assessments in Source Academy. .

The main Implementation behind this is to add a simple `isClickable` flag in order to ensure that students **are not** allowed to change the Source Chapter of Assessments. 

This is a fix from fixing the commit history from #1194

Fixes #953

### Type of change

### How to Test

1. Login to Source Academy as any user. 
2. Go to an Assessment 
3. Ensure the Source Chapter matches with the `ChapterSelect` button 
4. `ChapterSelect` **should not** be clickable. 

### Checklist

- [X] I have tested this code

### DIFF
--- a/src/components/workspace/controlBar/chapterSelect.tsx
+++ b/src/components/workspace/controlBar/chapterSelect.tsx
@@ -10,7 +10,7 @@ export type ChapterSelectProps = {
   handleChapterSelect?: (i: IChapter, e: React.ChangeEvent<HTMLSelectElement>) => void;
   sourceChapter: number;
   sourceVariant: Variant;
-  isClickable: boolean;
+  disabled: boolean;
   key: string;
 };
 
@@ -39,25 +39,14 @@ export function ChapterSelect(props: ChapterSelectProps) {
   );
   const ChapterSelectComponent = Select.ofType<IChapter>();
 
-  const chapterSelector = (currentChap: number, currentVariant: Variant, isClickable: boolean) => {
-    const stopPopup = (e: any) => {
-      e.stopPropagation();
-    };
-
-    return isClickable ? (
+  const chapterSelector = (currentChap: number, currentVariant: Variant, disabled: boolean) => {
+    return (
       <div>
         <Button
           className={Classes.MINIMAL}
           text={styliseChapter(currentChap, currentVariant)}
-          rightIcon={IconNames.DOUBLE_CARET_VERTICAL}
-        />
-      </div>
-    ) : (
-      <div onClick={stopPopup}>
-        <Button
-          className={Classes.MINIMAL}
-          text={styliseChapter(currentChap, currentVariant)}
-          rightIcon={null}
+          rightIcon={disabled ? null : IconNames.DOUBLE_CARET_VERTICAL}
+          disabled={disabled}
         />
       </div>
     );
@@ -67,7 +56,7 @@ export function ChapterSelect(props: ChapterSelectProps) {
     currentChap: number,
     currentVariant: Variant,
     handleSelect = (i: IChapter, e: React.ChangeEvent<HTMLSelectElement>) => {},
-    isClickable: boolean
+    disabled: boolean
   ) => (
     <ChapterSelectComponent
       className={Classes.MINIMAL}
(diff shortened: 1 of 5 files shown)


src/components/workspace/controlBar/chapterSelect.tsx
```diff
@@ -38,10 +39,35 @@ export function ChapterSelect(props: ChapterSelectProps) {
   );
   const ChapterSelectComponent = Select.ofType<IChapter>();
 
+  const chapterSelector = (currentChap: number, currentVariant: Variant, isClickable: boolean) => {
+    const stopPopup = (e: any) => {
```
Aulud, line 43: This seems like a very primitive approach to intercepting all mouse events, and the typing here of the event as `any` opens the possibility of invoking `stopPropagation` on a non-event.

The Blueprint [`Button`](https://blueprintjs.com/docs/#select/select-component.javascript-api) and [`Select`](https://blueprintjs.com/docs/#select/select-component.javascript-api) components allow you to specify a `disabled` prop for if that component should respond to mouse events. Using both of these props (needed since otherwise the event will bubble) allows us to get rid of the wrapping `<div>`.

The final result looks like this, and attempting to hover over it shows the disabled mouse pointer. From a UX perspective, this is probably better as it doesn't give the impression to students they can click on the chapter select button.

image

src/components/workspace/controlBar/chapterSelect.tsx
```diff
@@ -38,10 +39,35 @@ export function ChapterSelect(props: ChapterSelectProps) {
...
+      e.stopPropagation();
+    };
+
+    return isClickable ? (
+      <div>
+        <Button
+          className={Classes.MINIMAL}
+          text={styliseChapter(currentChap, currentVariant)}
+          rightIcon={IconNames.DOUBLE_CARET_VERTICAL}
```
Aulud, line 52: We can inline the value of `rightIcon` with a ternary statement on the value of `isClickable`.

seanlowjk: Aights!

seanlowjk:
Yea after looking at it, it is right that we can just use the `disabled` flag for both the `Select` and `Button` Components which make up the `ChapterSelect` Component. 

Thanks for the reviews! @Aulud @angelsl

src/components/workspace/controlBar/chapterSelect.tsx
```diff
@@ -38,25 +39,41 @@ export function ChapterSelect(props: ChapterSelectProps) {
   );
   const ChapterSelectComponent = Select.ofType<IChapter>();
 
+  const chapterSelector = (currentChap: number, currentVariant: Variant, disabled: boolean) => {
```
angelsl, line 42: Maybe make it `disabled?` here, and...

src/components/workspace/controlBar/chapterSelect.tsx
```diff
@@ -38,25 +39,41 @@ export function ChapterSelect(props: ChapterSelectProps) {
...
+    return (
+      <div>
+        <Button
+          className={Classes.MINIMAL}
+          text={styliseChapter(currentChap, currentVariant)}
+          rightIcon={disabled ? null : IconNames.DOUBLE_CARET_VERTICAL}
+          disabled={disabled}
```
angelsl, line 49: `disabled={disabled || false}` here, so that Sourcereel, Sourcecast etc don't need to be changed.

seanlowjk: Aight Aight

src/components/workspace/controlBar/chapterSelect.tsx
seanlowjk: 💯

seanlowjk:
Big Kudos to @angelsl for reminding about Optional Props :D

PR #1356, opened 2020-07-12
Add sorting extension to sort all imports

sciffany:
### Description

This PR adds simple-import-sort dependency, which sorts the imports in order of
1) Module imports
2) Relative imports
3) Absolute imports

and then alphabetically

Ignores two files because the sorting rule contradicts the order required by import dependencies

### Type of change

### How to Test

Shuffle imports and either run `yarn eslint --fix` or save file if your prettier saveOnFormat is true.

### Checklist

- [ ] I have tested this code


.eslintignore
```diff
@@ -0,0 +1,2 @@
+src/commons/editor/Editor.tsx
+src/commons/sideContent/SideContentSubstVisualizer.tsx
```
angelsl, line 2: Maybe just

```tsx
/* eslint-disable simple-import-sort/sort */
```

at the top of these files instead of disabling ESLint entirely on those files?

PR #1371, opened 2020-07-19
WIP Add external symbols for typechecking

openorclose:
### Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

### Type of change

### How to Test

Load a library using the dropdown (and not by `import`), type a constant from that library and ctrl shift m to typecheck. There should not be an error saying that it is undefined.

### Checklist

- [x] I have tested this code


martin-henz:
This is a hopelessly outdated PR, and addresses a legacy component of the Source Academy: external symbols. The new feature is modules.

Issue #1369, opened 2020-07-18
Side content fails to correctly update active tab in Redux store


Aulud:
## [Bug] Side content fails to correctly update active tab in Redux store
Major bug.

### Bug description
After the refactor of `SideContent` to support dynamic tabs in PR #1352, the value of the selected tab of the current workspace (`sideContentActiveTab` prop) in the Redux store is no longer correctly updated whenever the selected tab is changed by the user.

This breaks the run-all functionality of the Autograder when it is the currently selected tab, since this is executed via a condition in the saga for `EVAL_EDITOR` which checks that the value of the selected tab of the current workspace is `SideContentTypes.Autograder`. This can be verified by logging the value of the side content tab observed in the saga below (see `activeTab` in the relevant code section below).

https://github.com/source-academy/cadet-frontend/blob/8f617f50885e1f282f7ede6a9fc2379372614c58/src/commons/sagas/WorkspaceSaga.ts#L710-L716

### Replication instructions
1. Navigate to any assessment or submission, then proceed to any **programming** (not MCQ) question
2. Bring up the side-content Autograder (airplane icon), and evaluate the editor's contents, and verify that none of the testcases are run

Last updated 18 Jul 2020, 11:45PM

bnjmnt4n:
I believe this bug still exists in a more subtle form: when navigating from 1 page in an assessment to another (not sure if I got the terminology right), if I'm on the test cases tab in the first page, upon navigation to the second page the test cases tab will still be shown, but the expected behaviour of clicking run to evaluate test cases will not work. Only after manually switching tabs back and forth to test cases again would this work.

I'm guessing the easiest way to fix this might be to update https://github.com/source-academy/cadet-frontend/blob/5bf07b67b3dac4676a99a9c95657c6698523e353/src/commons/sideContent/SideContent.tsx#L60-L64 to either reset the tab to the default tab on page change or set it to the same tab as before, although I'm not too sure how the redux store is structured to make it do so on page change.

Issue #812, opened 2019-08-06
Refactor evalCode saga for Autograder run-all functionality


Aulud:
## Refactor evalCode saga for Autograder run-all functionality
Summary: Deferred refactoring to continue support for Mission Autograder run-all functionality from #801.

### Todos
- Remove `sideContentActiveTab` attribute from Redux store along with the associated action `UPDATE_ACTIVE_TAB` with its action creator and reducer, in line with the project to clean up the Redux state tree
- Remove conditional behaviour in `evalCode` saga for run-all testcase functionality
- Re-implement run-all testcase functionality in line with #607 by constructing a different 'Run' button in assessment and grading workspaces, to be passed to `ControlBar`
   - This button will contain its own internal `activeTab` prop inherited from the state of the containing workspace, and will invoke either `EVAL_EDITOR` or `EVAL_EDITOR_WITH_TESTCASES` depending on the value of its `activeTab` prop
   - `activeTab` state of the containing workspace will be updated by the `onChange` handler of the respective workspace's `SideContent` component
- Add a new action `EVAL_EDITOR_WITH_TESTCASES` with its action creator, reducer and saga to be invoked by the alternative 'Run' button
- Repair all broken tests from the refactor

chownces:
Addressed by #1860

PR #1375, opened 2020-07-19
Fix side content firing handleActiveTabChange on every re-render

Aulud:
## [Bugfix] Fix side content firing handleActiveTabChange on every re-render
Summary: Fixes issue #1369.

### Changelog
- Modify the `handleActiveTabChange` effect to only set the initial active tab once when the component mounts
  - This effect was migrated from the `componentDidMount` lifecycle method, which only runs once during component mounting
  - Since `tabs` is in its dependency array and changes every re-render, this effect runs every re-render and sets the active tab in the Redux store to the default tab, overriding the active tab set by the tab change callback - this is clearly unintended
  - This requires the empty dependency array `[]`, and follows the recommendations set out in the React documentation [here](https://reactjs.org/docs/hooks-reference.html#useeffect) for `React.useEffect`
- Memoise the rendered side-content tabs and tab change callback
  - This avoids re-rendering the side-content tabs if both the rendered tabs (`dynamicTabs`) and workspace has not changed, predicated on the parent workspace having memoised the base tabs (`tabs` prop)
    - For example, typing into the REPL should not lead to the side-content tabs re-rendering
    - This behaviour is currently limited to only the Playground, which has been refactored to memoise the base tabs passed to `SideContent`
  - The definition of some pure functions, e.g. `renderTab`, have been moved into the corresponding effects to minimise their scope and avoid unnecessary re-computation on each re-render

### Verification instructions
1. Navigate to any assessment or submission, then proceed to any **programming** (not MCQ) question
2. Bring up the side-content Autograder (airplane icon), and evaluate the editor's contents (click the `Run` button or press Shift-Enter), and verify that all the testcases are run
   - The Autograder requires the value of the active tab in that workspace to be `SideContentType.autograder` for it to run all testcases, showing the active tab is updated correctly
   - Alternatively, you can log the value of `activeTab` below to verify it matches the currently selected tab
    https://github.com/source-academy/cadet-frontend/blob/8f617f50885e1f282f7ede6a9fc2379372614c58/src/commons/sagas/WorkspaceSaga.ts#L710-L716

Last updated 21 July 2020, 5:00PM

### DIFF
--- a/src/commons/sideContent/SideContent.tsx
+++ b/src/commons/sideContent/SideContent.tsx
@@ -53,7 +53,7 @@ type StateProps = {
 };
 
 const SideContent = (props: SideContentProps) => {
-  const { tabs, handleActiveTabChange, defaultSelectedTabId } = props;
+  const { tabs, defaultSelectedTabId, handleActiveTabChange, onChange } = props;
   const [dynamicTabs, setDynamicTabs] = React.useState(tabs);
   const workspaces = useSelector((state: OverallState) => state.workspaces);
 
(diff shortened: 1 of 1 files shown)


Aulud:
@anthony-halim No problem haha. This is a temporary fix for now - I had some work on another laptop to refactor the active tab state out of the Redux store (issue #812) entirely for the Autograder, but it's stalled for a bit as that laptop is undergoing keyboard repairs. Hopefully will manage to get a PR up for it before the semester starts.

PR #1387, opened 2020-07-23
Added utility to report student infinite loops to Sentry

JoeyChenSmart:
### Description

An infinite loop detector was added to js-slang (see https://github.com/source-academy/js-slang/pull/575).
In an effort to improve this detector, it will be helpful to collect code written by students with infinite loops as example/test data.

This utility will use Sentry to log infinite loop errors made by the student.
Both infinite loop errors detected by the infinite loop detector, and errors that slip past (and appear as stack overflows) will be sent to Sentry.

image

Note that all the infinite loops will be logged under a single Sentry issue (all events logged have the same Sentry 'fingerprint')

image

Code can be found in Sentry's report (as above).

All of the data can be later downloaded using the [Sentry API](https://docs.sentry.io/api/events/get-group-events/).

### Type of change

### How to Test

2. Run any program that will cause an infinite loop. Example: `function z(x){return z(x);}z(1);`
3. Check that the issue "Error: Infinite Loop" is added on the Sentry dashboard (similar to pic above)

### Checklist

- [x] I have tested this code


martin-henz:
So you are storing the error "Infinite Loop" as well as the offending "Code". Do you also store which type of infinite loop was detected? 
* stack overflow (not detected by infinite loop detector)
* various types of loops that the infinite loop detector detects
@JoeyChenSmart

src/commons/utils/InfiniteLoopReporter.ts
```diff
@@ -0,0 +1,162 @@
...
+      idx++;
+    }
+    if (idx >= topArr.length) {
+      const top = stack[stack.length - 1];
+      result += ']';
+      if (top && top[1] !== top[0].length) result += ',';
+    }
+  }
+  return result;
+}
```
angelsl, line 81: You could probably use `JSON.stringify` directly, but with a replacer passed in as the 2nd argument that checks and removes circular structures and/or structures that are too deep.

src/commons/utils/InfiniteLoopReporter.ts
```diff
@@ -0,0 +1,162 @@
...
+ */
+function infiniteLoopErrorType(errorString: string): string {
+  if (errorString.includes('Infinite recursion (or runtime error) detected')) {
+    if (errorString.includes('forget your base case')) return 'no_base_case';
+    else if (
+      errorString.includes('Did you call a value that is outside the range of your function?')
+    )
+      return 'input_out_of_domain';
+    else if (errorString.includes('Check your recursive function calls.')) return 'no_state_change';
+  }
```
angelsl, line 19: Might be better to have a special error type enum where these errors are thrown (in js-slang) so that this won't break if someone changes the error message wording

src/commons/utils/InfiniteLoopReporter.ts
```diff
@@ -0,0 +1,162 @@
...
+  else if (errorString === 'InternalError: too much recursion') return 'stack_overflow';
+  // chrome
+  else if (errorString === 'RangeError: Maximum call stack size exceeded') return 'stack_overflow';
+  // edge
+  else if (errorString === 'Error: Out of stack space') return 'stack_overflow';
+  // error from transpiler
+  else if (errorString.includes('Potential infinite recursion detected'))
+    return 'source_protection_recursion';
+  else if (errorString.includes('Potential infinite loop detected'))
+    return 'source_protection_loop';
```
angelsl, line 30: Same as L12-19

src/commons/utils/InfiniteLoopReporter.ts
```diff
@@ -0,0 +1,162 @@
...
+    )
+      return 'input_out_of_domain';
+    else if (errorString.includes('Check your recursive function calls.')) return 'no_state_change';
+  }
+  // firefox
+  else if (errorString === 'InternalError: too much recursion') return 'stack_overflow';
+  // chrome
+  else if (errorString === 'RangeError: Maximum call stack size exceeded') return 'stack_overflow';
+  // edge
+  else if (errorString === 'Error: Out of stack space') return 'stack_overflow';
```
angelsl, line 25: I'm a bit wary of hardcoding this since they could very well change between browser versions (and Safari is missing here, of the few popular browsers), but I'm not sure if there's any better way..

JoeyChenSmart:
> So you are storing the error "Infinite Loop" as well as the offending "Code". Do you also store which type of infinite loop was detected?
Yup, forgot to mention it but the 'type' will be recorded under a tag
image

> 
> * stack overflow (not detected by infinite loop detector)
> * various types of loops that the infinite loop detector detects
>   @JoeyChenSmart
> 
> Another question: Are we just catching stack-overflow, or also "time limit exceeded" ?

We are catching all errors which we can recognize as infinite loops, namely

- loops the infinite loop detector detects
- loops that slip past but are caught by the transpiler (the time limit exceeded ones that say 'Potential infinite loop detected...')
- stack overflows

This way we can also collect infinite loop data for Source 3&4, which are not supported by the current infinite loop detector

src/commons/utils/InfiniteLoopReporter.ts
JoeyChenSmart: hmm my main concern was that JSON.stringify will quote the function code, and I couldn't find a satisfactory (non hacky) way to avoid this.
i.e. `[1,"two", (x)=>x]` will be stringified to `"[1, \"two\", \"(x)=>x\"]"` instead of the desired `"[1, \"two\", (x)=>x]"`

src/commons/utils/InfiniteLoopReporter.ts
JoeyChenSmart: ok I'll work on adding this to js-slang as an enum, same with L27-30.

JoeyChenSmart:
I made a new PR https://github.com/source-academy/js-slang/pull/687 in js-slang, adding the above Enum object. (Also moved the `infiniteLoopErrorType` function there).

Also made changes to this repo to use that function, but I'm not sure if I should push these changes now, or just wait until the changes to js-slang are made.

src/commons/utils/InfiniteLoopReporter.ts
JoeyChenSmart: I did some testing and it turns out Safari's error message is the same as Chrome's. (added this info as a comment in the code). Also moved this bit to an enum in js-slang

martin-henz:
> > So you are storing the error "Infinite Loop" as well as the offending "Code". Do you also store which type of infinite loop was detected?
> 
> Yup, forgot to mention it but the 'type' will be recorded under a tag
> image
> 
> > * stack overflow (not detected by infinite loop detector)
> > * various types of loops that the infinite loop detector detects
> >   @JoeyChenSmart
> > 
> > Another question: Are we just catching stack-overflow, or also "time limit exceeded" ?
> 
> We are catching all errors which we can recognize as infinite loops, namely
> 
> * loops the infinite loop detector detects
> * loops that slip past but are caught by the transpiler (the time limit exceeded ones that say 'Potential infinite loop detected...')
> * stack overflows
> 
> This way we can also collect infinite loop data for Source 3&4, which are not supported by the current infinite loop detector

Excellent!

martin-henz:
Before you close this issue, please write a wiki page in the frontend wiki on this. Basically just summarize the discussion above.

JoeyChenSmart:
> Before you close this issue, please write a wiki page in the frontend wiki on this. Basically just summarize the discussion above.

I've summarized the discussion in a wiki page here https://github.com/JoeyChenSmart/cadet-frontend/wiki/Infinite-Loop-Reporting 
Just need someone to merge it with the main wiki once this is closed.

angelsl:
> Also made changes to this repo to use that function, but I'm not sure if I should push these changes now, or just wait until the changes to js-slang are made.

I've merged that and bumped js-slang here (on master). You can push those changes now :)

JoeyChenSmart:
> > Also made changes to this repo to use that function, but I'm not sure if I should push these changes now, or just wait until the changes to js-slang are made.
> 
> I've merged that and bumped js-slang here (on master). You can push those changes now :)

Alright, thanks! Made the changes.

PR #1398, opened 2020-07-29
Disable the REPL for the WebAssembly variant

### LINKED ISSUE #1392: Playground REPL with Source 1 WASM repeats past error messages
## [Bug] Playground REPL with Source 1 WASM repeats past error messages
Interpreter or REPL-related bug.

### Bug description
Evaluating a program in the REPL with the Source sublanguage set to Source 1 WebAssembly returns an output that repeats all previous errors in the REPL **if at least one exists**, even if the program is syntactically and semantically sound. Additionally, it appends any new errors to all past REPL outputs (see below).

We begin by evaluating an invalid program (a valid program here does not manifest any bug in subsequent steps). The REPL now contains this, which is correct:

image

Evaluating another invalid program returns an output that includes the previous error (line 1 of the first program) with the current error (line 2 of the second program). It also appends the current error to the output of the previous program. Both of these are likely unintended - this is the REPL now:

image

If we continue by evaluating a valid program (`true;`), the expected output (`true`) is not produced, but instead the combined error is output as the result of this third program - this is the REPL now:

image

If we continue again by evaluating now an invalid program (this time with an undeclared variable `z`), we notice the output (red error message) does not contain the _undeclared variable error_; instead, the error is reported through the hint message instead - this is the REPL now:

image

For comparison, this is the standalone output from evaluating that same invalid program in the REPL, which does correctly report the _undeclared variable error_:

image

More oddly, if we continue by evaluating an empty program, the REPL repeats the combined errors as the output, even though there should really be none - this is the REPL now:

image

This bug **does not occur with any other Source sublanguage**, so it may be isolated to the implementation of Source 1 WASM.

### Replication instructions
1. Proceed to the Playground either in your local development front-end or [here](https://source-academy.github.io/playground)
2. Set the Playground's sublanguage to _Source 1 WebAssembly_
3. Evaluate in the REPL a sequence of both valid and invalid Source 1 programs into the REPL, beginning with an invalid program
    - With an invalid program, the REPL should append the new error to all previous outputs from invalid programs
    - With a valid program, the REPL will fail to output the intended value, instead returning an output with all previous error messages

Last updated 27 Jul 2020, 11:45PM

btzy:
### Description

This PR disables the REPL for the `wasm` variant in a similar way as the `concurrent` variant.  Running `yarn install` modified yarn.lock, so I also added the new lock file to this PR.

Fixes #1392.

### Type of change

### How to Test

Rebuild cadet-frontend, set the language variant to Source 1 WebAssembly, and observe that the REPL input box is now hidden.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/repl/Repl.tsx
+++ b/src/commons/repl/Repl.tsx
@@ -47,7 +47,9 @@ class Repl extends React.PureComponent<ReplProps, {}> {
             className={classNames('repl-input-parent', 'row', Classes.CARD, Classes.ELEVATION_0)}
             handlers={handlers}
           >
-            {this.props.sourceVariant !== 'concurrent' ? <ReplInput {...inputProps} /> : null}
+            {this.props.sourceVariant !== 'concurrent' && this.props.sourceVariant !== 'wasm' ? (
+              <ReplInput {...inputProps} />
+            ) : null}
           </HotKeys>
         </div>
       </div>
--- a/src/pages/playground/Playground.tsx
+++ b/src/pages/playground/Playground.tsx
@@ -419,7 +419,10 @@ const Playground: React.FC<PlaygroundProps> = props => {
         persistenceButtons,
         executionTime
       ],
-      replButtons: [props.sourceVariant !== 'concurrent' ? evalButton : null, clearButton]
+      replButtons: [
+        props.sourceVariant !== 'concurrent' && props.sourceVariant !== 'wasm' ? evalButton : null,
+        clearButton
+      ]
     },
     editorProps: {
       sourceChapter: props.sourceChapter,


btzy:
@Aulud Oops, I didn't realise those buttons existed.  I've also removed that button now.

btzy:
I'm not sure why "4 weeks" has changed to "3 weeks" in this test... I tried removing the second commit but this issue still remains.

```
FAIL src/commons/achievement/__tests__/AchievementTask.tsx
  ● AchievementView component renders correctly
    expect(received).toMatchSnapshot()
    Snapshot name: `AchievementView component renders correctly 1`
    - Snapshot
    + Received
    @@ -27,11 +27,11 @@
                          </svg>
                        </span>
                      </Blueprint3.Icon>
                      <span style={{...}}>
                        <p>
    -                     4 Weeks
    +                     3 Weeks
                        </p>
                      </span>
                    </div>
                  </AchievementDeadline>
                  <AchievementExp exp={250}>
      18 |   const sampleComponent = <AchievementTask {...mockProps} />;
      19 |   const tree = mount(sampleComponent);
    > 20 |   expect(tree.debug()).toMatchSnapshot();
         |                        ^
      21 | });
      22 | 
      at Object.<anonymous> (src/commons/achievement/__tests__/AchievementTask.tsx:20:24)
 › 1 snapshot failed.
```

angelsl:
> I'm not sure why "4 weeks" has changed to "3 weeks" in this test...

It's probably not due to changes made here. I think there was a time-dependent snapshot test merged by accident...

martin-henz:
yea a bit of a pain. for now I'm just updating the test cases. The incentives team can fix this later.

martin-henz:
wait a bit. I'm merging my PR and then you can just re-base, and you'll be all set.

martin-henz:
ok, test cases pass now.

PR #1417, opened 2020-08-02
Implement Keystroke Logging

seanlowjk:
### Description

This PR adds new keystroke logging features for both `Playground` and `AssessmentWorkspace`

For `Playground`, the **following** are recorded. 

1. Chapter Select 
2. External Library Select 
3. Code Changes 
4. Cursor Changes 
5. Highlighting Changes 

For `AssessmentWorkspace`, the **following** are recorded. 

1. Code Changes 
2. Cursor Changes 
3. Highlighting Changes 

### Type of change

### How to Test

**Playground**

1. Go into `Playground` 
2. Make any of the 5 changes listed above. 
3. Click Run
4. You should see in your console the payload 

**AssessmentWorkspace**

1. Go into **any assessment**  
2. Make any of the 3 changes listed above. 
3. Click Save
4. You should see in your console the payload 

### TODOs 
 
1. Remove console.log before production 
2. Fix backend URL as when required 

### Checklist

- [x] I have tested this code


src/commons/application/reducers/SessionsReducer.ts
```diff
@@ -88,6 +91,23 @@ export const SessionsReducer: Reducer<SessionState> = (
         ...state,
         gameState: action.payload
       };
+    case ADD_KEYSTROKE_LOG:
+      const logsCopy = state.keystrokeLogs;
```
anthony-halim, line 95: Use [...state.keystrokeLogs] to ensure that the object is copied instead of its reference being copied.

src/commons/application/types/SessionTypes.ts
```diff
@@ -1,4 +1,5 @@
 import { Grading, GradingOverview } from '../../../features/grading/GradingTypes';
+import { Input } from '../../../features/sourceRecorder/SourceRecorderTypes';
```
anthony-halim, line 2: Move `Input` into `ApplicationTypes.ts`

src/commons/sagas/BackendSaga.ts
```diff
@@ -602,6 +604,27 @@ function* BackendSaga(): SagaIterator {
...
+    const respMsg = yield postKeystrokeLogs(tokens, questionId, playbackData);
+    if (!respMsg) {
+      yield handleResponseError(respMsg);
+    } else if (respMsg === 'OK') {
+      yield call(showSuccessMessage, 'Keystrokes Uploaded successfully!', 2000);
+    } else {
+      yield call(showWarningMessage, respMsg, 10000);
+      return;
+    }
+    yield put(actions.fetchAssessmentOverviews());
```
anthony-halim, line 625: Please add a comment why uploading keystroke logs continues to fetching assessment overviews.

src/features/keystrokes/KeystrokesHelper.tsx
```diff
@@ -0,0 +1,20 @@
+export const ONE_HOUR_IN_MILLISECONDS = 3600000;
+const ONE_BYTE_IN_BITS = 8;
+const ONE_KB_IN_BYTES = 1024;
+const FIVE_MB_IN_KB = 5 * 1024;
```
anthony-halim, line 4: SHOUTY convention is saved for actions, just use camelCase for this. Additionally, constants should be in `KeystrokesTypes.ts` instead

src/features/keystrokes/KeystrokesHelper.tsx
```diff
@@ -0,0 +1,20 @@
...
+    }
+  }
+  return allStrings ? 3 + (allStrings.length * 16) / (ONE_BYTE_IN_BITS * ONE_KB_IN_BYTES) : 0;
+};
+
+export const hasExceededLocalStorageSpace = () => {
+  return getLocalStorageSpace() > FIVE_MB_IN_KB;
+};
+
+export const PLAYGROUND_QUESTION_ID: number = -1;
```
anthony-halim, line 20: Same as the above

src/commons/sagas/BackendSaga.ts
seanlowjk: Oops might have accidentally screwed up, that shouldn't exist.

src/commons/application/types/SessionTypes.ts
seanlowjk: Technically, most of the recording types are defined in `SourceRecorderTypes`, so it wouldn't make much sense to just shift `Input` into `ApplicationTypes` 

If that's the case, we need to shift everything into `ApplicationTypes` which might not be very nice 

Maybe @jiayushe can also advice on this >.<

jiayushe: Hmm what I usually do in React projects is to put all the data types in one folder, for example `src/types/<component>.ts`, while leaving the action constants in the individual action files.

seanlowjk: Gotcha! 

So in this case would it make sense to shift the declared types of `Input` etc etc all to `ApplicationTypes` since we would be using them in `Playground` and `AssessmentWorkspace` now given that we would doing keystroke recording for these 2 components? o.O

anthony-halim: You can either pull `SourceRecorder` into commons, or just put the `Input` into the `ApplicationTypes`. Point being is `Input` is used by more than one feature, hence should be under `commons/`.

seanlowjk: 👍

src/features/keystrokes/KeystrokesHelper.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const playgroundPlayback: PlaybackData = JSON.parse(
+    playgroundLogs ? playgroundLogs : JSON.stringify(defaultPlaybackData)
+  );
+  playgroundPlayback.inputs.push(newInput);
+  localStorage.setItem('PlaygroundLogs', JSON.stringify(playgroundPlayback));
+
+  const neww = localStorage.getItem('PlaygroundLogs');
+  const hehexd: PlaybackData = JSON.parse(neww ? neww : JSON.stringify(defaultPlaybackData));
+  console.log(hehexd);
+};
```
blackening, line 59: Which part of the browser prevents them from opening more than 1 playground tab?

What happens when they do so?

What happens when 1 set of logs upload while the user is still on the playground?

src/pages/playground/Playground.tsx
```diff
@@ -175,13 +197,61 @@ const Playground: React.FC<PlaygroundProps> = props => {
...
+        assessmentLogs
+      );
+    }
+
+    if (playgroundLogs.inputs.length !== 0) {
+      props.handleKeystrokeUpload(playgroundQuestionId, playgroundQuestionId, playgroundLogs);
+    }
+
+    resetPlaygroundLogging(props.sourceChapter, props.externalLibraryName, props.editorValue);
+  }, [props]);
```
blackening, line 218: What happens when the logs are uploaded when the user is still on the playground?

src/pages/playground/Playground.tsx
```diff
@@ -175,13 +197,61 @@ const Playground: React.FC<PlaygroundProps> = props => {
...
+
+  const uploadPerHour = React.useCallback(() => {
```
blackening, line 220: Knowing that 70kb of data was generated with short tests, do you think a 1 hour interval is suitable?

What happens when all logs are uploaded 1 hour after the start of the practical exam?

src/pages/playground/Playground.tsx
```diff
@@ -175,13 +197,61 @@ const Playground: React.FC<PlaygroundProps> = props => {
...
+    const interval = setInterval(() => {
+      uploadLogs();
+    }, oneHourInMilliSeconds);
+    return () => clearInterval(interval);
+  }, [uploadLogs]);
+
+  // Second useEffect called here just in case.
+  React.useEffect(() => {
+    uploadPerHour();
+  }, [uploadPerHour]);
```
blackening, line 230: Just in case isn't a good enough reason to put a chunk of code in. Please consider what needs to be done for correctness.

blackening: Especially since your just in case looks like this:

```
 public uploadPerHour = () => {
    const interval = setInterval(() => {
      this.uploadLogs();
    }, oneHourInMilliSeconds);
    return () => clearInterval(interval);
  };
```

src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
```diff
@@ -166,6 +205,46 @@ class AssessmentWorkspace extends React.Component<
...
+    const assessmentLogs = getAssessmentLogs();
+    const playgroundLogs = getPlaygroundLogs();
+
+    if (assessmentLogs.inputs.length !== 0) {
+      this.props.handleKeystrokeUpload(
+        assessmentIDs.assessmentId,
+        assessmentIDs.questionId,
+        assessmentLogs
+      );
+    }
```
blackening, line 226: Have you tried uploading logs when the system is offline? 

Does error propagation prevent logs from being reset?

blackening (dismissed):
Please fix all correctness problems.

src/features/keystrokes/KeystrokesHelper.tsx
blackening: Also what kind of naming is that.

blackening: Original lines being reviewed.
```
export const savePlaygroundLog = (newInput: Input) => {
  const playgroundLogs: string | null = localStorage.getItem('PlaygroundLogs');
  const playgroundPlayback: PlaybackData = JSON.parse(
    playgroundLogs ? playgroundLogs : JSON.stringify(defaultPlaybackData)
  );
  playgroundPlayback.inputs.push(newInput);
  localStorage.setItem('PlaygroundLogs', JSON.stringify(playgroundPlayback));

  const neww = localStorage.getItem('PlaygroundLogs');
  const hehexd: PlaybackData = JSON.parse(neww ? neww : JSON.stringify(defaultPlaybackData));
  console.log(hehexd);
};
```

seanlowjk: O wait I forgot to remove the console.logs. Psps

src/pages/playground/Playground.tsx
seanlowjk: In that case would it be better to set a limit to the # of logs that are stored? 

Then a time limit might not be necessary as such!

src/features/keystrokes/KeystrokesHelper.tsx
seanlowjk: If they open more than one tab, the keystroke will will still get pushed onto our `localStorage` !

src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
seanlowjk: Yep, there is supposed to be some form of mechanism which helps to do so. 

Unfortunately, I think the stuff I wrote didn't get committed >< 

let me commit it now!

src/pages/playground/Playground.tsx
seanlowjk: Yea actually as long as we have the following checks whenever we push a log: 

1) If localStorage hit capacity OR 
2) if # of logs is way too big 

We can just send the payload immediately. Then mayeb a timer is not necessary? What is your opinion on this? ><

src/pages/playground/Playground.tsx
seanlowjk: localStorage is reset wrt to playground logs **if and only if** the logs are uploaded successfully

src/features/keystrokes/KeystrokesHelper.tsx
blackening: And if you try to reconstruct the state of both tabs what happens?

src/pages/playground/Playground.tsx
blackening: Which piece of code says so? How do you know if the logs are successfully uploaded?

src/pages/playground/Playground.tsx
```diff
@@ -214,7 +215,9 @@ const Playground: React.FC<PlaygroundProps> = props => {
       props.handleKeystrokeUpload(playgroundQuestionId, playgroundQuestionId, playgroundLogs);
     }
 
-    resetPlaygroundLogging(props.sourceChapter, props.externalLibraryName, props.editorValue);
+    if (getResetLoggingFlag()) {
```
blackening, line 217: This won't work, async issues.

seanlowjk:
Mainly now what will happen is that when a student 'tries' to upload his or her logs, we will check if they have successfully done so. 

if so, we can reset the logs, else we don't 

What do we mean by reset? We reset by deleting those logs from our storage that have been sent before

image

src/pages/playground/Playground.tsx
blackening: i think timer is okay, just need to make sure there's 1 timer, and it's something like

interval + random offset within interval, otherwise all the uploads will bunch together and nuke the system.

seanlowjk:
yes you are right @blackening 

As a result, i will need to save the logs from sessionStorage to localStorage before the page closes. Then as discussed in the group, 

I will upload the pending logs together

blackening:
Deployment instructions:

1. Grab the cadet-logger and throw it on a server.
2. After that, go cadet-frontend/.env and add in the logging URL. In my case: 
REACT_APP_CADET_LOGGER=http://localhost:8001/assessment-logger

3. yarn build, ignore the error message regarding publicPath, then yarn build-serviceworker.

-------------------

Architecture:
- All instances of SA will insert logs independently into indexedDB.
- ServiceWorker (guaranteed to be singleton) will grab logs and upload them periodically.

angelsl (approved):

PR #1428, opened 2020-08-04
Hide nav bar buttons on sm/md views to prevent overflow

jiayushe:
### Description

Hide nav bar buttons on sm/md views to prevent overflow

### Type of change

### How to Test

- Check that no overflow in academy nav bar even in sm/md view

### Checklist

- [x] I have tested this code


jiayushe:
Yea that's intended. The texts for staff buttons are hidden for `xs`, `sm` and `md`, while the rest are only hidden for `xs` and `sm`. This is because the staff buttons will still cause overflow in the nav bar in `md` width.

PR #1432, opened 2020-08-04
Fix game not terminating properly after leaving page

angelsl:
### Description

Fix the game not terminating properly when navigating away from the game page.

If a user clicks on the "Source Academy" button, React Router actually navigates
to /academy, and then immediately back to /academy/game. This often leads to
multiple instances of the game running (just repeatedly click on the Source
Academy logo button and you can easily reproduce it).

The reason is that the navigation back to /academy/game causes the Game page to
be reloaded, and `createSourceAcademyGame` called, before the destructor (the
lambda returned from `useEffect`) of the initial Game page runs.

Thus when the destructor eventually runs, `SourceAcademyGame.getInstance()` now
refers to the **new** game instance. The old game instance is left in the
background.

```
React.useEffect(() => {
  createSourceAcademyGame();
  return () => {
    SourceAcademyGame.getInstance().isMounted = false;
    SourceAcademyGame.getInstance().stopAllSounds();
  };
}, []);
```

Calling `destroy` also explicitly terminates all the event loops etc. set up by
Phaser immediately which should help with the memory and CPU issues persisting
after leaving the Game page.

### Type of change

### How to test

Open the game and repeatedly click the "Source Academy" logo button. See (or actually, hear) that there is only ever one game instance.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/pages/academy/game/Game.tsx
+++ b/src/pages/academy/game/Game.tsx
@@ -16,10 +16,11 @@ function Game() {
   const [isUsingMock, setIsUsingMock] = React.useState(false);
 
   React.useEffect(() => {
-    createSourceAcademyGame();
+    const game = createSourceAcademyGame();
     return () => {
-      SourceAcademyGame.getInstance().isMounted = false;
-      SourceAcademyGame.getInstance().stopAllSounds();
+      game.isMounted = false;
+      game.stopAllSounds();
+      game.destroy(true);
     };
   }, []);
 


src/pages/academy/game/Game.tsx
```diff
@@ -16,10 +16,11 @@ function Game() {
...
 
   React.useEffect(() => {
-    createSourceAcademyGame();
+    const game = createSourceAcademyGame();
     return () => {
-      SourceAcademyGame.getInstance().isMounted = false;
-      SourceAcademyGame.getInstance().stopAllSounds();
+      game.isMounted = false;
+      game.stopAllSounds();
+      game.destroy(false);
```
anthony-halim, line 23: This should be `game.destroy(true)`. During testing, I found that `game.destroy(false)` leads to weird behaviour in which the game needs to be loaded at least twice to work. So far, `game.destroy(true)` seems to do the bug fix while not leading to weird behaviour.

PR #1475, opened 2020-08-19
Hide autograding results until submission is manually graded

angelsl:
You may want to ignore whitespace changes when reviewing this PR.

### Description

Hide autograding results until submission is manually graded.

The profile tab needs to be updated too. @Aulud 

### Type of change

### How to test

Look at an assessment that has been autograded but not manually graded.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/assessment/Assessment.tsx
+++ b/src/commons/assessment/Assessment.tsx
@@ -341,55 +341,56 @@ class Assessment extends React.Component<AssessmentProps, State> {
     index: number,
     renderAttemptButton: boolean,
     renderGradingStatus: boolean
-  ) => (
-    <div key={index}>
-      <Card className="row listing" elevation={Elevation.ONE}>
-        <div className="col-xs-3 listing-picture">
-          <NotificationBadge
-            className="badge"
-            notificationFilter={filterNotificationsByAssessment(overview.id)}
-            large={true}
-          />
-          <img
-            alt="Assessment"
-            className={`cover-image-${overview.status}`}
-            src={overview.coverImage ? overview.coverImage : defaultCoverImage}
-          />
-        </div>
-        <div className="col-xs-9 listing-text">
-          {this.makeOverviewCardTitle(overview, index, renderGradingStatus)}
-          <div className="listing-grade">
-            <H6>
-              {beforeNow(overview.openAt)
-                ? `Grade: ${overview.grade} / ${overview.maxGrade}`
-                : `Max Grade: ${overview.maxGrade}`}
-            </H6>
+  ) => {
+    const showGrade = overview.gradingStatus === 'graded' || overview.category === 'Path';
+    return (
+      <div key={index}>
+        <Card className="row listing" elevation={Elevation.ONE}>
+          <div className="col-xs-3 listing-picture">
+            <NotificationBadge
+              className="badge"
+              notificationFilter={filterNotificationsByAssessment(overview.id)}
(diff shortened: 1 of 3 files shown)


src/commons/assessment/Assessment.tsx
```diff
@@ -341,55 +341,58 @@ class Assessment extends React.Component<AssessmentProps, State> {
...
-                ? `XP: ${overview.xp} / ${overview.maxXp}`
-                : `Max XP: ${overview.maxXp}`}
-            </H6>
-          </div>
-          <div className="listing-description">
-            <Markdown content={overview.shortSummary} />
+  ) => {
+    const showGrade =
+      (renderGradingStatus && overview.gradingStatus === 'graded') ||
+      (beforeNow(overview.openAt) && overview.category === 'Path');
```
Aulud, line 347: I suppose only requiring the assessment be fully graded for non-paths is intentionally more lenient than also requiring the assessment be submitted? This allows students to still view the total grade/XP when their Avenger unsubmits a fully graded submission (since we no longer delete the manual grading data during unsubmission), e.g. to facilitate "corrections", which is probably desirable.

However, the escape hatch for paths will still display the total grade/XP for open paths prior to submission (both of which will be 0 anyway until autograding runs at submission), which is at odds to other assessment types.

image

I'm not sure but the `beforeNow(overview.openAt)` condition feels redundant, since it depends on the user's system time, which is out of our control. Additionally, it may inconvenience staff if they're trying to test out a path's testcases (especially private ones) prior to its actual open date, by hiding the total grade and XP awarded by autograding when they submit.

We can probably replace it with `overview.status === 'submitted'` instead, since that's the point where the total grade/XP awarded by the autograding becomes relevant to the student for paths. Students also shouldn't be able to upload answers nor finalise their submission if the actual time is before the backend's system time, so this condition is probably stronger overall.

Edit: We might want to remove `renderGradingStatus` from the condition also, since staff are now allowed to submit assessments for testing purposes - right now it hides the grading status, total grade and XP for staff submissions on assessments under 'Upcoming', i.e. those nearing their open date. The sidequest below is a staff submission that is fully graded.

image

Aulud (commented):
Works as intended for the autograding results in the Autograder, but we might need to consider more scenarios for when the total grade/XP should be shown in the assessment listings. :\

src/commons/assessment/Assessment.tsx
angelsl: > However, the escape hatch for paths will still display the total grade/XP for open paths prior to submission (both of which will be 0 anyway until autograding runs at submission), which is at odds to other assessment types.

I was thinking of treating paths as just being 'graded' all the time.

> I'm not sure but the `beforeNow(overview.openAt)` condition feels redundant, since it depends on the user's system time, which is out of our control. Additionally, it may inconvenience staff if they're trying to test out a path's testcases (especially private ones) prior to its actual open date, by hiding the total grade and XP awarded by autograding when they submit.

Fair. So we take time out of the consideration. Show the score if it is graded, or it is a path?

Aulud: > I was thinking of treating paths as just being 'graded' all the time.

Ah okay, that works.

> Fair. So we take time out of the consideration. Show the score if it is graded, or it is a path?

Yep, sounds good!

Aulud (approved):
Looks good to go.

PR #1484, opened 2020-08-24
Create a new canvas when re-initialising WebGL

angelsl:
### Description

Create a new canvas when re-initialising WebGL (instead of reusing the old one).

Also comment out some unused functions.

Fixes curves not working if the WebGL context was initially initialised for runes.

### Type of change

### How to test

Open playground, switch to runes, then switch to curves, and then try to render a curve. It should work.

### Checklist

- [x] I have tested this code


Aulud (approved):
The fix works but... yeah it seems odd both `open_viewport` and `getReadyWebGL` remain in the graphics library even though they're never invoked anywhere.

PR #1487, opened 2020-08-26
Fix grading editor not resetting on question change

angelsl:
### Description

Fix the grading editor not resetting when clicking next/previous.

Bug introduced by https://github.com/source-academy/cadet-frontend/commit/802a24b8ccdda4c23fb270dacc72c314ca885b04. The grading editor has been fundamentally broken though, and only worked because of the bug that was fixed in that commit.

### Type of change

### How to test

Open the grading workspace and go between questions. See that the grading editor resets to show the comment of the question navigated to upon each navigation.

### Checklist

- [x] I have tested this code


Aulud (approved):
Works as advertised - I'm worried now about what other skeletons we unleashed from the closet when we fixed the base `Application` unnecessarily re-rendering.

PR #1490, opened 2020-08-28
Add ability to disable the frontend during specified periods

angelsl:
### Description

Add the ability to disable the frontend during specified periods.

image

### Type of change

### How to test

Configure a disablement period locally and see that it renders the Disabled page during the disablement period, and renders the original page after. (Note that any state in the original page is lost, of course.)

### Checklist

- [x] I have tested this code


Aulud (approved):
Have got it working locally, and it correctly prevents me from accessing any part of the Academy except for the modal components in the dropdown (user profile, help dialog, etc.).

The implementation seems fine, so it looks good for the upcoming RA (assuming other playgrounds such as https://source-academy.github.io are also disabled).

PR #1492, opened 2020-08-30
Allow staff to login when academy is disabled

angelsl:
### Description

Allow staff to login and bypass disablements.

### Type of change

### How to test

Set a disablement and see that the login page is accessible when logged out.

Log in as staff or admin and see that the academy works as expected.

Log in as student and see that the academy is still disabled.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/utils/Constants.ts
+++ b/src/commons/utils/Constants.ts
@@ -43,19 +43,21 @@ for (let i = 1; ; ++i) {
 
 const disablePeriods: Array<{ start: Moment; end: Moment; reason?: string }> = [];
 
-for (let i = 1; ; ++i) {
-  const startStr = process.env[`REACT_APP_DISABLE${i}_START`];
-  const endStr = process.env[`REACT_APP_DISABLE${i}_END`];
-  if (!startStr || !endStr) {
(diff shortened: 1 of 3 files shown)


src/commons/utils/Constants.ts
```diff
@@ -43,19 +43,21 @@ for (let i = 1; ; ++i) {
...
-  const endStr = process.env[`REACT_APP_DISABLE${i}_END`];
-  if (!startStr || !endStr) {
-    break;
-  }
-  const reason = process.env[`REACT_APP_DISABLE${i}_REASON`];
-  const start = moment(startStr);
-  const end = moment(endStr);
-  if (end.isBefore(start) || moment().isAfter(end)) {
-    continue;
+if (!test) {
```
Aulud, line 46: Is this supposed to be `isTest`? `test` isn't defined anywhere in `Constants.ts`.

Aulud (changes requested):
Have tested locally and haven't found a way to access any SA features as a non-authenticated/student user. Seems to work for staff and admin users - so it's good to go (apart from that typo below).

PR #1506, opened 2020-09-02
Fix playground hash links

### LINKED ISSUE #1472: Source language selection in query string not working
https://source-academy.github.io/playground#chap=1

does not reliably open Source §1. Instead the language choice seems to be over-ridden with some local state.
 
The URL query string should always override any other playground setting.

Similarly, the selection of the library is not working.

This is important for the proper functioning of the textbook.

Screenshot 2020-08-18 at 6 33 12 PM

angelsl:
### Description

Fix playground hash links.

Fixes #1472.

### Type of change

### How to test

Change your language to something other than Source 4 with no library, then try this link: http://localhost:8000/playground#chap=4&prgrm=PTAEGUEkGEAVQFLlARgHQCZMChsDMBXAOwGMAXASwHsjQ8KAjACiIEpQBvbUH0AJwCmZAn1q0AvJNAAGbr3m8A-DLkL5ALlASpKVWvnLd+tZvrNaAWlTsA1HUYtQVjKwDc2AL64zTAGxugA

It should reliably change the language to Source 4 with no library.

### Checklist

- [x] I have tested this code


src/commons/sagas/WorkspaceSaga.ts
```diff
@@ -459,16 +458,6 @@ export default function* WorkspaceSaga(): SagaIterator {
     }
   }
 
-  /**
```
Aulud, line 462: Nice, we're removing more dead code.

src/pages/playground/Playground.tsx
```diff
@@ -104,6 +107,36 @@ export type StateProps = {
 
 const keyMap = { goGreen: 'h u l k' };
 
+function handleHash(hash: string, props: PlaygroundProps) {
```
Aulud, line 110: Nice this is actually cleaner than the previous function `parsePlayground`, which honestly did not need the entire list of helper parsing functions.

Aulud (approved):
Found no issues in testing, and the overall fix looks in line with what I expected (moving it out of `Application.tsx` and into the Playground component itself). It's a bit disturbing how much the props required by the Playground gets chained all the way down from Application.

PR #1511, opened 2020-09-06
update Inspector's list of built ins 

podocarp:
### Description

This list keeps track of names of all the built ins so that they will not be displayed when inspecting.

### Type of change

### How to Test

Run with breakpoint. Notice on the live version you get some built ins such as `yellow` displayed under global scope. This version does not display them.

### Checklist

- [x] I have tested this code


Aulud (commented):
Hmm, just an observation - this doesn't seem to be the full list of built-ins though (for example, `stack_frac` is included in this list but not `beside_frac`, even though both are from the RUNES library).

martin-henz (approved):
looks good

PR #1513, opened 2020-09-09
Improve Pix & Flix Library 

rrtheonlyone:
### Description

This PR improves the current `Pix & Flix` library by:
- improving code quality of internal functions
- adding runtime checks to prevent browser crashing
- updating pixel representation 

Users using the library will be able to apply filters by manipulating a 2D grid of pixels where each pixel is a `[Red, Blue, Green, Alpha]` value. The change here from the original implementation is that we also allow users to modify the degree of opacity with the `Alpha` value. 

We also add in an `_FPS` field which is set to `20` by default. This is number of frames we draw per second on the screen. In future, this can perhaps be something users can adjust - those with a more powerful system may be able to handle higher FPS values. 

Runtime checks are done on the filter function passed in by the user. If there is any error, we just reset the filter and print a message to user.

We also played around with using GPU Acceleration on the video library code.

### Type of change

- [X] This change requires a documentation update

### How to Test

This is a link containing the code snippets that can be accelerated by the GPU: https://share.sourceacademy.nus.edu.sg/gpucodesnippetsrr 

To test:
- copy this library over and override the current one
- run local version of `cadet-frontend` and head over to playground
- switch library to `pix & flix` and try it out!

### Checklist

- [X] I have tested this code
- [X] I have updated the documentation


martin-henz (approved):
The library is a big cleanup. Played a bit. Works much better.

PR #1515, opened 2020-09-09
SOUNDS: Use 16 bits per sample

### LINKED ISSUE #1514: SOUNDS: 8 bits per sample sounds bad
```
const test_sound = sine_sound(400, 0.4);

const soft_sound = make_sound(t => get_wave(test_sound)(t) / 2 / 2 / 2 / 2 / 2 / 2, get_duration(test_sound));
play(soft_sound);
```

Something like this sounds very bad with 8 bits per sample, the (default) setting we're currently using for riffwave.js

There's a reason why most DAWs don't even support 8 bits per sample -- it sounds very bad when sounds get soft.

thomastanck:
### Description

Makes sounds sound less bad.

Fixes #1514 

### Type of change

### How to Test

Deploy locally and test following code on playground with sounds library enabled:

```
const test_sound = sine_sound(400, 0.4);

const soft_sound = make_sound(t => get_wave(test_sound)(t) / 2 / 2 / 2 / 2 / 2 / 2, get_duration(test_sound));
play(soft_sound);
```

It should not sound terrible (like a sawtooth/square wave) and instead should just sound like a soft sine wave.

### Checklist

- [x] I have tested this code

### DIFF
--- a/public/externalLibs/sound/sounds.js
+++ b/public/externalLibs/sound/sounds.js
@@ -57,10 +57,11 @@ function discretize_from(wave, duration, elapsed_duration, sample_length, data) 
     }
 }
 
-// Quantize real amplitude values into standard 4-bit PCM levels
+// Quantize real amplitude values into standard 16-bit PCM levels
+// The range is [-32768, 32767] (see riffwave.js)
 function quantize(data) {
     for (var i = 0; i < data.length; i++) {
-        data[i] = Math.round((data[i] + 1) * 126);
+        data[i] = Math.floor(data[i] * 32767.999);
     }
     return data;
 }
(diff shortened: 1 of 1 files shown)


public/externalLibs/sound/sounds.js
```diff
@@ -60,7 +60,7 @@ function discretize_from(wave, duration, elapsed_duration, sample_length, data)
 // Quantize real amplitude values into standard 4-bit PCM levels
 function quantize(data) {
     for (var i = 0; i < data.length; i++) {
-        data[i] = Math.round((data[i] + 1) * 126);
+        data[i] = Math.round(data[i] * 32767);
```
martin-henz, line 63: why not "+ 1"? what is the range for the result?

public/externalLibs/sound/sounds.js
```diff
@@ -60,7 +60,7 @@ function discretize_from(wave, duration, elapsed_duration, sample_length, data)
 // Quantize real amplitude values into standard 4-bit PCM levels
```
martin-henz, line 60: should say 16-bit?

thomastanck:
Fixed.

public/externalLibs/sound/sounds.js
thomastanck: The range is [-32768, 32767] (see riffwave.js)

(added a comment as well)

martin-henz (approved):

PR #1517, opened 2020-09-17
Fix Selected MCQ Option being carried over to all questions

rrtheonlyone:
### Description

There is an issue where once we select an option for a MCQ question, it updates all the other questions. The reason is because the state `mcqOption` in `McqChooser.tsx` is not being updated even if the `props` change.

The state here is simply unnecessary, refer to: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html. 

Removing it fixes the problem (tested on staging backend)

### Type of change

### How to Test

1. Open any assessment with MCQ questions
2. Choose an option
3. See if it is carried over to other questions

### Checklist

- [X] I have tested this code

### DIFF
--- a/src/commons/mcqChooser/McqChooser.tsx
+++ b/src/commons/mcqChooser/McqChooser.tsx
@@ -10,24 +10,14 @@ export type McqChooserProps = {
   handleMCQSubmit: (choiceId: number) => void;
 };
 
-type State = {
-  mcqOption: number | null;
-};
-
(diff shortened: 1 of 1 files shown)


angelsl (approved):

Aulud (approved):
I'm surprised we have not encountered this bug before previously; this derived state has been around since https://github.com/source-academy/cadet-frontend/pull/228 - maybe it got unearthed by the PR that fixed the re-rendering of the base `Application`.

PR #1532, opened 2020-10-11
Rename apply_filter to install_filter in Pix&Flix

rrtheonlyone:
### Description

Small change to naming. Also improved error message printing a little


martin-henz (approved):
Great. Yes install_filter is a better name.

PR #1536, opened 2020-10-13
Pix&Flix: Display warning message when applied filters error.

bnjmnt4n:
### Description

Display a warning message when the applied filter errors. This adds a callback mechanism to the _VD library.

Note: seems like the video_lib.js file was using CRLF, and I edited it to LF. https://github.com/source-academy/cadet-frontend/pull/1536/commits/f4cc2d3d7a360a5176605f68716da64225aaec03?diff=unified&w=1 provides a more meaningful non-whitespace diff.

### Type of change

### How to Test

Note: I have not tested the code since I couldn't get the setup running locally yet.

### Checklist

- [ ] I have tested this code

### DIFF
--- a/public/externalLibs/video/video_lib.js
+++ b/public/externalLibs/video/video_lib.js
@@ -53,7 +53,7 @@ function alpha_of(px) {
 }
 
 /**
- * Assigns the red, green, blue and alpha components of a pixel 
+ * Assigns the red, green, blue and alpha components of a pixel
  * <CODE>px</CODE> to given values
  * @param {px} Pixel - given Pixel
  * @param {r} Number - the red component as a number between 0 and 255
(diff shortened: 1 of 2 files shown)


rrtheonlyone:
Thank you for the PR and your contribution 💯 

Yes, passing in a callback is efficient and it is a good way to display on the academy. However, there are 2 pain points in your current solution:
- It is better to show the error in the REPL rather than as a toast notification which disappears after awhile
- You may have realised many of our errors are actually `undefined` - this is because they originate from `js-slang` and we need to parse before displaying them 

I have built upon your idea in this PR if you want to have a look: https://github.com/source-academy/cadet-frontend/pull/1535. Great work on coming up with this solution!

bnjmnt4n:
Cool, I didn't see #1535 when I started on this PR last night. Seems like it addresses most of the issues with working with the pixnflix library in a much more thorough manner than mine! Displaying the toast was I guess a stopgap measure as I couldn't really figure out how the REPL worked with just a cursory glance of the code.

PR #1539, opened 2020-10-21
Fix calls to undefined errLogger in video_lib

angelsl:
Fixes this issue:

image

### DIFF
--- a/public/externalLibs/video/video_lib.js
+++ b/public/externalLibs/video/video_lib.js
@@ -183,7 +183,7 @@ _VD.loadMedia = function() {
     if (!navigator.mediaDevices.getUserMedia) {
         const errMsg = 'The browser you are using does not support getUserMedia';
         console.error(errMsg);
-        _VD.errLogger(errMsg, false);
+        _VD.errLogger && _VD.errLogger(errMsg, false);
         return;
     }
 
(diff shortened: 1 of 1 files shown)


thomastanck (approved):
LGTM

rrtheonlyone:
I was under the impression that this error is caused by an outdated library version (fixed by force refresh). How do I actually reproduce this error?

We always do pass in the necessary call back when initialising the `_VD` object:

https://github.com/source-academy/cadet-frontend/blob/789101774dd6a29109f8e2aaf9de04f19e12da48/src/commons/sideContent/SideContentVideoDisplay.tsx#L47-L56

angelsl:
> I was under the impression that this error is caused by an outdated client (fixed by force refresh).

Yeah, seems like you're right. Sentry shows this error is only happening for [this](https://github.com/NUS-CS1101S/cadet-frontend/commits/0bbb4d86056fb148e0a1add4dfe345b80f2821f3) and [this](https://github.com/NUS-CS1101S/cadet-frontend/commits/767bf4b2b69327e855c79bb19dbaaa74535f430c) deploy. In that case this PR isn't going to solve anything then, since they won't get the update unless they force-refresh anyway..

Closing.

PR #1574, opened 2020-12-20
Update and refactor list visualizer code

### LINKED ISSUE #943: DATA VIZ: toplevel pair with null head
```draw_data(list(null))```

gives 

IMAGE 2019-09-23 09:42:32

but should give

Screenshot 2019-09-23 at 9 38 58 AM

### PR DESCRIPTION
**Pull request by jeffsieu**
#### Code refactoring
Updates the list visualizer code to use ES6 style javascript.
This includes:
- Using `const`, `let` instead of `var`
- Using `class` instead of accessing objects' `prototype`

Also, redundant code repetition have been incrementally removed commit by commit, mainly via the use of OOP.
Overall motivation is to simply clean up the code a bit.

#### Changes to output behavior
Also, this changes `pair(null, null)` to show as 
image
instead of 
image

In other words, whether `null` is the `head `or `tail` of a `pair`, it will show up as a crossed box.
### Type of change

Fixes #943 

### How to test

```js
const f = x = >x;
const hh = pair(null, f);
const hh_head_tail = pair(f, null);
const hh_head = pair(null, hh_head_tail);
set_head(hh_head, hh);
set_head(hh, hh_head);
draw_data(hh);
```
image

This code uses most of all the features of `draw_data`, and looks the same before and after the changes.

### Checklist

- [x] I have tested this code


angelsl:
There was a missing `super()` call in `NullDrawable`'s constructor. Fixed that.

angelsl (approved):

PR #1586, opened 2021-01-16
Fix environment visualiser bugs

CrownKira:
### Description

- Fix repeated elements in missingEnvs array which result in crash upon invoking draw_env function
- Fix frame width calculation issues which result in canvas getting cut off
- Refactor some code to use ES6 syntax
- Remove some irrelevant comments

### Type of change

### How to test

```javascript
let x = null;
for (let i = 0; i < 5; i = i + 1) {
    x = pair(() => i, x);
}

debugger;
```
```javascript
function f(n) {
    if (n <= 1) {
        debugger;
        return 1;
    } else {
        return n * f(n-1);
    }
}

f(5);
```

### Checklist

- [x] I have tested this code

### DIFF
--- a/public/externalLibs/env_visualizer/visualizer.js
+++ b/public/externalLibs/env_visualizer/visualizer.js
@@ -4,25 +4,25 @@
   | Configurations
   |--------------------------------------------------------------------------
   */
-  const container = document.createElement("div");
-  container.id = "env-visualizer-container";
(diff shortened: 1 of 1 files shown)


angelsl (approved):
Seems to work, thanks. Please avoid including formatting changes in the same commit as a functional change, otherwise it's basically impossible to see what functional changes were made..

PR #1593, opened 2021-01-26
Add color points to CURVES library

zhenghanlee:
### Description

- Added `make_color_point(x, y, r, g, b)` constructor function to CURVES library
- Added new Point class to replace the old Point construct ([x, y] pair) to support the additional z-dimension and color property
- Updated transformation functions to support the new implementation of Point objects

### Type of change

### How to test

- Run an instance of cadet-frontend, and navigate to an instance of playground
- Include CURVES library
- Call constructor function `make_color_point(x, y, r, g, b)` to construct a color point
- To construct a color curve, declare a function that takes in t as a parameter (ranging from 0 to 1) and returns a color point
- To draw curves, use `draw_connected_*` and `draw_points_on_*` functions
- Use `rotate_around_origin`, `rotate_pi_over_2`, `scale_x_y`, `translate` and other transformation functions to transform color curves
- Use `connect_rigidly` and `connect_ends` to connect two (color) curves

### Checklist

- [x] I have tested this code

### To-Do List

- [ ] Documentation to be updated


MontyPython28:
Seems like the previous issues were fixed- all functions are working properly now.
Just one small note- when I set the RGB values to negative numbers or numbers greater than 255, the curve created is the same as the curve created when I use 0 and 255 respectively. (E.g. `make_color_point(-200, -1, 300)` creates the same curve as  `make_color_point(0,0,255)`)

martin-henz:
> Seems like the previous issues were fixed- all functions are working properly now.
> Just one small note- when I set the RGB values to negative numbers or numbers greater than 255, the curve created is the same as the curve created when I use 0 and 255 respectively. (E.g. `make_color_point(-200, -1, 300)` creates the same curve as `make_color_point(0,0,255)`)

Maybe we can document this behaviour in the specs for make_color_point?

martin-henz (approved):
As approved by @MontyPython28

zhenghanlee:
> > Seems like the previous issues were fixed- all functions are working properly now.
> > Just one small note- when I set the RGB values to negative numbers or numbers greater than 255, the curve created is the same as the curve created when I use 0 and 255 respectively. (E.g. `make_color_point(-200, -1, 300)` creates the same curve as `make_color_point(0,0,255)`)
> 
> Maybe we can document this behaviour in the specs for make_color_point?

Yes I have updated js-slang accordingly.

PR #1626, opened 2021-02-18
Fix dialog box disabling mobile workspace while on non-mobile device

chownces:
### Description

Fixed the above bug, and renamed isMobile booleans to isMobileBreakpoint to clearly denote that these booleans only check for the mobile breakpoint of 768px.

Previously (dialog box pops up even on non-mobile device):
image

Now:
image

### Type of change

### How to test

- Resize desktop browser such that width > height. The dialog box should not appear.
- Dialog box should still appear in the mobile workspace on mobile devices when in landscape

### Checklist

- [x] I have tested this code


CrownKira (approved):
Tested and issue is fixed. lgtm!

PR #1636, opened 2021-03-01
3D transformations

yongxiangng:
### Description
Add 3D translation and 3D scaling for curves. Resolved name conflict between rune and curve lib (when translate and scale are used) by renaming translate to translate_curve and scale to scale_proportional in the rune library. I have yet to update the documentation.

`translate_curve` has an optional 3rd argument to dictate amount to be translate in z-axis.
`scale_x_y_z` scale in x direction by first argument, y direction by second argument, z direction by third argument. third argument is optional (maybe delete `scale_x_y`?)
`scale_proportional` now uses `scale_x_y_z` in webGLhi_graph_ce.js
`rotate_around_origin` and `rotate_pi_over_2` now works with 3d curves. Used to return a 2d curves
`deriv_t` works with 3d curves although it is not available on SA (it's not in ExternalTypes.ts).
`gosper_curve` uses composition of other transformation

_**To be tested before merging**_
`translate_curve`
`scale_x_y`
`scale_x_y_z`
`scale_proportional`
`connect_rigidly`
`connect_ends`
`rotate_around_origin`
`rotate_pi_over_2`
`gosper_curve`

### Type of change

### How to test
```
let curve1 = t=>make_3D_color_point(t, t, t, 255*math_cos(t), 255*math_sin(t), t*255);
let curve2 = t=>make_3D_color_point(t, t*t*t, t, 255*math_cos(t), 255*math_sin(t), t*255);

draw_3D_connected(500)(translate_curve(0.1, 0.1)(curve1));
draw_3D_connected(500)(translate_curve(0.1, 0.1, 0.1)(curve1));

draw_3D_connected(500)(scale_x_y(1, 2)(curve1));
draw_3D_connected(500)(scale_x_y_z(1, 2, 3)(curve1));
draw_3D_connected(500)(scale_proportional(1.1)(curve1));

draw_3D_connected_full_view_proportional(500)(connect_rigidly(curve1,curve2));

draw_3D_connected_full_view_proportional(500)(connect_ends(curve1,curve2));

draw_3D_connected(500)(rotate_around_origin(math_PI/4)(curve1));

draw_3D_connected(500)(rotate_pi_over_2(curve1));

draw_3D_connected_full_view_proportional(500)(gosper_curve(7));
```

### Checklist

- [x] I have tested this code


shermanyuen (approved):
- `translate_curve`: It works with two arguments and three arguments. A suggestion is to further overload this function if only one argument is provided, such that only the x-direction is translated.
- `scale_x_y` and `scale_x_y_z`: They work. Yes I also suggest to only have one function, maybe `scale`, that can take on one argument (only scales the x-direction), two arguments (only scales the x- and y-directions) or three arguments (scales all three directions). `scale_proportional` will remain as a different function from this new `scale` such that `scale_proportional(s) = scale(s, s, s)`.
- `connect_rigidly` and `connect_ends`: They work.
- `rotate_around_origin` and `rotate_pi_over_2`: Noted that these functions rotate the curve along the z-axis, and it is not possible to use these functions to rotate along the x- or y-axes.
- `gosper_curve`: Noted that this function returns a curve different from the traditional definition of the Gosper curve:
gosper
canvas

Issue #1555, opened 2020-11-02
Workspace and editor refactor and optimisation


angelsl:
Creating this tracking issue to plan what I need to do during the winter break.

# Refactoring

## Workspace component duplication

There are currently five copies of the workspace component that contains the editor:

- Playground: this one has basically all of the features enabled
- Assessment
- Grading
- Sourcereel
- Sourcecast

Each of these components started as copies of each other. The Playground is the most "up-to-date", as it is where all the CP3108 and CS4215 development went into first. I also optimised it over summer, so it's now the fastest workspace out of the five (more on this later).

The biggest issue with this is that each of the workspaces deviate ever so slightly from each other, and a lot of logic is duplicated across the components (e.g. when to show certain side content tabs, etc).

## Workspace state

Each of the 5 workspaces have their state in the Redux store. This makes no sense whatsoever, as the bulk of the state is local to the workspace, and should not be in the Redux store. (In fact I would argue that Redux is a mistake for this project, because we don't have that much global state.)

Because the state is in the Redux store, we use Sagas to deal with many things. There is no way for Sagas to return results back to the workspaces other than by updating the Redux store, and this has led to a gradual ballooning of the Redux store workspace state. Each state object is now huge!

## The plan

- Identify what is common across all the workspaces, and pull them out into a unified Workspace component.
  - The main things that differ across the workspaces are the buttons that are enabled and the side content panels that are enabled.
- Trim down the workspace state as far as possible. Ideally, remove it all from the Redux store, so that we don't have "singleton" workspaces. (Right now, it's virtually impossible to create a new workspace without also adding it to the Redux store!)

# Optimising  

Currently in the assessment workspace, each keystroke causes at least 6 React renders at minimum.

Some latency measurements, by dumping keystrokes using `libinput debug-events`, recording my screen at the refresh rate (60 Hz) and inspecting the playback frame-by-frame, and counting the frames between the "key pressed" output from `libinput` to the character appearing in the editor

- Geany (a code editor, C++): consistently 0 frames
- VS Code (code editor, JS Monaco): 0-4 frames
- Source Academy Playground: 0-3 frames
- Source Academy Assessment workspace: 4-13 frames

There is clearly something wrong with the Assessment workspace.

PR #1649, opened 2021-03-08
adding support for file persistence using GitHub

chekjun:
### Description

Added Open, Save, Quick save features for GitHub.

### Type of change

### How to test

Deploy the front-end. Check that the GitHub control buttons now have options for "Open", "Save", "Save As" and "Log In".
Click on log in and sign in to GitHub.
Click on Open, select repository, then select a file to load.
Click on Save As, select repository, then select a file or name a new file to save over.
Click on Save to quick save current work into previously selected file.


src/commons/gitHubOverlay/ConfirmDialog.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    handleGitHubCancelConfirmationDialog();
+  };
+
+  const cancelHandler = () => {
+    handleGitHubCancelConfirmationDialog();
+  };
+
+  return (
+    <Dialog isOpen={isOpen} usePortal={false}>
```
chownces, line 40: not sure if this can be done with `src/commons/utils/DialogHelper.tsx`

chownces (changes requested):
Looks good overall! Some other minor comments:
- Are there any plans to make the Github login state persist through refreshes? Currently, we have to relogin after a refresh, or leaving the Playground. This feature might also be useful when implementing the 'backend-light' features
- Might want to add an 'alert' when the user successfully logs in or logs out (see Blueprint's Toaster component)
- To implement the file extension feature where committed files are of `*.js`
- The GithubCallback page/ messages from the previous PR is to be styled (currently it simply displays black text over the background image, and is a little unreadable)

src/commons/application/ApplicationTypes.ts
```diff
@@ -302,7 +302,18 @@ export const defaultSession: SessionState = {
...
+  githubLoginID: '',
+  githubName: '',
+  githubEmail: '',
+  userRepos: [],
+  githubRepositoryFilepath: '',
+  githubSaveMode: '',
+  githubRepositoryName: '',
+  githubCommitMessage: '',
+  pickerType: '',
+  isPickerOpen: false,
```
chownces, line 315: can also check with @angelsl if all these are needed in the Redux store (especially the isPickerOpen and isDialogOpen) as i think there are plans to trim down on the store and remove state that is local to the workspace #1555.

and also, im assuming these are placed under `defaultSession` instead of `defaultPlayground` as there are plans to make the Github login persist through refreshes and across workspaces

angelsl (commented):
Oh my, please don't stuff everything in the Redux store!

src/commons/sagas/GitHubPersistenceSaga.ts
```diff
@@ -45,7 +55,170 @@ function* githubLoginSaga() {
...
+      committer: { name: githubName, email: githubEmail },
+      author: { name: githubName, email: githubEmail }
+    });
+
+    showSuccessMessage('Successfully created file!', 1000);
+    yield put(actions.setPickerDialogStatus(false));
+  } catch (err) {
+    console.error(err);
+    showWarningMessage('Something went wrong when trying to save the file.', 1000);
+  }
```
angelsl, line 221: Likewise for these. These can be done in a component (say, the GitHubOverlay) instead. State that is local to the GitHubOverlay should be tracked locally in that component (i.e. picker type and dialogue status).

And the GitHub login ID/name/email, that is just read off the Octokit instance, no need to have those as separate props in the Redux store.

angelsl: By the way. If you are selecting things from the store in a Saga, use [select](https://redux-saga.js.org/docs/api#selectselector-args). Please don't use the `store` global. Only do that if you are not in a Saga context and have no other way to get the information.

angelsl (changes requested):
Just a few notes on the overall architecture. I recognise you will likely have to restructure your changes significantly, but we really cannot continue stuffing things into the Redux store like that.

src/commons/controlBar/ControlBarGitHubButtons.tsx
```diff
@@ -27,13 +27,12 @@ const stateToIntent: { [state in GitHubState]: Intent } = {
...
 
 export const ControlBarGitHubButtons: React.FC<ControlBarGitHubButtonsProps> = props => {
-  // The 'loggedInAs' is not used directly in this code block
-  // However, keeping it in will ensure that the component re-renders immediately
-  // Or else, the re-render has to be triggered by something else
-
   const isMobileBreakpoint = useMediaQuery({ maxWidth: 768 });
   const isLoggedIn = store.getState().session.githubOctokitInstance !== undefined;
+
   const shouldDisableButtons = !isLoggedIn;
```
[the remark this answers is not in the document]
ChengGeng97: It was a deliberate decision to disable rather than hide them so that the user would be immediately aware what functionalities the feature has.

martin-henz (approved):
The team confirmed that they have handled the outstanding change requests.

PR #1661, opened 2021-03-16
Bug fix and rotate_around_origin update

zhenghanlee:
#### `generateCurve`

The function now handles scaling and translation of curves manually on the coordinates of each generated point, and the cube no longer needs to be scaled or translated to fit the curve. This solves the following problem of the range of a certain axes being too small for the curve to render properly:
```
draw_3D_connected_full_view(1000)(t => make_3D_point(math_exp(-15)*t, math_sin(2*math_PI*t), math_pow(1-t, 2)));
```
If '-15' is further decreased pass a threshold, the program will treat the range of x to be exactly 0. This change also fixes the problem regarding cube not rendering when values of the coordinates are too large.
```
draw_3D_connected_full_view(1000)(t => make_3D_point(math_exp(111111)*t, math_sin(2*math_PI*t), math_pow(1-t, 2)));
```
Hence, the limitation on the coordinates has been removed. It will be documented that WebGL stops rendering points beyond a certain limit.
However, this change does not fix the following problem of rotations not working properly on `draw_3D_connected_full_view`:
```
let curve1 = t => make_3D_point(0, 0.5*math_cos(2*math_PI*t)+0.5, 0.5*math_sin(2*math_PI*t)+0.5);
// curve1 is a circle
let curve2 = rotate_around_origin(math_PI/4*(-2))(curve1);
// curve2 is a rotation of curve1 by 90 degrees clockwise
draw_3D_connected_full_view(1000)(curve2);
```
Users can go about this issue by adding a small value (e.g. '0.0000001') to the rotation, but that will incur a non-zero range (for y-axis in this case) and the curve will not be rendered parallel to the x-axis, as opposed to `draw_3D_connected_full_view_proportional`.

#### `rotate_around_origin`

The function now accepts 1 or 3 arguments (throwing error for 2 arguments). However, users have to understand that rotations in 3D are generally not commutative, and that the rotation called on 3 arguments (a, b, c) is an extrinsic rotation with Euler angles (a, b, c) about x, y, and z axis.

#### Removed Functions

Removed the following undocumented and unused functions:
    * `alternative_unit_circle`
    * `squeeze_rectangular_portion`
    * `squeeze_full_view`
    * `full_view_proportional`

### Type of change

### How to test

1. Pull and build js-slang from master branch, and [follow the steps here](https://github.com/source-academy/js-slang#using-your-js-slang-in-local-source-academy) to link it to the cadet-front end.
2. Run an instance of cadet-frontend, and navigate to playground
3. Include CURVES library

### Checklist

- [x] I have tested this code

### DIFF
--- a/public/externalLibs/graphics/tempCodeRunnerFile.js
+++ b/public/externalLibs/graphics/tempCodeRunnerFile.js
@@ -1,0 +1,1 @@
+scale_x_y_z
--- a/public/externalLibs/graphics/tempCodeRunnerFile.js
+++ b/public/externalLibs/graphics/tempCodeRunnerFile.js
@@ -1,1 +1,0 @@
-scale_x_y_z
--- a/public/externalLibs/graphics/webGLcurve.js
+++ b/public/externalLibs/graphics/webGLcurve.js
@@ -139,31 +139,23 @@ function generateCurve(scaleMode, drawMode, numPoints, func, space, isFullView) 
     -0.5,// amount to rotate in radians
     [0, 0, 1])     // axis to rotate around Z (dynamic)
     cubeRotation += 0.1 * Math.PI
+  } else {
+    min_z = max_z = 0
   }
 
   if (scaleMode == 'fit') {
-    var center = space == '3D' ? [(min_x + max_x) / 2, (min_y + max_y) / 2, (min_z + max_z) / 2] : [(min_x + max_x) / 2, (min_y + max_y) / 2]
+    var center = [(min_x + max_x) / 2, (min_y + max_y) / 2, (min_z + max_z) / 2]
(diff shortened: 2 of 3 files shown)


yongxiangng (approved):
`generateCurve`
Working fine as intended. As suggested by Sherman, we could set a limit. I've tried `var MAX_LIMIT = Math.max(max_x - min_x, max_y - min_y, max_z - min_z) * 0.01` and it works fine, and curve would show correctly with or without the 0.0000001 rotation. The MAX_LIMIT is 1% of the maximum dimension of the curve, i.e. if dimension along one axis is too lean comparatively, we will not scale it.

`rotate_around_origin`
Documentation would then require explicit mentions of the order which the rotations are applied. x then y then z. Also, the rotation works on the z axis when 1 argument is supplied (and not the x).

**Removed Functions**
All good

PR #1665, opened 2021-03-21
Add stepper functionality for multiple draw_data calls

jeffsieu:
### Description

Allows for visualization of multiple `draw_data` calls using a stepper-like UI. Stepping through the UI can be done using the left and right arrows on a keyboard (when the right panel is in focus), as well as through clicking the previous and next buttons

Calling
```js
for_each(draw_data, build_list(10, i => build_list(i+1, x => x)));
```
displays this:
image
(additional images omitted)
image
(additional images omitted)

### Type of change

### How to test

Call
```js
for_each(draw_data, build_list(10, i => build_list(i+1, x => x)));
```

or just `draw_data` in general.

### Checklist

- [x] I have tested this code


public/externalLibs/visualizer/visualizer.js
```diff
@@ -809,77 +856,112 @@
...
+        });
+        layer.add(text);
+      }
+      // adjust the position
+      layer.offsetX(minLeft - 20);
+      layer.offsetY(0);
+      layer.draw();
+
+      // Show the first structure
+      if (stepList.length == 0) {
```
CrownKira, line 912: Use ===  instead of == for stricter comparison.

public/externalLibs/visualizer/visualizer.js
```diff
@@ -809,77 +856,112 @@
...
-     * some of the drawing being cut off. Hence the width and height formulas used are approximations.
-     */
-    let stage = new Konva.Stage({
-      width: findListWidth(xs) * 60 + 60,
-      height: findListHeight(xs) * 60 + 100,
-      container: 'list-visualizer-container'
-    });
     minLeft = 500;
     nodelist = [];
     fnNodeList = [];
```
CrownKira, line 861: These two variables are not defined.

public/externalLibs/visualizer/visualizer.js
```diff
@@ -809,77 +856,112 @@
...
-    currentListVisualizer = layerList.length - 1;
+    // Set the first step as the step to be displayed
+    currentStepIndex = 0;
   }
 
   /**
-   *  Shows the layer with a given ID while hiding the others.
+   * Shows the step with a given id while hiding the others.
+   * 
+   * @param {Number} id The id of the step to be shown
```
CrownKira, line 931: Consider using lowercase for primitive types in JSDoc.

src/commons/utils/JsSlangHelper.ts
```diff
@@ -76,10 +76,11 @@ function cadetAlert(value: any) {
  *
  * @param list the list to be visualised.
  */
-function visualiseList(list: any) {
+function visualiseList(...xs: any[]) {
```
CrownKira, line 79: I think prof @martin-henz suggested during one of our meetings that we should standardize and spell visualize with a Z instead of S.

CrownKira (changes requested):
Some other minor eslint errors I wasn't able to include in the comment since they are part of the existing code: 
visualizer.js: 
- Consider removing useless constructors at line 170 and 120
- There is an unnecessary semicolon at line 1

CrownKira:
**String and numbers getting cut off**
Screenshot 2021-03-23 at 12 50 44 PM

**Side content not responsive sometimes**
It happened to me sometimes. I am not sure how to reproduce this. 
Maybe try this code and run it several times:
`for_each(draw_data, build_list(5, i => build_list(i, j => build_list(j+1, x => x+1))));`

Screenshot 2021-03-23 at 12 51 54 PM

martin-henz:
Currently this PR fails CI: https://github.com/source-academy/cadet-frontend/runs/2189556597?check_suite_focus=true

public/externalLibs/visualizer/visualizer.js
CrownKira: Is there a reason for using == instead of === for comparison? 
This rule has not been turned off so eslint is giving me an error on this line. 
https://eslint.org/docs/rules/eqeqeq

CrownKira (approved):

chownces (commented):
Hi! Something minor, but it would be nice if the scrolling on mobile could be fixed too. Currently, the user is unable to scroll the canvas if the scroll is initiated on a 'list structure' on touch devices. Also, the user is only able to scroll horizontally OR vertically (i.e. NOT diagonally), leading to a clunky user experience considering that long lists are drawn diagonally.

CrownKira:
> Hi! Something minor, but it would be nice if the scrolling on mobile could be fixed too. Currently, the user is unable to scroll the canvas if the scroll is initiated on a 'list structure' on touch devices. Also, the user is only able to scroll horizontally OR vertically (i.e. NOT diagonally), leading to a clunky user experience considering that long lists are drawn diagonally.

"By default Konva will prevent default behaviour off all pointer interactions with a stage.
That will prevent unexpected scrolling of a page when you are trying to drag&drop a shape on a mobile device."
We (env vis team) have encountered similar issues. This can be fixed by setting `preventDefault` property of a shape to `false`.
You may refer to this: 
https://konvajs.org/docs/events/Mobile_Scrolling.html

jeffsieu:
> Hi! Something minor, but it would be nice if the scrolling on mobile could be fixed too. Currently, the user is unable to scroll the canvas if the scroll is initiated on a 'list structure' on touch devices. Also, the user is only able to scroll horizontally OR vertically (i.e. NOT diagonally), leading to a clunky user experience considering that long lists are drawn diagonally.
> 
> Other than that, LGTM on mobile!

Hi! I've just made a new commit fixing this.

PR #1676, opened 2021-03-28
Environment Visualizer Rewrite

caipng:
### Description

**Summary of changes**
This is a complete rewrite of the existing Environment Visualizer. We have removed the old `visualizer.js`and `concrete.js` scripts from `public/externalLibs/env_visualizer/`. The feature now resides at `src/features/envVisualizer/` and is exposed via `EnvVisualizer.tsx` in that directory. The corresponding `SideContentEnvVisualizer.tsx` and `JsSlangHelper.ts` has been updated to use our rewritten visualizer.

The new visualizer uses the `Konva` and `React-Konva` canvas libraries. We make use of Typescript and try to follow OOP concepts with JS ES6 classes. Drawing logic and dimension calculations etc pertaining to a certain type (array, for example) are encapsulated within their own class files. Overall, this has led to better correctness, extensibility, maintainability, and collaboration; before, it was a single file with ~3K LoC.

This is a non-exhaustive list of changes/new features:
- Array visualization
- Global functions visualization
- Truncating on long texts
- Individual hovering of arrows/frames/function objects/etc.
- Numerous bugs/edge cases fixed
- Documentation

**New Dependencies:** `Konva`, `React-Konva`, `Konva-Node`

### Type of change

### How to test

~There is only manual testing for now.~ Tests and snapshots has since been added!

The code has been tested and verified against an extensive list of known problematic or complicated visualizations. A list of code samples may be found [here](https://sa-env-visualizer-beta-lscymrvuy-crownkira.vercel.app/).
```
const fn = () => "L";
const x = ["long string", pair(() => 1, () => 2), list(1, pair(2, 3), () => 3), () => "THIS", 5, 6];
const y = list(x[1], x[2], tail(x[1]), tail(x[2]), fn);

debugger;
```
Screenshot 2021-03-28 at 4 22 07 PM

### Checklist

- [x] I have tested this code

The code has been mostly documented. What remains is to update the [wiki page](https://github.com/source-academy/cadet-frontend/wiki/Environment-Model-Visualiser).


src/features/envVisualizer/components/Arrow.tsx
```diff
@@ -0,0 +1,118 @@
...
+  readonly y: number;
+  readonly height: number;
+  readonly width: number;
+  readonly points: number[];
+
+  constructor(readonly from: Visible, readonly to: Visible) {
+    this.x = from.x;
+    this.y = from.y;
+
+    if (from instanceof Frame) {
```
jeffsieu, line 27: Would it be better to define a "from coordinate" and "to coordinate" for arrow drawing that is overridden within each subclass of Visible, instead of accounting for all of them here?

src/features/envVisualizer/components/Level.tsx
```diff
@@ -0,0 +1,78 @@
...
+    /** the level above this */
+    readonly parentLevel: Level | null
+  ) {
+    this.x = Config.CanvasPaddingX;
+    this.y = Config.CanvasPaddingY;
+    this.parentLevel && (this.y += this.parentLevel.height + this.parentLevel.y);
+
+    // initialize frames
+    const frames: Frame[] = [];
+    if (this.parentLevel) {
```
jeffsieu, line 27: Is it possible to do something like:

```ts
if (this.parentLevel) {
  this.parentLevel.frames.forEach(
    frame =>
      frame.environment.childEnvs?.forEach(env => {
        const newFrame = new Frame(
          env,
          frame,
          frames.length > 0 ? frames[frames.length - 1] : null,
          this
        );
        frames.push(newFrame);
        env.frame = newFrame;
      })
  );
}

```

src/features/envVisualizer/components/Text.tsx
```diff
@@ -0,0 +1,107 @@
...
+      this.width = widthOf(truncatedText);
+      this.str = truncatedText;
+    } else {
+      this.width = Math.max(Config.TextMinWidth, widthOf(str));
+    }
+  }
+
+  onMouseEnter = ({ currentTarget }: KonvaEventObject<MouseEvent>) => {
+    const container = currentTarget.getStage()?.container();
+    container && (container.style.cursor = 'pointer');
```
jeffsieu, line 63: `container?.style.cursor = 'pointer';`

src/features/envVisualizer/components/Text.tsx
```diff
@@ -0,0 +1,107 @@
...
+    this.labelRef.current.show();
+    currentTarget.getLayer()?.draw();
+  };
+
+  onMouseLeave = ({ currentTarget }: KonvaEventObject<MouseEvent>) => {
+    const container = currentTarget.getStage()?.container();
+    container && (container.style.cursor = 'default');
```
jeffsieu, line 70: `container?.style.cursor = 'default';`

jeffsieu (changes requested):
LGTM, apart from some slight syntax issues

src/features/envVisualizer/components/Text.tsx
caipng: Optional property assignment is [not supported](https://github.com/tc39/proposal-optional-chaining#not-supported).

src/features/envVisualizer/components/Text.tsx
caipng: Optional property assignment is [not supported](https://github.com/tc39/proposal-optional-chaining#not-supported).

src/features/envVisualizer/components/Arrow.tsx
caipng: Thanks for your suggestion, this should be possible. However, one concern may be that we should not add too many fields to the classes and have it become harder to understand and maintain. 2 extra coordinates fields just for drawing of arrows added to each class may be unnecessarily adding more fields and having logic meant for drawing arrows be split into those classes. But I do see where you are coming from as well, since the coordinates to draw from and to are specific to the classes themselves. What are your thoughts? If you feel strongly about this I'll try to change it.

jeffsieu: Hmm. But I think drawing arrows is sort of a "core" function of the program since there will be many instances where this happens in more complicated structures. So I do see an advantage in moving the arrow logic into each of those classes.

caipng: alright, will change it up. thanks for the suggestion!

caipng: addressed [here](https://github.com/source-academy/cadet-frontend/pull/1676/commits/c5d80c1f9a7fb9f54f8a89deeef2d229ebf1587b)

jeffsieu (approved):
Looks good to me

PR #1697, opened 2021-04-15
Stepper: jumping feature

Zacchua:
### Description

implements the jumping feature which allows users to jump in between function calls of the same incarnation of the function. Requires src/index.ts and src/stepper/stepper.ts from js-slang sa2122-stepper-rebased branch

### Type of change

### How to test
```
function factorial(n) {
    return n === 1 ? 1 : n * factorial(n - 1);
}
factorial(5);
```
running the above code in the playground, user should be able to jump between steps 4, 12, 20, 28 and 36, which correspond to function call steps in the evaluation of factorial(5), using the >> and << buttons. 

### Checklist

- [x] I have tested this code


martin-henz (approved):
looks good to me

PR #1699, opened 2021-04-15
Auto-generate TypeScript client with types from backend's documented Swagger API

bnjmnt4n:
### Description

Using https://github.com/acacode/swagger-typescript-api, a TypeScript-based client with types is generated from the cadet backend's `swagger.json`.

TODO:
- [ ] Add documentation
  - Basic instructions
    - Copy `swagger.json` to `src/commons/api/`
    - Run `yarn run update-cadet-api`
- [ ] Avoid typecasting
- [ ] Ensure generated types are accurate
- [ ] Switch from `fetch` to the new API globally

### Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### How to test

Run the test suite.

### Checklist

- [ ] I have tested this code


src/commons/api/swagger.json
```diff
@@ -0,0 +1,3361 @@
+{
```
bnjmnt4n, line 1: This file is copied from cadet. Do you think there should be a yarn command to copy the file over for convenience?

src/commons/sagas/RequestsSaga.ts
```diff
@@ -125,25 +127,22 @@ export const getUser = async (tokens: Tokens): Promise<User | null> => {
  * Will be updated after a separate db for student progress is ready
  */
 export const getAchievements = async (tokens: Tokens): Promise<AchievementItem[] | null> => {
```
bnjmnt4n, line 129: I'm not sure about the semantics of redux-saga, but I'm sure there's a reason for each request function accepting tokens separately instead of just reading from the store. That being said, do you think a reasonable default for `customFetch` is to read from the store if no tokens are passed?

src/commons/sagas/RequestsSaga.ts
```diff
@@ -224,15 +218,11 @@ export async function bulkUpdateAchievements(
...
-    body: { achievements: achievements },
-    noHeaderAccept: true,
-    refreshToken: tokens.refreshToken,
-    shouldAutoLogout: false,
-    shouldRefresh: true
-  });
-
+  // TODO
+  const resp = await Cadet.adminAchievements.bulkUpdate(
+    (achievements as unknown) as Achievement[],
```
bnjmnt4n, line 223: There's still quite a bit of typecasting around the file, as there are still slight differences between the auto-generated types and the handwritten types. I'll look into all of these cases later.

bnjmnt4n (commented):
I wanted to get a cursory review of this PR, to see if we're going on the right track. This PR allows a TypeScript client with types defined as per the backend's Swagger API to be generated automatically using https://github.com/acacode/swagger-typescript-api. The API is properly typed to accept the correct input arguments (whether query, path, JSON or FormData), and should reflect the accurate types of the response as defined by the backend. (Note that there are still differences between the types which should be fixed before this can be merged.) The API client also parses the response into `resp.data` and `resp.error` respectively if an error occurs.

angelsl:
Yes, yes, so much yes. This is probably the best thing that's happened in the frontend code-quality wise ever since it was written. Looking forward to seeing this merged :)

martin-henz:
Hmm, anyone looking after this PR? @bnjmnt4n ? There is enthusiastic support from @angelsl . But still in draft and currently with merge conflict...

bnjmnt4n:
Sorry for the delays. I was planning to come back to it, and will take a look at it hopefully this weekend.

martin-henz:
Reviewing our open PRs and this one comes up again.  @bnjmnt4n do you have time these days? There is/was enthusiastic support from @angelsl . But still in draft and currently with merge conflict...

angelsl:
This will need quite a bit of time to properly convert all the places in the frontend that make requests to the backend. And we need to make sure the backend Swagger definitions are accurate as well (unfortunately Elixir is dynamically typed so it's not possible to generate Swagger definitions from type information...)

It'll be tedious but worth it.

PR #1700, opened 2021-04-16
Data visualizer: Improve visualization of arrays and improve overall design

jeffsieu:
### Description

Allows for visualization of arrays. Also improves on some of the existing issues with the design.

```js
const xs = list(1,2,3);
draw_data(xs, xs, xs, xs, xs, xs, xs, xs);
draw_data(xs, xs, xs, xs, xs, xs, xs, xs);
```
Previously:
image
Now:
image

#### Changes to Tree, TreeDrawer

Notice that now lists are no longer diagonal, but instead more vertical.
This comes as a by-product of revamping the calculation of positions to draw children. Previously, the TreeDrawer assumed that a tree node only has a left or a right child, and calculated offsets based on left and right children.
Now, the behavior of  tree node has been changed to allow for multiple children, aka a non-binary tree. As such, PairDrawable and PairTreeNode have been changed to ArrayDrawable and ArrayTreeNode respectively, which are more generalised versions of their predecessors that work for k children instead of just 2.

findNodeWIdth and findNodeHeight have also been reworked to accommodate for k children.

Now, the data visualizer is actually able to calculate the exact width and height of a drawing, thereby saving space (there was previously unnecessary padding added to account for edge cases that may not be present).

#### Visual changes
Each structure has now been wrapped inside a "card" to allow for a cleaner visual distinction between adjacent structures. Drawn structures should also be easier to show side by side, now that they assume a more vertical layout.

### Type of change

### How to test

```js
const h = pair(null, null);
set_tail(h, h);
const thing = [h, h, null, [h], list(1,2,3, [4,5,6],2,3,4,57,7,7,7,7,7)];
draw_data(thing);
draw_data(thing, thing, thing);
```
Observe that the visualizer is able to output the structure in a horizontally compact manner.
Expected output:
image

### Checklist

- [x] I have tested this code


martin-henz (approved):
I agree the new design is neater.

PR #1703, opened 2021-04-17
Refactor listVisualizer to dataVisualizer and minor fixes

VimuthM:
### Description

Refactored all instances of ListVisualizer to DataVisualizer to maintain consistency and to represent the tool more accurately.

### Type of change

- [X] This change requires a documentation update

### Checklist

- [X] I have tested this code


chownces (changes requested):
- Bug with Hotkeys. To reproduce: call `draw_data` __once__, and press the left or right arrow keys with the data visualizer tab in focus (the app shld break)
- The scrolling issue on mobile (touch devices) is back again: the tab cannot be scrolled when the scrolling is initiated on a drawn structure
- Currently, the UI is slightly different on the mobile version (as the mobile playground is styled separately)
    - (Need to check again with the Modules team) But I think it would be good if you could separate out the CSS for the dataviz tab, and style it once and for all in the React component (and remove it from the SCSS file)
    - This would also facilitate the eventual possibility of migrating the entire dataviz logic to the new modules system
- I need to double check with the team regarding the new UI for the dataviz tab (I think the single structure call with the new background looks a little weird now)
Ok:
image

May look a little weird with the background color, and on its own:
image

- It would be nice if you could ctrl-f the entire repo and check the spelling of 'visualiser'

martin-henz (approved):
Looks fine!

VimuthM:
> * Bug with Hotkeys. To reproduce: call `draw_data` **once**, and press the left or right arrow keys with the data visualizer tab in focus (the app shld break)
@chownces, having some difficulty reproducing this right now. A single call to `draw_data` shouldn't show the left right arrows in the first place...

chownces (approved):

PR #1748, opened 2021-05-25
Upgrade Sourceror for REPL support

btzy:
### Description

Sourceror 0.8.0 now supports REPL.  This PR makes minor changes to cadet-frontend to integrate with Sourceror.

Most things you can do in the editor can now also be done in the REPL.  However:
- You have to run the editor once before you can use the REPL (if not, a descriptive error will be presented).
- You cannot write import declarations in the REPL.

Essentially, we save the linear memory and globals of the original WebAssembly instance after it has finished running, and when invoking the REPL we load the saved linear memory and globals into the new instance.  In order to ensure "compatibility" (i.e. the new instance can use the old memory and globals), we can't add additional imports, and the code from the REPL cannot use statically allocated memory (which is okay, but leads to slightly slower string constants and the lack of location information when the new code triggers a runtime error).

### Type of change

### How to test

Open the playground and set the language to Source §1 WebAssembly, and play around with the REPL.

image

### Checklist

- [x] I have tested this code

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "redux-mock-store": "^1.5.4",
     "redux-saga": "^1.1.3",
     "showdown": "^1.9.1",
-    "sourceror": "^0.7.3",
+    "sourceror": "^0.8.0",
     "typesafe-actions": "^5.1.0",
     "uuid": "^8.3.2",
     "xml2js": "^0.4.23",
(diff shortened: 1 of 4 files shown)


angelsl:
Seems like CloudFlare Pages doesn't create deploy previews on pull requests from forks.

That sucks...

angelsl (approved):

PR #1754, opened 2021-05-31
Fix issue with WorkspaceSaga tests

jonas-chow:
### Description

Event handling in WorkspaceSaga no longer results in console error when running tests.
The problem was that any explicit reference to the store caused some error to be thrown, fixed by using a saga instead of a normal function and putting an action instead.

### Type of change

### How to test

Double check that the test doesn't have any errors. Also might be worth checking that events still get registered properly, and that events don't register when not logged in.

### Checklist

- [x] I have tested this code


angelsl:
This happens because the dependency on `store` results in an import loop.

On a side note, you should reset your branch to match upstream. (But no issue for the PR.)

YaleChen299 (approved):
Everything is working now. Thanks!

PR #1755, opened 2021-05-31
Optimise display()

blackening:
### Description

Fix: https://github.com/source-academy/cadet-frontend/issues/1494 

Please pick either to use the mutating yet redux breaking version (5x speedup), or the well-behaved version which copies the entire buffer every time a display is performed.

Description of bugfix:

1) Buffering and throttling the redux store updates, because every time the store is updated, entire editor is re-rendered. Rate is hardcoded to 150ms.
   * Note that for correctness, console buffer needs to be dumped before errors.
2) Mutating (push) instead of copying entire buffer at every call.

### Type of change

### How to test

Open up playground, fire up a loop with the display function. 

Previously it would choke on 300, now it runs up to 600k on my computer. 

### Checklist

- [x] I have tested this code


martin-henz:
Great improvement! That's how a console should be. (I'm not qualified to actually review the PR.)

martin-henz:
Looks like we have a merge conflict. Can you take a look @blackening ?

martin-henz:
This PR isn't clear. It says "Please pick either..." but it doesn't say how to do the picking, and who is doing the picking. Can you clarify @blackening ?

angelsl:
@martin-henz He intended for the reviewer to pick.

I will take care of this today, no worries.

martin-henz:
> @martin-henz He intended for the reviewer to pick.
> 
> I will take care of this today, no worries.

ok

angelsl (approved):
Using the version that still obeys the React contract.

This yields a ~16x speedup (likely slightly more, since I did the comparison between the dev. build of this, and the current production build).

We can revisit this again in future if that is still not good enough. But I don't think it's worth the further 5x (so that would be ~80x from the baseline, but...)

PR #1772, opened 2021-06-10
Fix body overflow scroll during navbar dropdown

chownces:
### Description

Fixes the body overflow scroll which appears when the navbar dropdown is opened. Also removes console warning for componentWillMount deprecation.

Tested on desktop Chrome, and IOS Safari

### How to test
To test on Android before merging


YaleChen299 (approved):
tested and all looks good!

PR #1773, opened 2021-06-10
Update most dependencies

chownces:
### Description

Updated most dependencies.

Not updated:
- husky: requires new configuration for pre-push hooks
- konva: huge version update that is breaking
- eslint: throws a `failed to load config "react-app" to extend from` locally
- @types/estree: needs to be upgraded together with version in js-slang


martin-henz (approved):
All working. I'm able to start the frontend with no errors and play around with it. Build also works error-free.

PR #1805, opened 2021-06-18
Interactive-SICP: Mobile bug fixes

samuelfangjw:
### Description

1) Left-align text instead of justify
Justified text is hard to read when there are short lines (common when viewing on mobile) and is generally thought to be a [bad idea](https://designforhackers.com/blog/justify-text-html-css/) on web text. Changed to left align instead.

2) Set default image scale to 100%.
Images on mobile were not scaled properly, requiring users to scroll left and right to see the full image. Images now default to 100% scale. This can be overridden by changing web_scale property in xml files.
Before:
Screenshot 2021-06-18 at 4 24 14 PM

After:
Screenshot 2021-06-18 at 4 25 41 PM

This change affects the desktop version as well for consistency, with images now scaled to 100% width unless otherwise specified.
Screenshot 2021-06-18 at 4 27 05 PM

3) Table of contents on mobile version was slightly off centre, I've changed it to be a bit more centralised, and added a slightly larger gap between the caret and the label so that it is easier to press.

Before: 
Screenshot 2021-06-18 at 4 28 36 PM

After:
Screenshot 2021-06-18 at 4 28 45 PM

### Type of change

### How to test

Ensure changes in description are working correctly.

### Checklist

- [x] I have tested this code
- [ ] ~I have updated the documentation~

### DIFF
--- a/src/features/sicp/parser/ParseJson.tsx
+++ b/src/features/sicp/parser/ParseJson.tsx
@@ -130,7 +130,7 @@ const handleImage = (obj: JsonType, refs: React.MutableRefObject<{}>) => {
       <img
         src={Constants.interactiveSicpDataUrl + obj['src']}
         alt={obj['id']}
-        width={obj['scale'] || "100%"}
+        width={obj['scale'] || '100%'}
       />
     );
   } else if (obj['snippet']) {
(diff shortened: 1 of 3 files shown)


martin-henz (approved):
Good catch! Yes for small displays, the text should be lef-aligned.

PR #1812, opened 2021-06-19
Interactive Sicp Bug Fixes

samuelfangjw:
### Description

For the following screenshots, before is on the left and after is on the right.

1) Indent child nodes of table of contents, to better represent hierarchical structure of table of contents.
Screenshot 2021-06-19 at 8 53 26 PM

2) Added an error boundary to catch errors and display a fallback UI. 

3) Tweak font size for math so that they are the same size as surrounding text.
Screenshot 2021-06-19 at 8 58 11 PM

4) Tweak font size of snippets and removed the border.
Screenshot 2021-06-19 at 9 04 49 PM

5) Change redirect component in table of contents to use useHistory hook instead. Previously, when viewing on GitHub pages, if one goes to a new page via the table of contents the new page will not be added to the browser history (cannot press the back button). This should fix that. 

6) Fix issue where direct link to an element will not work on first page load. Also ensure that we always scroll to the top of the element.

### Type of change

### How to test

Verify changes have been made

### Checklist

- [x] I have tested this code


martin-henz (approved):
looks fine

PR #1832, opened 2021-06-26
SICP JS: Rework program snippets

### LINKED ISSUE #1830: Env viz: wrong line highlighted when code is folded
Screenshot 2021-06-26 at 6 16 45 PM

The blue highlighted line which indicates the current breakpoint does not take into account code that has been folded.

Code used in image:
```
function a() {
    return 1;
}

function b(x) {
    return x + a();
}

b(1);
```

In the example above, the body of function a has been folded by pressing the little arrow to the right of the line number.  After pressing run the env viz correctly draws the current environment at the breakpoint (line 8). However, line 10 is highlighted instead.

samuelfangjw:
### Description

This PR is dependent on source-academy/sicp#545.

1. Fixes #1830.
2. Reworked how snippet dependencies are hidden. Dependencies are now hidden by code folding instead. This solves the environment model not working with dependencies bug and the bug where some program snippets cannot be run when dependencies are hidden because the prepend was evaluated in a different scope to the main body.

Screenshot 2021-06-26 at 8 38 07 PM

In the image above, taken from the third program snippet in section 3.2.3, lines 2 to 15 are hidden. A user can click on the yellow button to reveal dependencies.

Screenshot 2021-06-26 at 8 39 09 PM

With this solution the only way to hide dependencies again is to close and reopen the snippet.

### Type of change

### How to test

1. Ensure correct lines are highlighted when using env visualiser with and without code folding.
2. Ensure snippets with dependencies are hidden correctly and can be revealed by pressing the yellow button. The env viz should also work correctly when dependencies are hidden.

### Checklist

- [x] I have tested this code


martin-henz (approved):
Works fine.

PR #1844, opened 2021-07-02
Fix dart sass division deprecation warnings during compilation

chownces:
### Type of change

### Checklist

- [ ] I have tested this code

### DIFF
--- a/src/styles/_contributors.scss
+++ b/src/styles/_contributors.scss
@@ -100,13 +100,13 @@ div.inPermalink {
     text-align: center;
     vertical-align: top;
     display: inline-block;
-    width: (100% / 5);
-    height: (100% / 5);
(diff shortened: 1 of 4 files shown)


src/styles/_contributors.scss
```diff
@@ -100,13 +100,13 @@ div.inPermalink {
     text-align: center;
     vertical-align: top;
     display: inline-block;
-    width: (100% / 5);
-    height: (100% / 5);
+    width: math.div(100%, 5);
```
samuelfangjw, line 103: Would it be better to change this to 20% and the one below to 33.3% since both sides of the equations are constants?

samuelfangjw (commented):
LGTM! Just a minor comment for you to consider below.

samuelfangjw (approved):

PR #1868, opened 2021-07-23
Fix AssessmentWorkspace syntax highlighting bug

chownces:
### Description

The Source chapter, variant and external library props were not passed into the editor component of AssessmentWorkspace. Probably accidentally left out in the previous refactor a year ago.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
+++ b/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
@@ -757,6 +757,9 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
       ? {
           editorSessionId: '',
           editorValue: props.editorValue!,
+          sourceChapter: question.library.chapter || 4,
+          sourceVariant: 'default' as Variant,
+          externalLibrary: question.library.external.name || 'NONE',
           handleDeclarationNavigate: props.handleDeclarationNavigate,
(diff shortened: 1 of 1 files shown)


src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
```diff
@@ -757,6 +757,9 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
       ? {
           editorSessionId: '',
           editorValue: props.editorValue!,
+          sourceChapter: question.library.chapter || 4,
```
martin-henz, line 760: Hmm, the documentation says that DEPLOYMENT element is used for getting the "chapter"; not sure if this is right.

chownces: If believe the XML parses the Source Chapter from the DEPLOYMENT tag, and ultimately stores it as a jsonb (with the rest of the library details)

During assessment retrieval, this is how the chapter (and rest of the library details) per question are sent to the frontend

can @YaleChen299 double confirm?

YaleChen299: Yes. `interpreter` attribute under `DEPLOYMENT` is parsed as `library.chapter`.

martin-henz (approved):

PR #1870, opened 2021-07-26
Bump js-slang and update infinite loop detector

JoeyChenSmart:
### Description

Bump js-slang to 0.5.12.

The new version of js-slang implements an overhauled infinite loop detector which breaks certain features depending on the old implementation (namely the reporting infinite loop to sentry feature).

The new reporting feature can be turned on/off depending on whether the user has given approval or not.
This approval flag (state.session.experimentApproval in the state) is set to 'false' as default.
A future change should check for the user's given approval status and set this flag accordingly.

Also, as discussed in the meeting, we will only show the 'enhanced' infinite loop errors 50% of the time.

### Type of change

### How to test

Run an infinite loop into the playground. The new error should show up 50% of the time. Refresh the page to 'reroll' for different results.

image

### Checklist

- [x] I have tested this code

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "connected-react-router": "^6.9.1",
     "flexboxgrid": "^6.3.1",
     "flexboxgrid-helpers": "^1.1.3",
-    "js-slang": "^0.5.10",
+    "js-slang": "^0.5.12",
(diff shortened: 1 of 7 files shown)


martin-henz (approved):
Looks fine

PR #1874, opened 2021-07-27
Fix data race with js-slang context when running all testcases

angelsl:
### Description

Fix a data race with the js-slang context when running all testcases (clicking Run with the testcases panel open).

### Type of change

### How to test

Test on staging or production.

### Checklist

- [x] I have tested this code


chownces:
I think this issue is not entirely fixed. The error messages do not appear when running the testcases in a clean context (e.g. immediately after refreshing the page). But if we run the program in the editor before running the testcases, the error messages will still appear when the testcases are executed.

angelsl:
> I think this issue is not entirely fixed. The error messages do not appear when running the testcases in a clean context (e.g. immediately after refreshing the page). But if we run the program in the editor before running the testcases, the error messages will still appear when the testcases are executed.

I hate Redux

martin-henz (approved):
Looks good

PR #1877, opened 2021-07-28
Fix output not appearing in snippets which are not evaluated

### LINKED ISSUE #1876: Missing code/text in section 4.4.1
There is missing code in [section 4.4.1](https://sourceacademy.org/sicpjs/4.4.1#p25).  There should be some more text after 
>For example, the query
`lives_near($x, list("Bitdiddle", "Ben"))` 
results in

samuelfangjw:
### Description

Fixes #1876 

### Type of change

### How to test

For snippets which are not evaluated, output should be visible (if present). For example, the chapter 4.4.1, `lives_near(list, "Reasoner"...` should be visible, as seen in the screenshot below.

Screenshot 2021-07-28 at 12 58 00 PM

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/features/sicp/parser/ParseJson.tsx
+++ b/src/features/sicp/parser/ParseJson.tsx
@@ -123,7 +123,12 @@ const handleSnippet = (obj: JsonType) => {
   if (obj['latex']) {
     return <Pre>{handleLatex(obj['body']!)}</Pre>;
   } else if (typeof obj['eval'] === 'boolean' && !obj['eval']) {
-    return <Pre>{obj['body']}</Pre>;
+    return (
+      <>
+        {obj['body'] && <Pre>{obj['body']}</Pre>}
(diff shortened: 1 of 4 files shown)


martin-henz (changes requested):
I think we should always show the expected output, regardless whether the program is runnable or not.

samuelfangjw:
> I think we should always show the expected output, regardless whether the program is runnable or not.

This is already the case for all runnable snippets that have `<JAVASCRIPT_OUTPUT>`, apologies if my initial comment was misleading. 
Screenshot 2021-07-28 at 1 45 33 PM

I've taken this chance to also italicise `<JAVASCRIPT_OUTPUT>`, as per the specs, as well tweak the styles of code snippets very slightly so that snippets can be displayed without having to scroll horizontally on larger screens.

martin-henz (approved):
all good

PR #1893, opened 2021-08-04
Fix logout routing bug

chownces:
### Description

Since the `/achievements` and `/sourcecast` routes are conditionally rendered by course, upon logout, these routes no longer exist and will 404 instead of bringing the user to the login page.

Using `history.push('/login')` for good measure

### Type of change

### How to test

Logout in `/achievements` or `/sourcecast`. You should be redirected to `/login` instead of `/404`

### Checklist

- [x] I have tested this code


angelsl:
Maybe we should make those routes redirect to /login if not logged in, instead of not rendering those routes at all?

chownces:
addressed by #1903

PR #1910, opened 2021-08-08
Fix context data race and optimise assessment and playground workspaces

angelsl:
### Type of change

### How to test

Test that testcases work in all situations.

### Checklist

- [x] I have tested this code


chownces (approved):
LGTM. Btw i feel we should add a `yield put(actions.clearReplOutput(workspaceLocation))` inside `RUN_ALL_TESTCASES`. Else the repl gets a little weird with consecutive testcase runs

PR #1911, opened 2021-08-08
Remove playground external library button

chownces:
### Description

Cant really remove much atm as sourcecast still uses these library select buttons. Also keeping the libraries in the frontend for now as removing it might be breaking and the semester is starting.

### How to test

Playground should work as per normal

### Checklist

- [x] I have tested this code


angelsl (approved):

angelsl:
CF Pages didn't build a preview, strange..

PR #1919, opened 2021-08-10
Recompute achievement completion status and other achievement UI updates

jonas-chow:
### Description

goalProgress.completed is now recomputed when fetching the goal progress from the backend to avoid the goal from being incorrectly completed when goal was previously completed but target count is increased.

Also allows completion text and description of achievements to be multi-lined.

Achievement milestone levels are increased by 1, to reflect that 36k xp (level 37) is full credit

### Type of change

### How to test

Create a goal and an achievement that uses that goal. 
Complete the goal.
Increase the goal's target count so that you would not complete the goal.
Return the the achievement dashboard and verify that the achievement and goal are not completed.

Check that your level starts at 1, and goes up by 1 every 1000 xp

### Checklist

- [x] I have tested this code


chownces (approved):
Thanks for the updates!

PR #1920, opened 2021-08-12
Fix infinite loop reporting not sending coinflip

JoeyChenSmart:
### Description

Now sends the 'experimentCoinflip' flag to Sentry when reporting infinite loops.

Also removed a test case that fails occasionally for unknown reasons (the test case intentionally causes a stack overflow which is sometimes not thrown)

### Type of change

### How to test

 2. Set `approval` to `true` in the infinite loop reporting part of WorkspaceSaga.ts (~line793)
 3. Run any program that will cause an infinite loop. Example: function f(x){f(x);}f(1);
 4. Check that the issue "Error: Infinite Loop" is added on the Sentry dashboard

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/sagas/WorkspaceSaga.ts
+++ b/src/commons/sagas/WorkspaceSaga.ts
@@ -790,16 +790,24 @@ export function* evalCode(
 
     // report infinite loops but only for 'vanilla'/default source
     if (context.variant === undefined || context.variant === 'default') {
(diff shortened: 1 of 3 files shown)


chownces:
Not really sure whats up with that testcase. I've poked around more, it seems to always fail on my Macbook, but always passes on my Linux setup. Passes GitHub CI too...

chownces (approved):
lgtm

PR #1929, opened 2021-08-16
Restore original functionality to run-all-testcases button

angelsl:
### Description

Restore the original behaviour of the run button when the autograder tab is active, which is to run the program followed by all testcases.

Also switch to `takeLatest` instead of `takeEvery` to avoid race conditions if two runs are executed concurrently.

### Type of change

### How to test

Click on "Run" while the autograder tab is active on a question which has no testcases.

Program should still run as usual.

### Checklist

- [x] I have tested this code


angelsl:
I set up the deploy preview to work with the staging backend, by the way.

Give it a try: https://c351664b.cadet-frontend.pages.dev/courses/18/missions/188/0

chownces (approved):
lgtm!

PR #1932, opened 2021-08-18
Fix mcq button highlight when no solution is sent from backend

chownces:
### Description

Previously, when no solution was sent from the backend, any selection will be highlighted red.

Now, they are highlighted in a neutral grey colour:

image

### Type of change

### How to test

- Test against an MCQ question with 'showsolution' = false. Selecting any option should highlight it grey instead of red.

### Checklist

- [x] I have tested this code


martin-henz (approved):
looks good

PR #1955, opened 2021-09-03
Send infinite loop related programs to Sentry

JoeyChenSmart:
### Description

We would like to know whether a student has successfully 'fixed' their infinite loop.
To find out, after an infinite loop is encountered, we will send all programs the student runs to Sentry.
Created a flag in `state.session` called `hadPreviousInfiniteLoop` to keep track of this.

Also created another flag `state.session.sessionId` which will be set to the system time when it is initialized.

### Type of change

### How to test

1. Set approval to true in the infinite loop reporting part of WorkspaceSaga.ts (~line793)
1. Run any program that will cause an infinite loop. Example: `function f(x){f(x);}f(1);`
1. Run any non-infinite looping program. Example: `1+1;`
1. Check that the issue "Error: Non Infinite Loop" is added on the Sentry dashboard

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/application/ApplicationTypes.ts
+++ b/src/commons/application/ApplicationTypes.ts
@@ -301,6 +301,8 @@ export const defaultSession: SessionState = {
   assessmentOverviews: undefined,
   agreedToResearch: undefined,
   experimentCoinflip: Math.random() < 0.5,
+  hadPreviousInfiniteLoop: false,
+  sessionId: Date.now(),
   githubOctokitObject: { octokit: undefined },
   gradingOverviews: undefined,
   gradings: new Map<number, Grading>(),
--- a/src/commons/application/actions/SessionActions.ts
+++ b/src/commons/application/actions/SessionActions.ts
@@ -60,6 +60,7 @@ import {
   UPDATE_COURSE_RESEARCH_AGREEMENT,
   UPDATE_GRADING,
   UPDATE_GRADING_OVERVIEWS,
+  UPDATE_INFINITE_LOOP_ENCOUNTERED,
   UPDATE_LATEST_VIEWED_COURSE,
   UPDATE_NOTIFICATIONS,
   UPDATE_USER_ROLE,
@@ -232,3 +233,5 @@ export const deleteUserCourseRegistration = (courseRegId: number) =>
 
 export const updateCourseResearchAgreement = (agreedToResearch: boolean) =>
   action(UPDATE_COURSE_RESEARCH_AGREEMENT, { agreedToResearch });
+
+export const updateInfiniteLoopEncountered = () => action(UPDATE_INFINITE_LOOP_ENCOUNTERED);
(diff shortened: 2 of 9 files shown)


chownces (commented):
Hi, if im not wrong the `hadPreviousInfiniteLoop` boolean is not reset when changing assessment questions, or after reporting a non error program. This might result in these 'unrelated' programs being sent to Sentry. Should we fix this or is it ok to leave it as it is?

JoeyChenSmart:
> Hi, if im not wrong the `hadPreviousInfiniteLoop` boolean is not reset when changing assessment questions, or after reporting a non error program. This might result in these 'unrelated' programs being sent to Sentry. Should we fix this or is it ok to leave it as it is?

Yeah, I'll filter those 'unrelated' programs out when processing the data.

I also thought of reporting non error programs only when in the same assessment question (after encountering an infinite loop), but we'll still get 'unrelated' programs (if the student rewrites the code from scratch etc), and the logic is more complicated.

angelsl:
@JoeyChenSmart Is it not sufficient to just log the first non-infinite loop program run after an infinite loop?

Because if you leave it like this, that means once someone hits an infinite loop it's going to send _every subsequent run_ until they refresh or close that instance of the frontend.

That could add up to a lot of runs, and we don't have an infinite Sentry quota (although 5 million is a lot, but still...).

JoeyChenSmart:
Yeah actually after thinking about this for a while I think it should be sufficient to log the first non-infinite loop (if the student submits 'unrelated' code we can just assume he gave up, at least for the moment). I'll make the changes and push again.

martin-henz:
Yes, we need all subsequent runs after the infinite loop occurs. 

In the future we might collect all student programs, or even all keystrokes by students, but for Joey's work, it is sufficient to record the programs in the same session after an infinite loop.

JoeyChenSmart:
I was actually thinking measuring whether or not the student fixes the infinite loop in a single 'sitting' (i.e. immediately after getting the error, without taking a break or trying a different problem halfway) might be a better measure of how well the improved error messages are helping the students.
Since it will be more likely that the student is actually using the error message to debug their program (if they run 'unrelated' code, the error message will not be showing anymore).

Using this measure we can get by with recording only the first program after the infinite loop.

angelsl:
We can collect all programs run, I guess, I'm just a bit worried about hitting Sentry quotas.

Sentry is mainly designed for error tracking, rather than tracing events like that, although it certainly works. So if we want to capture all programs, I think we should spin up our own backend, whether as part of the Elixir backend or a separate microservice (like the cadet-logger).

JoeyChenSmart:
Currently we get about ~1000 infinite loops per week, so using the first approach (collecting all programs in the same session as an infinite loop, but not all programs), the amount of data sounds manageable.

The rate of infinite loops will likely increase though, last year we had about 50 000 total.

If this is fine I'll revert the previous commit to the first approach.

angelsl:
> Currently we get about ~1000 infinite loops per week, so using the first approach (collecting all programs in the same session as an infinite loop, but not all programs), the amount of data sounds manageable.

Hmm. OK then. I mean, we are on an open-source sponsored plan, so our quota is pretty high, but.. I have to say again that Sentry wasn't meant for general telemetry collection like this.

If we decide to collect _all_ programs, or even keystrokes, we cannot use Sentry for that.

chownces (approved):
Let's settle with this arrangement for the timebeing then.

PR #1957, opened 2021-09-07
Update step size in grading editor

jiayushe:
### Description

- Use `50` as step size in grading editor, since most marking schemes have `50` as the minimum breakdown, so it is easier to grade by clicking.
- This does not affect manual entering, so xps that are not multiples of `50` can still be given.

### Type of change

### Checklist

- [x] I have tested this code


chownces (approved):
Thank you!

PR #1982, opened 2021-10-15
EnvViz: Display anonymous functions

### LINKED ISSUE #1973: EnvViz: anonymous function objects are not displayed
https://share.sourceacademy.nus.edu.sg/02vj2

function curry(f) {
return x => y => f(x, y);
 }
 curry(math_pow)(3)(4);
 undefined;

Screenshot 2021-10-09 at 9 55 33 AM

However, in the lectures, we also show the function objects that get created: 

Screenshot 2021-10-09 at 10 00 57 AM

A "complete" solution would also need to show all pairs that get created during the program run, regardless whether they are referred to by names or not.

CrownKira:
### Description

Depends on https://github.com/source-academy/js-slang/pull/1147
Fixes #1973 

Previously:
Screenshot 2021-10-16 at 1 22 18 AM
Now:
Screenshot 2021-10-16 at 1 18 40 AM

### Type of change

### How to test

```
function curry(f) {
    return x => y => f(x, y);
}
curry(math_pow)(3)(4);
debugger;
```

### Checklist

- [x] I have tested this code


martin-henz:
the test case above works, but this test case does not work:

Screenshot 2021-10-31 at 10 56 40 PM

(I've bumped js-slang to 0.5.21; please update this PR accordingly.)

CrownKira:
> the test case above works, but this test case does not work:
> 
> Screenshot 2021-10-31 at 10 56 40 PM
> 
> (I've bumped js-slang to 0.5.21; please update this PR accordingly.)

Bumped! 
The issue seems to be caused by js-slang. Perhaps we should create an issue there?

CrownKira:
```
const x = 5; 
x=>x;
{
    u=>u;
}
debugger;

```
This is another bug related to the previous one. The problem seems to be related to js-slang context generator not creating a frame if it only contains dummy bindings. (ie. frames that don't contain any name-to-value reference)

martin-henz:
In this case

139589606-cffbff52-2911-4571-b86f-11e2d3b7978f

there should be no frame. The environment model does not have empty frames. Instead, the function object should have the global environment as its environment:

139589606-cffbff52-2911-4571-b86f-11e2d3b7978f

martin-henz:
I'm merging this PR, because it's an obvious improvement over the current system. I'll create a new issue.

martin-henz (approved):
Looks fine. There is a remaining issue that will be filed in js-slang.

PR #1991, opened 2021-10-19
Upgrading array-move

martin-henz:
### Description

bumping array-move instead of
https://github.com/source-academy/frontend/pull/1925


chownces:
This is not updated yet. Requires some tweaks to our jest config due to lack of CommonJS support in this module.

Alternatively, we can just rewrite the function ourselves. Will just be a few liner.

martin-henz:
Ah, I see. Sorry. Which way should we go with this?

chownces:
Just looked through the underlying implementation, v3.0.1 is essentially the same as v4, the only difference being the renaming of some variables.

I guess we can keep the current version then.

PR #2010, opened 2021-11-07
Fix autocomplete in AssessmentWorkspace

chownces:
### Description

Previously, the AssessmentWorkspace and GradingWorkspace autocomplete were broken as the Variant received from the backend was `null` instead of `default`. As a result, `chapterName` became something like `3_null` and the `builtinSuggestions` became an empty array as seen below:

```
      // line 135 WorkspaceSaga.ts
      let chapterName = context.chapter.toString();
      if (context.variant !== 'default') {
        chapterName += '_' + context.variant;
      }

      const builtinSuggestions = Documentation.builtins[chapterName] || [];
```

### Type of change

### How to test

Test that the default chapter builtins show up in the autocomplete in AssessmentWorkspace and GradingWorkspace

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/sagas/RequestsSaga.ts
+++ b/src/commons/sagas/RequestsSaga.ts
@@ -538,6 +538,7 @@ export const getAssessment = async (id: number, tokens: Tokens): Promise<Assessm
       question.prepend = question.prepend || '';
       question.postpend = question.postpend || '';
       question.testcases = question.testcases || [];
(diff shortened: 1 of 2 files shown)


angelsl:
Oops, didn't see this.. I'll use my fix instead, since it also deals with `null` in `execTimeMs`

PR #2036, opened 2021-12-16
Streamline playground-only installation config

### LINKED ISSUE #1994: Developer-friendly getting-started
Currently the README says in: https://github.com/source-academy/frontend#installation-of-source-academy-latest-version-available-in-github-pages

the following:
```
Copy the .env.example file as .env and set the following variables:

REACT_APP_USE_BACKEND=FALSE
REACT_APP_PLAYGROUND_ONLY=TRUE
You may want to refer to Setting up your environment below for more configuration options, if you wish to set up e.g. the GitHub or Google Drive integrations.
```
If we make these two setting the default, we can get rid of item 4 in the vanilla setup instructions: no need to deal with .env as a beginner.

chownces:
### Description

Closes #1994 

### Type of change

### How to test

`yarn start` without any `.env` file. This should load the playground-only configuration

### Checklist

- [x] I have tested this code

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,17 +21,12 @@ The Source Academy (<https://sourceacademy.org/>) is an immersive online experie
 1. Install a stable version of NodeJS. The active LTS or current version should work fine.
 2. Clone this repository and navigate to it using "cd" in your command line or shell tool.
 3. Run `yarn install` to install dependencies.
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Works perfectly: Checked out, then yarn, then yarn start. That's a smooth onboarding!

PR #2041, opened 2021-12-22
Native Javascript Support

shenyih0ng:
Goal:  **Add ability to execute Javascript programs in Source Academy frontend**

> This is a bare-bones implementation, features such as SICP function integration and security concerns will be addressed in future PRs. Unit testing will also be added once SICP functions are fully integrated in the upcoming PR

Reference:
-  [Sprint #1](https://github.com/orgs/source-academy/teams/js-slang-22-23/discussions/1)
-  [Sprint #2](https://github.com/orgs/source-academy/teams/js-slang-22-23/discussions/4)

### Identifying `nativeJS`

- In order to identify `nativeJS`, we reused the `SourceLanguage` type as implementing a separate interface to handle "non-source" language might cause too much unnecessary complexity as of now.
    - Setting the `chapter === -1` will be a indicator for `nativeJS`

    https://github.com/source-academy/frontend/blob/059a6d3e7953764fcb2bca442128adf4f6abba48/src/commons/application/ApplicationTypes.ts#L128-L132 

### `nativeJS` Evaluation

**We will be using the `eval()` function without any additional parsing or instrumentation to evaluate `nativeJS` program strings**

- Without additional instrumentation, it will be hard for us to catch non-terminating programs (e.g. infinite loops), therefore we have added new UI features and recovery information to guide users.
    - When the program is still under execution, the "Run" button will be in the color green (more on this below)
    - Recovery information is added to the introduction tab

https://user-images.githubusercontent.com/47914880/148048824-cdf0641b-8da2-450c-9596-84a2abce3b6c.mov

- [EvalErrorLocator.ts](https://github.com/source-academy/frontend/blob/d8d917138a9352c8e67d4f591ef2fad0119bd941/src/commons/nativeJS/EvalErrorLocator.ts) In order to provide rich error messages, we have a `ErrorLocator` to help extract line & column numbers by using the stack trace of `Error()` objects (`Error.stack`)

https://user-images.githubusercontent.com/47914880/148048970-f4f29288-2b93-44f9-8d52-210498c64f6d.mov

### Syntax Highlighting

- Rules are extended from [`source` syntax highlighting rules](https://github.com/source-academy/js-slang/blob/4005580cf4b62bffe922045260b47313a9f7f37e/src/editors/ace/modes/source.ts).
    - Main changes would include allowing javascript keywords and operators
    - Color scheme is inherited from the source mode

highlighting

---

### Unblocking the UI before running program

We included a generic wrapper to convert the current microtask of running a program (be it `source` or `nativeJS`) to a task by using `setTimeout`. This will allow the UI to have the opp. to render and display the indicator that a program is currently running.

https://github.com/source-academy/frontend/blob/d8d917138a9352c8e67d4f591ef2fad0119bd941/src/commons/utils/RunHelper.ts#L13-L18

```javascript
// Example usage
const nativeJSResult: NativeJSEvalResult = yield call(runWrapper, evalNativeJSProgram, program);
const sourceResult = call(runWrapper, runInContext, code, context, {
                                            scheduler: 'preemptive',
                                            originalMaxExecTime: execTime,
                                            stepLimit: stepLimit,
                                            throwInfiniteLoops: throwInfiniteLoops,
                                            useSubst: substActiveAndCorrectChapter})
```

_We only used this wrapper for `nativeJS` evaluation and have not integrated it with the execution of source programs_

> However, this still does not allow UI to be updated or users to interact with UI in the process of running a program.


salty-flower:
Is making commits to this draft the recommended approach for me to make (possibly) any contributions? Or shall I open up a separate pull request 😂

shenyih0ng:
> Is making commits to this draft the recommended approach for me to make (possibly) any contributions? Or shall I open up a separate pull request 😂

I think pushing to the current branch would be better? If not we will have 2 pull requests implementing the same feature 😅

martin-henz:
Yes, that's right. You both should push to the same branch.

salty-flower:
Got it :)

salty-flower:
Goal:  **Add ability to execute Javascript programs in Source Academy frontend**
Refrence: [Sprint #2](https://github.com/orgs/source-academy/teams/js-slang-22-23/discussions/4)

### TODOs

- [ ] Unit test

## Implementation Details
- **[Frontend]** Improve error message: achieved by parsing `Error.stack`.
  - adding a error locator
    https://github.com/source-academy/frontend/blob/d80af35d2e01664ad5d49f74a0b42837a7c4d4b0/src/commons/sagas/NativeJsSaga.ts#L44-L86
  - pass the full error object
    https://github.com/source-academy/frontend/blob/d80af35d2e01664ad5d49f74a0b42837a7c4d4b0/src/commons/nativeJS/NativeJSEval.ts#L8

2. **[UX]** Recovery info display
   https://github.com/source-academy/frontend/blob/d80af35d2e01664ad5d49f74a0b42837a7c4d4b0/src/commons/controlBar/ControlBarAutorunButtons.tsx#L60-L67
    https://github.com/source-academy/frontend/blob/d80af35d2e01664ad5d49f74a0b42837a7c4d4b0/src/commons/controlBar/ControlBarAutorunButtons.tsx#L77-L80
4. **[UI]** Rename and relocate menu items
5. **[UI]** Change the phrasing of the text on the right to cater to full JavaScript mode
6. **[UI]** Syntax highlighting: [achieved](https://github.com/source-academy/frontend/commit/c8a70dcbd37cab98be69e17e2fde9c8a69b1fd3a#diff-6541a08dfb3bcb401b1550bb5aa6c5aa93d2cbaa8a5b6e2d67b85a71dec8c6bc) by first copying and then modifying `source.js` from `js-slang`.

## Demo
error displaying
highlighting

> this is just a basic demo, this is not intended to show that it works (more testing is definitely needed)

src/commons/sagas/NativeJsSaga.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    // Notify workspace & clear REPL
+    yield put(actions.updateWorkspace(workspace, { isRunning: true }));
+    yield put(actions.clearReplOutput(workspace));
+    const context: Context = yield select(
+      (state: OverallState) => state.workspaces[workspace].context
+    );
+    context.errors = [];
+
+    async function evalWrapper() {
```
shenyih0ng, line 33: I think it would be good if we can make this a generic wrapper so that we can also wrap `runInContext` as well because currently, long-running source programs would also block UI and not display an indicator (for source programs, I believe it is the "Stop" & "Pause" buttons) that it is running

Perhaps a generic wrapper could be something like this:

```typescript
export async function runWrapper<ReturnType>(
  runFunc: (...args: any[]) => ReturnType,
  ...runFuncArgs: any[]
): Promise<ReturnType> {
  return await new Promise((resolve, _) => setTimeout(() => resolve(runFunc(...runFuncArgs)), 0));
}

// Example usage
const nativeJSResult: NativeJSEvalResult = yield call(runWrapper, evalNativeJSProgram, program);
const sourceResult = call(runWrapper, runInContext, code, context, {
                                            scheduler: 'preemptive',
                                            originalMaxExecTime: execTime,
                                            stepLimit: stepLimit,
                                            throwInfiniteLoops: throwInfiniteLoops,
                                            useSubst: substActiveAndCorrectChapter})
```

What do you think?

src/commons/sagas/NativeJsSaga.ts
```diff
@@ -0,0 +1,95 @@
...
+         * This function uses the non-standard feature `Error.prototype.stack`.
+         * Not working for SyntaxError.
+         *
+         * Some examples:
+         * `@http://localhost:8000/static/js/main.chunk.js line 31925 > eval:1:1
+         *     evalNativeJSProgram@http://localhost:8000/static/js/main.chunk.js:31925:16` --Firefox 91.4.1esr
+         * `TypeError: window.dsl is not a function at eval (eval at evalNativeJSProgram (NativeJSEval.ts:6:1), <anonymous>:1:8)` --Chrome 98.0.4758.9
+         * Seeking better alternative for this.
+         */
+        function getErrorLocation() {
```
shenyih0ng, line 54: Should we put this under `/commons/nativeJS/NativeJSEval.ts` instead, since we will need to account for line number offsets for `eval` preludes? 

Then perhaps we can just return the final `ExceptionError` from `NativeJSEval.ts` also so that everything to do with evaluating and running the program string is under `NativeJSEval.ts` and let `NativeJsSaga.ts` handle the actions?

src/commons/sagas/NativeJsSaga.ts
salty-flower: I agree with this. Making it generic will make more sense.

src/commons/sagas/NativeJsSaga.ts
salty-flower: Yes, this is a good idea. I wasn't thoughtful enough when writing these. Let's move it.

src/commons/nativeJS/EvalErrorLocator.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (evalErrors.length) {
+    // we only want to get the initial eval error
+    // this is to account for nested `eval()` calls
+    const baseEvalError = evalErrors[evalErrors.length - 1];
+    const [lineNumStr, colNumStr] = baseEvalError.slice(1, 3);
+    return {
+      start: { line: parseInt(lineNumStr) - lineOffset, column: parseInt(colNumStr) },
+      end: { line: 0, column: 0 }
+    };
+  }
```
shenyih0ng, line 54: refactored error locator to account for nested `eval()` calls

Example: 
```javascript
// this is within the editor running "full Javascript"
eval("const a = a;")
```

The above snippet would provide the following error stack:

```
Uncaught ReferenceError: Cannot access 'a' before initialization
    at eval (eval at <anonymous> (eval at <anonymous> (playground:1)), <anonymous>:1:11)
    at eval (eval at <anonymous> (playground:1), <anonymous>:1:1)
    at <anonymous>:1:1
```
This means that we will be getting the error line of the inner `eval()` (which I think would be a bit confusing?), therefore  I refactored it to extract all eval errors and get the outermost error. In the case where we would want to also show nested `eval()` we can also extend the function to support it.

> this is of course in the case where we allow user-submitted full JS programs to use `eval()` as well

src/commons/controlBar/ControlBarAutorunButtons.tsx
```diff
@@ -24,22 +24,51 @@ type StateProps = {
...
+  const showDebuggerPause = () =>
+    !props.pauseDisabled &&
+    props.isRunning &&
+    !props.isDebugging &&
+    controlButton('Pause', IconNames.STOP, props.handleDebuggerPause);
+
+  const showDebuggerResume = () =>
+    !props.isRunning &&
+    props.isDebugging &&
+    controlButton('Resume', IconNames.CHEVRON_RIGHT, props.handleDebuggerResume);
```
angelsl, line 59: These don't need to be functions, you can just use the values directly..

src/commons/nativeJS/EvalErrorLocator.ts
```diff
@@ -0,0 +1,86 @@
+import { SourceError } from 'js-slang/dist/types';
```
angelsl, line 1: Aw man, this whole file feels so fragile and hacky..

But there's probably no real better way

angelsl (commented):
No objections from me. Looks pretty good!

martin-henz (approved):
@angelsl reviewed and has "no objections". There is one suggestion that we can take a look in a future PR.

PR #2043, opened 2021-12-29
Update dependencies

angelsl:
Remaining dependencies not bumped:

- konva ^7.2.5 → ^8.3.1 
  - Breaking changes in API; need manual fixing
- react-router-dom ^5.3.0 → ^6.2.1 
  - Breaking changes in API. They have essentially totally removed support for class components, which is difficult for us because we still have many components using `withRouter`.
- eslint: 7 → 8 
  - Needs react-scripts to be updated (CRA 4 needs ESLint 7)
- react-scripts: ^4.0.3 → ^5.0.0 
  - Need to fix Sourceror packaging to work with Webpack 5. Non-trivial.

### DIFF
--- a/.github/workflows/build-development.yml
+++ b/.github/workflows/build-development.yml
@@ -13,7 +13,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v2-beta
       with:
-        node-version: '14'
+        node-version: '16'
     - name: Setup Sentry CLI
       run: |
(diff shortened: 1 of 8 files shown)


angelsl:
I'm removing the Cloudflare Pages deploy preview, because Cloudflare Pages can no longer build our frontend due to our use of gpu.js.

chownces (approved):

Issue #1168, opened 2020-05-26
[Request] Enable alternative bindings on ace editor


JingYenLoh:
Ace editor already supports alternative bindings such as Vim or Emacs modes. This is a request to enable choosing those bindings, so we can edit source code in our preferred way.

martin-henz:
I'm quite interested in this. How would we present this to the users? Quite some time ago, someone floated the idea of a "Settings" menu item, along with "Help" and "About". In those settings, we could specify editor options and runtime limits, for example.

JingYenLoh:
I envisioned a dropdown above the editor, allowing the user to select the editing mode.

![image](https://user-images.githubusercontent.com/22059033/83027029-2a970180-a062-11ea-9dd8-246418693275.png)

However, I am unfamiliar with SA20/21 goals. A Settings page would make sense if cadet-frontend is intending to implement more user customizations. Additionally, keeping the option behind a Settings page would avoid confusion, as such editing modes are unrelated to the rest of the controls, which are all part of the module's learning goals.

Help and About are out of the scope of this request. I'd defer to cadet-frontend maintainers to decide those, in a separate issue.

martin-henz:
Thanks for the mock-up!

I think such a drop-down menu would be much too prominent. Students would typically change their Ace mode three or four times max throughout the entire semester, and the majority of students (I estimate) would not touch the Ace mode at all. That's why I thought a less prominent and more global setting would be preferrable. I think local browser storage would be fine for this. (As opposed to saving this on the server.) This way the setting would also work for casual users of the textbook.

Not sure if you understood my comment about "Help" and "About". I only wanted to suggest a place for the "Settings" option. A good place might be here:
* Settings
* About
* Help
in the dropdown for the button in the right upper corner.

martin-henz:
Regarding the goals: I'm adding "online" to the SA tagline:
 Immersive **online** experiential environment for learning programming

The editor plays a central role in this aspiration, so your comments and suggestions are very much appreciated!

martin-henz:
<img width="430" alt="Screenshot 2020-06-13 at 4 08 33 PM" src="https://user-images.githubusercontent.com/30220768/84563748-32f47980-ad90-11ea-82f2-39648e925c5f.png">

Things are moving fast. We already have a profile page!

We could have the following items as "Settings" in the profile page:

* Ace keyboard shortcuts, with the selection "default", "emacs" and "vi" (or whatever Ace allows)
[...]

PR #2055, opened 2022-01-06
Add Ace 'editor bindings' setting

chownces:
### Description

* Added a settings dialog box under the top right dropdown
* Implemented the Editor Bindings toggle with browser local storage
    * Design decision to do it this way without the backend so that non Source Academy @ NUS users can use this feature too
    * Setting should persist until the user clears his local storage

Refer to #1168 

image
image

### Type of change

### How to test

- [ ] Check that the toggle is working in the various workspaces (e.g. playground, missions, quests, paths, etc.), and persists when switching between different workspaces
- [ ] Check that the setting persists through a browser refresh

### Checklist

- [x] I have tested this code


src/commons/utils/Hooks.ts
```diff
@@ -77,3 +77,26 @@ export function useInput<T>(defaultValue: T) {
...
+ * When calling this hook, the value will take on the stored value in local storage (if any).
+ * If this key-value does not exist in local storage yet, the default value will be used.
+ */
+export function useLocalStorageState<T>(
+  key: string,
+  defaultValue: T
+): [T, React.Dispatch<React.SetStateAction<T>>] {
+  const [value, setValue] = React.useState<T>(() => {
+    const localStorageValue = window.localStorage.getItem(key);
+    return localStorageValue ? JSON.parse(localStorageValue) : defaultValue;
```
samuelfangjw, line 94: Consider catching any parsing errors (in case the local storage value is modified)

martin-henz: I suppose a parse error would only arise if the local storage were changed manually, somehow. Maybe it's not so important to worry about such a case.

PR #2061, opened 2022-01-07
fullJS security enhancements 

### LINKED ISSUE #2058: `fullJS` user disclaimer
- Upon switching to `fullJS` mode, there should be an indicator (possibly a modal) 
- This is to warn users that they are using an unsafe feature of sourceacademy

### LINKED ISSUE #2059: Disable link sharing for `fullJS`
- users should not be allowed to share fullJS code snippets via "Share" button
- disable loading links with `chap=fullJS` in URL query string

### LINKED ISSUE #2062: Restrict `fullJS` mode to public site (gh pages)

salty-flower:
- Fixes #2058 
- Fixes #2059 
- Fixes #2062  

### Description
As the title and previous issues suggest, this PR adds a disclaimer when user tries to switch to `fullJS` mode. Inside `fullJS`, there will be no `Share` button provided, and any link that includes `chap=-1` will be rejected with a warning.

### Demo
disclaimer
warning
source buttons
fullJS buttons

### DIFF
--- a/src/pages/playground/Playground.tsx
+++ b/src/pages/playground/Playground.tsx
@@ -12,7 +12,7 @@ import { HotKeys } from 'react-hotkeys';
 import { useSelector } from 'react-redux';
 import { useMediaQuery } from 'react-responsive';
 import { RouteComponentProps } from 'react-router';
-import { showWarningMessage } from 'src/commons/utils/NotificationsHelper';
+import { showNativeJSWarningOnUrlLoad } from 'src/commons/nativeJS/NativeJSUtils';
 
 import {
   InterpreterOutput,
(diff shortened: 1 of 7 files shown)


salty-flower:
#2062

src/pages/playground/Playground.tsx
```diff
@@ -138,27 +139,30 @@ export type StateProps = {
...
+      props.handleChapterSelect(chapter, variant);
+    }
 
-  const execTime = Math.max(stringParamToInt(qs.exec || '1000') || 1000, 1000);
-  if (execTime) {
-    props.handleChangeExecTime(execTime);
+    const execTime = Math.max(stringParamToInt(qs.exec || '1000') || 1000, 1000);
+    if (execTime) {
+      props.handleChangeExecTime(execTime);
+    }
```
shenyih0ng, line 165: I am not quite sure if it would be better UX if we should still load the program string in the default chapter selected when it is intended for `fullJS`? I am 50/50 on this, what do you think?

salty-flower: Hmm then it would be too easy to execute shared program, since users can simply switch to `fullJS` later... I am not sure either.

martin-henz: I'm voting for the cleanest, simplest solution, which I think is the one you have implemented.

martin-henz (approved):
Looks good!

PR #2075, opened 2022-01-24
Envviz/curve edge corners

ngxingyu:
### Description
Implemented rounded corners for arrows in Environment Visualizer diagram replacing sharp corners for greater clarity on direction of lines (turning left or right) particularly when multiple lines merge. Also looks better.

This implementation takes in the existing set of calculated points and computes the position of the control and endpoints of the quadratic curve (rounded corners) based on the lengths of the two adjacent lines and the original corner position. Below is an example of the rounded corners.

original
rounded

### Type of change

### How to test

Ran several sample programs and ensured that all arrows are pointing from / to the same locations, with the curved corners being displayed as expected

### Checklist

- [x] I have tested this code


martin-henz (approved):
That's a very nice improvement. I like the attention to detail as it comes to the calculation of the radius. Experience will tell whether this works all the time, and there might be ways to further improve. For now: both thumbs up!

Issue #2097, opened 2022-02-09
REPL: currently broken


martin-henz:
The REPL does not recognize declared names, even after running the program:
<img width="798" alt="Screen Shot 2022-02-09 at 4 48 58 PM" src="https://user-images.githubusercontent.com/30220768/153296159-eefa9c0f-8dbb-4096-aea8-49520a7ae296.png">

martin-henz:
Fixed by #2103

PR #2103, opened 2022-02-25
Fix repl

thomastanck:
### Description

Fixes issue #2097 due to recent changes in NativeJS.

When evaluating editor code, we normally clear the context first to reset the state of all outputs, js-slang internals, etc. to a blank initial state, then run the editor code in the fresh context.

The problem was that the context was being fetched before it was cleared. Most likely this meant that the context that the editor code was being evaluated in was a stale, now inaccessible, context, while the repl code was being evaluated in the fresh context that does not have the results of the editor code stored within.

### Type of change

### How to test

Run `const x = 1;` in editor, then run `x;` in REPL.

Screenshot

### Checklist

- [x] I have tested this code
- [ ] <s>I have updated the documentation</s>


chownces (approved):
Looks good!

salty-flower (approved):
The `context` was retrieved ahead due to the need to check for `fullJS` chapter and was not refreshed after if-else. Yes, thanks for spotting the error - the order of statements matters so much!

shenyih0ng:
Should we include a test case to ensure the behaviour stays the same in the future?

martin-henz:
I'm merging this because I have a demo coming up on Wednesday and want the REPL to work. I'll create an issue that we need a regression test for the REPL.

martin-henz:
fixes #2097

PR #2119, opened 2022-03-07
Envviz/new layout

ngxingyu:
### Description
Implement new layout and features for environment model diagram for better readability and ease of use. (Refer to wiki for more details)
- Prevented collisions of arrows with frames, placed all arrays on separate layer from frames and reduced distance from arrays to the objects referencing it.
- Positioned frames on a grid based on its depth from the global frame and the order in which the frame is created.
- Added support for huge diagrams based on [HTML5 Large Canvas Scrolling Demo](https://konvajs.org/docs/sandbox/Canvas_Scrolling.html#4-Emulate-screen-moving-with-transform)
- Added support for exporting to jpg (changing the mode of the diagram will set background to white and expand all tooltips, and click save to save the entire diagram as images of width < 25000px)
- Added tests for the placement of function objects and some of the lines.

### Type of change

### How to test

- [x] Check that the large diagrams can render
- [x] Printable mode should render the same diagram with white background and tooltips expanded with the function parameters and the start of the function body.
- [x] Check that the only lines passing through frames are lines originating from the same frame.
- [x] Check that hovering over a frame will highlight all arrows emerging from it and the array object it’s pointing to
- [x] Hovering over array will highlight all of its arrays units 
- [x] https://share.sourceacademy.nus.edu.sg/qb0ch
  - Arrows pointing to arrays within the same array level should be staggered horizontally
 
 0 | 1
:-------------------------:|:-------------------------:
diagram_0_0 | diagram_1_0

- [x] https://share.sourceacademy.nus.edu.sg/7m3j0
  - Within the same frame level, arrows from frames pointing to different parent frames should be on a different lane. (Middle right of the diagram)
diagram_0_0

- [x] https://share.sourceacademy.nus.edu.sg/jkrdv 
  - Example of arrows between arrays at different levels, arrows from arrays to fn objects, arrays to itself.
diagram_0_0

### Checklist
- [x] I have tested this code


CrownKira:
Hi, senior here, really appreciate the immense effort you guys have put into the env vis! Great job :D
I do not have time to look through the codebase since it is too huge, but I have played around with it for a while and here are my comments: 

- Some arrows still overlap
Screenshot 2022-03-14 at 5 16 53 PM

- There is a considerable drop in the responsiveness of user interaction. eg. hovering, scrolling; (Perhaps revise your collision detection algorithm or the grid rendering system. Try to memorize to prevent unnecessary computations or rerenders.)
- Having array structure on another layer is a good idea but it unnecessarily complicates things. (Instead of isolating the array layer, is going around the frame a possibility?) 
Comparison: 
Before: 
Screenshot 2022-03-14 at 5 14 51 PM
Screenshot 2022-03-14 at 5 40 30 PM

After: (As you can see, the space usage is not really optimised) 
Screenshot 2022-03-14 at 5 16 53 PM
(is there really a need for the arrow to take four turns just to get to the array?)
Screenshot 2022-03-14 at 5 36 34 PM

- Highlighting on hover is not working as expected (.. and I don't think it is a good idea to highlight the entire structure since it can get pretty costly thus compromising response time.) 
(There is a cutoff somewhere) 
Screenshot 2022-03-14 at 5 20 26 PM

Code snippets used for testing:
```
const fn = () => "L";
const x = ["long string", pair(() => 1, () => 2), list(1, pair(2, 3), () => 3), () => "THIS", 5, 6];
const y = list(x[1], x[2], tail(x[1]), tail(x[2]), fn);

debugger;
```
```
const x = [1,2,3];
debugger;
```

ngxingyu:
Thanks for your comments!
> Some arrows still overlap
- For arrows moving beside / above / below frames, we chose to allow some to overlap if there are more than 8 arrows in that lane so it would fit between the frames. For arrows moving horizontally between arrays, it's a current limitation which isn't handled in this PR but users can hover/click on the arrows to check where the arrow points to for now.

> There is a considerable drop in the responsiveness of user interaction. eg. hovering, scrolling; (Perhaps revise your collision detection algorithm or the grid rendering system. Try to memorize to prevent unnecessary computations or rerenders.) 
- I believe the responsiveness is mainly affected by the large number of event listeners in the entire diagram, we added onClick listeners to most of the objects, so it would probably be 1.5 times as laggy as the existing version. I've removed the highlighting of the entire array structure so it might be slightly faster.  Another possible reason it might be slower when scrolling might be due to the pixels out of screen needing to be rendered due to the large canvas implementation, but that should be negligible, the lag is likely due to the large number of event listeners konva has to keep track of especially for large diagrams, disabling the listeners outside of the screen might help, but in cases where there are many arrows in the screen, it would probably still lag.

> Having array structure on another layer is a good idea but it unnecessarily complicates things. (Instead of isolating the array layer, is going around the frame a possibility?)
- placing the array next to one of the frames referencing it might reduce the height of the diagram by the total height of all other bindings, but since the width of the diagram would be much greater than the height in most programs, we felt it's more important to optimize the width for wider diagrams
- placing the arrays beside the frame might be better in cases where the array is only referenced by bindings from one frame throughout the entire program execution as seen in the example above, but less so when the same array is referenced in multiple frames. Since array objects aren't associated with a particular frame and can move from one frame to another, we decided to just put the arrays on another layer so the frames and function objects won't shift horizontally during the execution of the program. Might be possible to place arrays referenced by single frame next to the frame while leaving the other arrays as it is currently, but it will take some time to implement.

> (is there really a need for the arrow to take four turns just to get to the array?)
- its possible to do it in two turns but we will have to check if there's any array to the left of that array and having the extra turns is probably not going to affect readability so we chose a simpler solution.

> Highlighting on hover is not working as expected (.. and I don't think it is a good idea to highlight the entire structure since it can get pretty costly thus compromising response time.)
- Agree that it has odd behaviors and might be expensive, so I removed the highlighting of the objects and arrows to and from arrays when hovering / clicking.

src/features/envVisualizer/components/arrows/GenericArrow.tsx
```diff
@@ -48,25 +81,59 @@ export class GenericArrow implements Visible, Hoverable {
...
-      strokeWidth: Number(Config.ArrowStrokeWidth)
-    });
-  };
-
+  }
+  onClick({ currentTarget }: KonvaEventObject<MouseEvent>) {
+    this.selected = !this.selected;
+    if (!this.isSelected()) {
+      if (
+        // !(this.source instanceof ArrayUnit && this.source.parent.isSelected()) &&
```
chownces, line 99: unused code. (lines 118-119 too). please have a glance through the whole PR again. will not add further comments on this as it's getting long

src/commons/mobileWorkspace/mobileSideContent/MobileSideContent.tsx
```diff
@@ -119,7 +128,12 @@ const MobileSideContent: React.FC<MobileSideContentProps & OwnProps> = props =>
   };
 
   const renderedTabs = React.useMemo(() => {
-    const renderTab = (tab: SideContentTab, workspaceLocation?: WorkspaceLocation) => {
+    const renderTab = (
+      tab: SideContentTab,
+      workspaceLocation?: WorkspaceLocation,
```
chownces, line 133: i believe these variables are unused. lets remove them

src/commons/mobileWorkspace/mobileSideContent/MobileSideContent.tsx
```diff
@@ -76,13 +78,20 @@ const MobileSideContent: React.FC<MobileSideContentProps & OwnProps> = props =>
    */
   const renderedPanels = () => {
     // TODO: Fix the CSS of all the panels (e.g. subst_visualizer)
-    const renderPanel = (tab: SideContentTab, workspaceLocation?: WorkspaceLocation) => {
+    const renderPanel = (
+      tab: SideContentTab,
+      workspaceLocation?: WorkspaceLocation,
+      width?: number,
```
chownces, line 84: same for these. `renderPanel` is never called with `width` nor `height`

src/features/envVisualizer/EnvVisualizer.tsx
```diff
@@ -9,15 +9,32 @@ type SetVis = (vis: React.ReactNode) => void;
...
+    EnvVisualizer.printableMode = !EnvVisualizer.printableMode;
+  }
+  public static getPrintableMode(): boolean {
+    return EnvVisualizer.printableMode;
+  }
 
   /** SideContentEnvVis initializes this onMount with the callback function */
-  static init(setVis: SetVis) {
+  static init(setVis: SetVis, width: number, height: number) {
+    Layout.visibleHeight = Math.max(isNaN(height) ? window.innerHeight - 200 : height - 200, 200);
```
chownces, line 23: can we assign names to these constants (e.g. 200, 50)? in general, we should avoid magic numbers/ values

src/features/envVisualizer/EnvVisualizerLayout.tsx
```diff
@@ -30,25 +32,70 @@ export class Layout {
...
+      Layout.stageWidth = Math.min(Layout.width, window.innerWidth);
+      Layout.stageHeight = Math.min(Layout.height, window.innerHeight);
+      Layout.stageRef.current.width(Layout.stageWidth);
+      Layout.stageRef.current.height(Layout.stageHeight);
+      EnvVisualizer.redraw();
+    }
+    Layout.invisiblePaddingVertical = (Layout.stageHeight - Layout.visibleHeight) / 2;
+    Layout.invisiblePaddingHorizontal = (Layout.stageWidth - Layout.visibleWidth) / 2;
+
+    const container: HTMLElement | null = this.scrollContainerRef.current as HTMLDivElement;
```
chownces, line 83: is this necessary? noticed that there is already some inline styling for this container below, which takes on the values dynamically.
(i haven't had the chance to run the code yet 😅)

chownces (changes requested):
Good work on the env visualizer! haven't had the chance to take an indepth look at the PR, but here are some comments in addition to those that karwi mentioned last week

Some additional comments:
1. Would be great if the checkbox text can be aligned better. Should try to avoid double scrolling containers too (see the attached screenshot)

image

2. I haven't had the chance to test this out on mobile, but would be great if the team can take some time to check that it works properly on IOS and Android too (e.g. common browsers such as Safari and Chrome)
    * let me know if you don't have access to these devices, i can help to test this too

src/features/envVisualizer/EnvVisualizerLayout.tsx
```diff
@@ -30,25 +32,70 @@ export class Layout {
...
   static environmentTree: EnvTree;
   /** the global environment */
   static globalEnvNode: EnvTreeNode;
-  /** array of levels, which themselves are arrays of frames */
-  static levels: Level[];
+  /** grid of frames */
+  static grid: Grid;
   /** memoized values */
   static values = new Map<Data, Value>();
+  static oldValues = new Map<Data, Value>();
```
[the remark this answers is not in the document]
ngxingyu: Removed since it has no use

src/features/envVisualizer/EnvVisualizerLayout.tsx
ngxingyu: Currently the dimensions of the div created in `static draw(): React.ReactNode` isn't set dynamically so this updateDimensions method is required to resize the diagram and the scroll container

martin-henz (approved):
All conversations in review by @chownces are resolved

PR #2134, opened 2022-04-07
Adding mission comments to achievement view

riccqi:
### Description

Added new component AchievementCommentCard to AchievementView, to display the relevant comments from the corresponding assessment. So that students have a more useful achievement view card. AchievementCommentCardContainer also added to connect with the redux store.

Added 

### Type of change

### Checklist

- [x] I have tested this code


src/commons/achievement/AchievementView.tsx
```diff
@@ -59,7 +75,13 @@ function AchievementView(props: AchievementViewProps) {
           ))}
         </span>
       </div>
-      <AchievementViewGoal goals={goals} />
+
+      {status === AchievementStatus.COMPLETED && assessment && assessment.type === 'Missions' && (
```
chownces, line 79: The frontend should not be hardcoded with "Missions". Perhaps you are looking for `isManuallyGraded` in assessment configs? We have this boolean column in the backend db to track assessment types that are manually graded. This assessment config is fetched when we load a new course/ refresh the page

src/commons/achievement/AchievementCommentCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  const toMission = (questionId: number) => {
+    const listingPath = `/courses/${courseId}/${assessmentTypeLink(assessment?.type)}`;
+    const assessmentWorkspacePath = listingPath + `/${assessment?.id.toString()}`;
+    history.push(assessmentWorkspacePath + `/${questionId}`);
+  };
+
+  return (
+    <div>
+      <h1 style={{ paddingLeft: '2rem' }}>Feedback</h1>
```
chownces, line 23: Let's shift all inline CSS into the .scss file

src/styles/_achievementComments.scss
```diff
@@ -0,0 +1,23 @@
+.mission-comments {
```
chownces, line 1: Can we use a more general term? e.g. assessment-comments. Same for 'mission-button' below

chownces (changes requested):
Good work! However, I have some additional comments:

1. Could we align the question number and comment? (the comment seems to be higher than the question number)
image

2. Could you check with Prof Martin/ Boyd if this is how we want to handle longer comments?
image

3. Could you remove the double scroll containers? (should only have 1)
Screenshot 2022-06-11 at 17 05 25

chownces:
Also, the card does not display multiline comments properly. 

For example, the grader enters this:
image

However, it shows up as this:
image

Perhaps we can consider using:
white-space: pre-wrap;
word-break: break-word;

Would be great if you guys could do the same for the grading tab in the assessment workspace:
image

src/commons/achievement/AchievementView.tsx
riccqi: Would this entail going into the redux store again to find the assessment config that matches the assessment that is currently selected? Since the assessment config is decoupled from the assessment object

src/commons/achievement/AchievementView.tsx
```diff
@@ -59,11 +84,23 @@ function AchievementView(props: AchievementViewProps) {
...
         </span>
       </div>
-      <AchievementViewGoal goals={goals} />
-      {status === AchievementStatus.COMPLETED && (
+
+      {status === AchievementStatus.COMPLETED &&
+        selectedAssessment &&
+        selectedAssessmentConfig &&
+        selectedAssessmentConfig.isManuallyGraded && (
+          <AchievementCommentCard courseId={courseId ?? 0} assessment={selectedAssessment} />
```
chownces, line 92: Let's not default the `courseId` to 0 here as it might cause unintentional bugs in the future. Can we handle this properly if the course is yet to be loaded?

src/pages/achievement/subcomponents/AchievementDashboard.tsx
```diff
@@ -176,7 +179,7 @@ function Dashboard(props: DispatchProps & StateProps) {
           </ul>
 
           <div className="view-container">
-            <AchievementView focusUuid={focusUuid} />
+            <AchievementView courseRegId={courseRegId ?? 0} focusUuid={focusUuid} />
```
chownces, line 182: Same as above

chownces (commented):
image

Is it possible to add a `cursor: pointer` to the 'To Question' button so that users explicitly know they can interact with the button? (coming from the UI/UX perspective)

chownces (approved):

PR #2152, opened 2022-06-21
Upgrade dependencies

chownces:
### Description

Upgraded most dependencies. However, there are several packages with breaking changes that have not been upgraded yet. Will take a look at those in the near future. Appreciate any help! Especially with the react/CRA upgrades.

### How to test

I also migrated BlueprintJS to version 4, so there might be slight variations in the UI. I have generally tested the different pages to check for consistency. Would appreciate if the reviewer could take a look through the UIs as well (including SICP).


martin-henz (approved):
Looked through the changes. All fine. Checked out and ran the frontend with no problems.

PR #2154, opened 2022-07-03
Bump dependencies

angelsl:
Bump dependency minor versions.

Note: I am forcing gl to ^5.0.3 to take a recent bugfix. If/when gpu.js updates, we can remove that forced resolution.


angelsl:
Closing in favour of #2154

PR #2163, opened 2022-07-26
Bump dependencies

chownces:
Bumped dependencies again. Here is the list of remaining dependencies that are yet to be bumped at this moment as they require a closer look:

image


martin-henz (approved):
I can build and start the frontend. Played around a bit. No issues.

Issue #2174, opened 2022-08-11
Sentry blocked by adblockers


JoeyChenSmart:
Seems like some ad-blockers might disable sentry entirely:
![image](https://user-images.githubusercontent.com/5654470/184117147-43cd3767-4078-4290-8354-77f8ae3766b0.png)

Not sure if this is a known issue, or there is some fix in place already.

Looks like there's a quick workaround from Sentry: https://docs.sentry.io/platforms/javascript/troubleshooting/#dealing-with-ad-blockers

martin-henz:
The sentry docs say ad blocker "may cause your application to behave unexpectedly." What does that mean for us? What is the behavior of the system when an add blocker is in place? Is this affecting the functionality of the frontend, or is it just preventing the logging on the server?

JoeyChenSmart:
AFAIK it won't affect any functionality other than the error logging, since the ad-blocker just prevents the frontend from sending requests to sentry's domain.

RichDom2185:
This is a user choice (if they choose to not be tracked) – we can't do anything about it unless we want to self host all the ingest endpoints. Since the app isn't crashing when this happens, I feel we can just close this issue.

PR #2175, opened 2022-08-11
Bump Sentry and update Infinite Loop Detector

JoeyChenSmart:
### Description

1. Bump sentry and remove deprecated sentry code in infinite loop error sentry reporting
2. Remove 'coinflip' in infinite loop detector (now all infinite loops are reported in sentry).

Side note: seems like sentry is blocked by adblockers, see https://github.com/source-academy/frontend/issues/2174
### Type of change

### How to test

1. Set approval to true in the infinite loop reporting part of WorkspaceSaga.ts (~line793)
1. Run any program that will cause an infinite loop. Example: function f(x){f(x);}f(1);
1. The 'infinite loop detected' message should appear in the browser all the time (can make sure by refreshing the page)
1. Check on sentry dashboard that the infinite loop error has been sent. Note that you may have to disable adblock for this to work (can check in developer console if a POST request is blocked, see issue above)

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/utils/InfiniteLoopReporter.ts
+++ b/src/commons/utils/InfiniteLoopReporter.ts
@@ -12,7 +12,6 @@ import {
  */
 function reportInfiniteLoopError(
   sessionId: number,
-  coinflip: boolean,
   errorType: InfiniteLoopErrorType,
   isStream: boolean,
   message: string,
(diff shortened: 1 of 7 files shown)


src/commons/utils/InfiniteLoopReporter.ts
```diff
@@ -78,7 +70,7 @@ function reportPotentialInfiniteLoop(
 function reportNonErrorProgram(sessionId: number, code: string[]) {
   Sentry.withScope(function (scope) {
     scope.clearBreadcrumbs();
-    scope.setLevel(Sentry.Severity.Info);
+    scope.setLevel('info');
```
chownces, line 73: Should we make this a constant in our repo instead?

JoeyChenSmart: I don't think this will be necessary. Seems like this is the 'canonical' way of setting the severity level in sentry (see [their docs](https://docs.sentry.io/platforms/javascript/usage/set-level/)). 

Also, the current (depreciated) `Sentry.Severity.Info` is already a constant enum] ([source](https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/severity.ts)).

chownces (approved):

Issue #2176, opened 2022-08-13
Add support for multiple files & directories


ianyong:
# Motivation

To support the development of larger software systems, the Source Academy needs to support projects with multiple files & directories for better code organisation & readability. This will also help to gently ease new programmers to working on larger-scale projects.

# Tasks

_Subject to change._

* [ ] Implement multiple files.
   * For now, all files will be isolated in the sense that they cannot reference each other & `run` will only run the currently opened file.
   * If there is only a single file, it cannot be deleted.
      * This restriction will be removed once tabs are added to the code editor.
* [ ] Extend local storage persistence to work with multiple files.
* [ ] Add tabs to the code editor.
   * At the same time, this change will allow the code editor to not always have at least one file open, thus removing the restriction of there being at least a single file.
* [ ] Implement directories & extend local storage persistence to work with directories.
* [ ] Implement local imports in js-slang.
   * Currently, [modules](https://github.com/source-academy/modules) only supports importing of the libraries listed [here](https://github.com/source-academy/modules/tree/master/src/bundles).
* [ ] Add run configurations.
   * With multiple files, there is a need to specify the entrypoint of the program.
* [ ] Add ability to toggle multi-file support in playground from Ground Control.
   * This feature should not be available from the start of the course.
* [ ] Implement multiple file support in [sling](https://github.com/source-academy/sling) & [sinter](https://github.com/source-academy/sinter) (robots).

# Extensions

* [ ] Implement backend persistence for multiple files & directories.
   * For missions, quests & contests.
   * Will also need to add a flag to toggle multiple files in the XML configuration.
* [ ] Add ‘Collapse All’ & ‘Expand All’ buttons in the file system sidebar.
* [ ] Rework Google Drive synchronisation to work with projects (which can consist of multiple files) instead of just single files.
   * The equivalent for GitHub will be deprecated if proper version control support is added to Source Academy.

PR #2197, opened 2022-08-28
Move editor out of MobileSideContent into MobileWorkspace

ianyong:
### Description

The following pairs of components are equivalent:
|Desktop View|Mobile View|
|-|-|
|`Workspace`|`MobileWorkspace`|
|`SideContent`|`MobileSideContent`|

However in the desktop view, the editor is a child of the `Workspace` component while in the mobile view, the editor is a child of the `MobileSideContent` component which is in itself a child of the `MobileWorkspace` component. This results in `MobileSideContent` not being truly agnostic of the kinds of tabs that are passed in, resulting in tight coupling.

For consistency, as well as to address the tight coupling as mentioned above, the `Editor` component should be moved out of `MobileSideContent` into `MobileWorkspace`. This was done by allowing the `body` field of `SideContentTab`s to be `null`, indicating that no panel should be rendered for the tab. When that happens, the editor that is underneath will be visible.

This also allows us to remove the forced rendering of the editor in the `changeTabsCallback` of the `MobileSideContent` component.

### Type of change

### How to test

Use developer tools to set the screen resolution to one that triggers the mobile view.

### DIFF
--- a/src/commons/mobileWorkspace/MobileWorkspace.tsx
+++ b/src/commons/mobileWorkspace/MobileWorkspace.tsx
@@ -212,9 +212,10 @@ const MobileWorkspace: React.FC<MobileWorkspaceProps> = props => {
         <ControlBar {...props.mobileSideContentProps.mobileControlBarProps} />
       )}
 
-      <div className="mobile-editor-panel">{createWorkspaceInput()}</div>
-
-      <MobileSideContent {...updatedMobileSideContentProps()} {...draggableReplProps} />
+      <div>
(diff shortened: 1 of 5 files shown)


chownces (changes requested):
Thanks for refactoring this. However, some CSS needs to be updated after this refactoring.

image
image

ianyong:
> Thanks for refactoring this. However, some CSS needs to be updated after this refactoring.
> 
> image
> 
> image

Thanks for catching this! It should be fixed now:

image

chownces (approved):

PR #2206, opened 2022-09-04
Move draggable REPL logic out of MobileSideContent into MobileWorkspace

ianyong:
### Description

The `MobileSideContent` component already takes in an `onChange` function that is called within the `changeTabsCallback`. Rather than handling the logic for displaying & disabling the draggable REPL inside `MobileSideContent`, we can wrap `props.mobileSideContentProps.onChange` in a new function in `MobileWorkspace` that also handles the draggable REPL logic. This new function is then passed into `MobileSideContent` as `onChange`. This helps to decouple `MobileSideContent` from `MobileWorkspace`, making it more extensible for future additions.

Part of the refactoring for https://github.com/source-academy/frontend/issues/2176.

### Type of change

### How to test

Use developer tools to set the screen resolution to one that triggers the mobile view and check that the behaviour is unchanged when switching between tabs.

### DIFF
--- a/src/commons/mobileWorkspace/MobileWorkspace.tsx
+++ b/src/commons/mobileWorkspace/MobileWorkspace.tsx
@@ -160,10 +160,32 @@ const MobileWorkspace: React.FC<MobileWorkspaceProps> = props => {
     document.documentElement.style.setProperty('--mobile-repl-height', '0px');
   };
 
-  const draggableReplProps = {
-    handleShowRepl: handleShowRepl(-300),
-    handleHideRepl: handleHideRepl,
-    disableRepl: setIsDraggableReplDisabled
+  const handleTabChangeForRepl = (newTabId: SideContentType, prevTabId: SideContentType) => {
+    // Evaluate program upon pressing the 'Run' tab on mobile
+    if (
+      (prevTabId === SideContentType.substVisualizer ||
+        prevTabId === SideContentType.autograder ||
+        prevTabId === SideContentType.testcases) &&
+      newTabId === SideContentType.mobileEditorRun
+    ) {
+      props.editorProps?.handleEditorEval();
+    } else if (newTabId === SideContentType.mobileEditorRun) {
+      props.editorProps?.handleEditorEval();
+      handleShowRepl(-300)();
+    } else {
+      handleHideRepl();
+    }
+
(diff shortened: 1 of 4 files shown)


martin-henz:
Can you check how the wiki documents this? Any need for an update there? Or maybe we add a page on this?

ianyong:
> Can you check how the wiki documents this? Any need for an update there? Or maybe we add a page on this?

The most relevant page seems to be [Mobile Development Guide](https://github.com/source-academy/frontend/wiki/Mobile-Development-Guide) which provides a high-level overview of the behaviour of the mobile workspace. There is no need to update that page because the behaviour has not changed at all.

As for whether we should add a page on this, I'm not very sure what that would achieve because my changes so far are to realign the responsibilities of the various components without any changes to behaviour. The `MobileSideContent` component should be very clear on its purpose, which is to **handle the selection of side content tabs, and the rendering of side content panels**. This is the case for `SideContent` (the desktop equivalent) but not `MobileSideContent`, whose responsibility had been wrongly expanded to also (prior to this PR and #2197):
* Render the editor (which is not a side content and is also not treated as such in the desktop version!)
  * This resulted in the need to pass the editor ref into `MobileSideContent` so that the editor can be force updated which is a code smell.
* Manipulate the draggable REPL (i.e., showing/hiding & enabling/disabling) whenever tabs are selected. By having the logic in `MobileSideContent`, it becomes tightly coupled to `MobileWorkspace` where the draggable REPL component resides in.
  * The design of the original desktop `SideContent` component was such that any such parent-component specific logic was passed in as a callback so as to abstract the details away from the `SideContent` component and allow it to be focused. For example, the `onChangeTabs` function in the `Playground` component.
* Push the run tab to the back of any other dynamically rendered tabs (such as runes). If you look at the code for this:
  https://github.com/source-academy/frontend/blob/505c1a5ac5573a5fcbc9693ea22e4c1f124f0c9a/src/commons/mobileWorkspace/mobileSideContent/MobileSideContent.tsx#L59-L67
  You will see that this introduces an implicit precondition that the last tab that is passed into `MobileSideContent` is the run tab.
  * I intend on removing this implicit precondition in another PR.

zhaojj2209 (approved):

PR #2214, opened 2022-09-12
Fix error & injection of iframe on hot reload

### LINKED ISSUE #2213: [Development Only] Error thrown & iframe injected when the application is hot reloaded
When the application is hot reloaded, the following error is logged in the console:
```
Uncaught ReferenceError: process is not defined
```

In addition, an iframe is injected that covers the entire application. To get around this, the iframe can be deleted from the DOM inspector or the frame can be manually reloaded. However, this makes development rather frustrating.

ianyong:
### Description

Fixes #2213.

Same issue as https://github.com/facebook/create-react-app/issues/11773.

There are two solutions outlined in the CRA issue above:
1. Upgrade `react-scripts` to v5.x.
1. Specify v6.0.9 of `react-error-overlay` via [selective dependency resolution](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/).

Unfortunately, we cannot upgrade `react-scripts` (CRA) to v5.0x because another dependency that we use, Create React App Configuration Override (CRACO) does not have a stable version that supports CRA v5.x yet:
> ℹ️ The latest CRACO release does not support CRA 5, but alpha builds will be regularly released over the next few days.
>
> -- https://github.com/dilanx/craco (as of 13 September 2022)

Attempting to do so results in errors when starting the frontend. As such, we have no choice but to go with the second option for now.

### Type of change

### How to test

1. Launch the frontend with developer tools open.
1. Make a change in the codebase to trigger a hot reload.
1. Check that the error described in the issue was not thrown, and that no iframe was injected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,11 +125,17 @@
     "husky": "^8.0.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.7.1",
+    "react-error-overlay": "6.0.9",
     "react-scripts": "^4.0.3",
     "react-test-renderer": "^18.2.0",
     "redux-saga-test-plan": "^4.0.5",
     "sass": "^1.54.0",
     "typescript": "~4.7.4"
+  },
+  "resolutions": {
+    "//": "See https://github.com/facebook/create-react-app/issues/11773 for more information.",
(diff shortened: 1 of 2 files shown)


zhaojj2209 (approved):
Hot reload works now, thanks for the fix!

PR #2223, opened 2022-09-17
Add ability to change cameras used for QR scanning

RichDom2185:
### Description

Some changes to improve the QR code scanning functionality based on feedback:
* Set mobile camera to open rear camera by default
* Allow toggling between different cameras

### Type of change

### How to test

* By default, on mobile, the rear camera should be the one that starts first
* There is now a "Change Camera" button which will try to toggle the camera. On desktop devices with only one camera, it will not do anything
  Screenshot 2022-09-17 at 5 32 29 PM


chownces (approved):
Tested on staging LGTM!

PR #2102, opened 2022-02-25
Update frontend for compatibility with js-slang PR #1220

leeyi45:
### Description
Since where tabs are stored changes with this [pull request](https://github.com/source-academy/js-slang/pull/1220), the frontend needs to be updated to properly retrieve them.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### Checklist

- [x] I have tested this code
- ~ I have updated the documentation~

PR #1617, opened 2021-02-15
Dynamic Module

MarcusTXK:
### Description

Updated the current SideContentHelper.ts to utilise the [new modules system](https://github.com/source-academy/modules/tree/modules/development)

 In `SideContentHelper`, `getDynamicTabs` recieves the debuggerContext and extracts the modules from it (using  `getModuleTabs`), and filters the modules based on their toSpawn condition. Side content tabs will only spawn if the `toSpawn` condition evaluates to true.

The relevant types needed were also added in `src/commons/sideContent/SideContentTypes.ts`.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### How to test
1) Pull js-slang from modules/development branch, and [follow the steps here](https://github.com/source-academy/js-slang#using-your-js-slang-in-local-source-academy) to link it to the cadet-front end

2) Pull cadet-frontend from modules/development branch, and setup the env file and run it. Set the module backend url to this:
`REACT_APP_MODULE_BACKEND_URL=http://localhost:8022`

3) Pull modules from modules/development branch, and [follow the steps here](https://github.com/source-academy/modules/tree/modules/development) to set it up and run it.

4) On the playground, the side content tab will spawn after the 'test' module is imported and the output in the REPL is "test":
image

### Checklist
- [x] I have tested this code

PR #2225, opened 2022-09-18
Remove unnecessary toSpawn property on SideContentTab type

ianyong:
### Context

To support multiple files & directories, there is a need for a file system view in the user interface. In the desktop view, this will be in the form of a sidebar on the left edge of the Source Academy that can be toggled open (much like most IDEs). Due to a lack of space in the mobile view, the file system UI will have to be a side content tab. Given the very similar purposes of the to-be-added sidebar and the existing side content tabs, I have been looking closely at the implementation of the latter so as to keep the interfaces consistent and interoperable.

One thing that stood out was the fact that the `SideContentTab` type has a `toSpawn` property of type `(context: DebuggerContext) => boolean`. However, all of the instances of `SideContentTab` in the frontend have `toSpawn` set to be the value `() => true`, which is a huge code smell.

Digging through the git history, I found that the `toSpawn` property was first introduced in #1352 to support the dynamic spawning of side content tabs. In #1617 however, a new `ModuleSideContent` type was introduced that corresponded to side content tabs from the [new modules system](https://github.com/source-academy/modules). The process of loading dynamic module tabs is as follows:
1. Using the debugger context, an array of `ModuleSideContent` is retrieved based on the modules imported.
1. Then, the array of `ModuleSideContent` is converted into an array of `SideContentTab`.
1. Finally, the `toSpawn` property (which is part of both `SideContentTab` and `ModuleSideContent`) is used to filter out the module tabs based on more granular rules specified within each module.
   * For example, the `rune` module will not spawn a tab if `show` is not called.

Given that the side content tabs declared in the frontend are static (i.e., will always be loaded), there is no reason for `SideContentTab` to have the `toSpawn` property. We can remove the property if we swap steps 2 and 3 above since the `ModuleSideContent` type already has the `toSpawn` property and represents dynamic module tabs (which should be the only side content tabs in which the `toSpawn` property is necessary).

### Description

Reorder the steps taken when processing the dynamic module tabs and remove the unnecessary `toSpawn` property on the `SideContentTab` type.

Also remove the unused `Modules` type. #2102 removed the only instance in which it was used but did not remove the type.

### Type of change

### How to test

Run Source programs that result in the loading of dynamic modules and verify that the behaviour is unchanged. For example, the following cases can be tested:
* Run a Source program that does not load any modules.
  ```javascript
  const x = 10;
  x * 3;
  ```
* Run a Source program that loads a module and should display a corresponding side content tab.
  ```javascript
  import { show, heart } from "rune";
  show(heart);
  ```
* Run a Source program that loads a module but should not display a corresponding side content tab.
  ```javascript
  import { show, heart } from "rune";
  ```

### DIFF
--- a/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
+++ b/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
@@ -375,15 +375,13 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
             label: `Task ${questionId + 1}`,
             iconName: IconNames.NINJA,
             body: <Markdown content={props.assessment!.questions[questionId].content} />,
-            id: SideContentType.questionOverview,
-            toSpawn: () => true
+            id: SideContentType.questionOverview
           },
           {
             label: `Contest Voting Briefing`,
             iconName: IconNames.BRIEFCASE,
             body: <Markdown content={props.assessment!.longSummary} />,
-            id: SideContentType.briefing,
-            toSpawn: () => true
+            id: SideContentType.briefing
           },
           {
             label: 'Contest Voting',
@@ -404,8 +402,7 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
                 }
               />
             ),
-            id: SideContentType.contestVoting,
-            toSpawn: () => true
+            id: SideContentType.contestVoting
           },
           {
             label: 'Contest Leaderboard',
@@ -419,8 +416,7 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
                 }
               />
             ),
-            id: SideContentType.contestLeaderboard,
-            toSpawn: () => false
+            id: SideContentType.contestLeaderboard
           }
         ]
       : [
@@ -433,8 +429,7 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
                 content={props.assessment!.questions[questionId].content}
               />
             ),
-            id: SideContentType.questionOverview,
-            toSpawn: () => true
+            id: SideContentType.questionOverview
           },
           {
             label: `Briefing`,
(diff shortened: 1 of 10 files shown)


leeyi45:
I've actually been looking into deprecating the entire `toSpawn` function. With module contexts, modules should be able to directly ask the frontend to load a tab when the corresponding module function is run instead of loading all tabs every run.

I've got some code for the frontend spun up, wondering if you have any opinions on it: https://github.com/source-academy/frontend/tree/event-tab-loading

Also tangentially related to this [issue](https://github.com/source-academy/modules/issues/147)

ianyong:
> I've actually been looking into deprecating the entire `toSpawn` function. With module contexts, modules should be able to directly ask the frontend to load a tab when the corresponding module function is run instead of loading all tabs every run.
> 
> I've got some code for the frontend spun up, wondering if you have any opinions on it: https://github.com/source-academy/frontend/tree/event-tab-loading
> 
> Also tangentially related to this [issue](https://github.com/source-academy/modules/issues/147)

I took a look through the diff of that branch against `master`. Is it possible to batch the request rather than querying the modules system for every module?

Other than that, I think it makes sense for the responsibility of determining whether side content tabs should be spawned to be moved into modules rather than be handled at the frontend (i.e., step 3 as mentioned in the description of the PR should be handled in modules instead).

For now, I would like this PR to be merged first so that I can proceed with my multiple files & directories implementation.

zhaojj2209 (approved):
LGTM!

@ianyong @leeyi45 do be prepared to resolve merge conflicts down the line depending on which of your feature branches get merged first, though, since this PR is to a feature branch and not to master.

martin-henz:
> 



> @ianyong @leeyi45 do be prepared to resolve merge conflicts down the line depending on which of your feature branches get merged first, though, since this PR is to a feature branch and not to master.

Yes, I agree. Feature branches are fine for a short while, but in my experience the longer ones get harder and harder to merge into the main branch. We've been burnt by this before, where at some point, we almost abandoned an entire semester of work because of merge issues.

chownces:
Tested the one and only `toSpawn: () => false` located in ContestLeaderboard. Works great!

ianyong:
> Tested the one and only `toSpawn: () => false` located in ContestLeaderboard. Works great!

Thanks for testing!

I realised I didn't explicitly state it in the PR description which is my bad. Since the `toSpawn` property is only ever used in `getDynamicTabs`, the side content tabs declared in the frontend will always be displayed even if `toSpawn` is a nullary function that returns `false`. In other words, the `toSpawn` property only has an effect if the side content tab came from the modules system, else it is never evaluated.

PR #2208, opened 2022-09-06
Improvements to the Module System 2

leeyi45:
With reference to this [PR](https://github.com/source-academy/modules/pull/153) and this [PR](https://github.com/source-academy/js-slang/pull/1289)

This PR also changes the `game` module to use the context system and removes the module params object.

PR #2227, opened 2022-09-20
Add HTML Chapter to SA

zhaojj2209:
### Description
Adds a new chapter to SA playground: HTML.
image

This PR will be updated again after #2208 and source-academy/js-slang#1298 have been merged, and js-slang version has been bumped (tentative date 29 Sep). PRing now so that any reviews and changes can be done while waiting.

### Type of change

### How to test

To test locally:
1. Pull this branch
2. Merge #2208 into branch
3. Link local js-slang master branch

Things to test (non-exhaustive):
- Running of HTML code
- Any console errors when running HTML code should be displayed in the SA output
- Mobile display
- Link sharing


src/commons/repl/Repl.tsx
```diff
@@ -40,7 +40,12 @@ type OwnProps = {
 
 const Repl = React.forwardRef<AceEditor, ReplProps>((props, ref) => {
   const cards = props.output.map((slice, index) => (
-    <Output output={slice} key={index} usingSubst={props.usingSubst || false} />
+    <Output
+      output={slice}
+      key={index}
+      usingSubst={props.usingSubst || false}
```
ianyong, line 46:
```suggestion
      usingSubst={props.usingSubst ?? false}
```

Given that the type of `props.usingSubst` is `boolean | undefined`, the nullish coalescing operator better captures the semantics. I understand that you didn't write this code, but it would be good if you could make this change.

src/commons/repl/Repl.tsx
```diff
@@ -81,6 +86,16 @@ export const Output: React.FC<OutputProps> = (props: OutputProps) => {
             <Pre className="logOutput">Check out the Stepper tab!</Pre>
           </Card>
         );
+      } else if (props.isHtml) {
+        return (
+          <Card>
+            <Pre className="logOutput">
```
ianyong, line 92:
```suggestion
            <Pre className="log-output">
```

CSS classnames should use kebab case by convention.

src/commons/repl/Repl.tsx
```diff
@@ -81,6 +86,16 @@ export const Output: React.FC<OutputProps> = (props: OutputProps) => {
...
+              Check out the HTML Display tab!
+              <br />
+              Please check the browser console for more detailed errors and warnings.
```
ianyong, line 95: A possible concern is that someone who might want to use the HTML display is probably a beginner and thus unfamiliar with the developer tools, in which case this statement makes no sense to them.

src/commons/sagas/WorkspaceSaga.ts
```diff
@@ -85,6 +90,15 @@ let breakpoints: string[] = [];
 export default function* WorkspaceSaga(): SagaIterator {
   let context: Context;
 
+  yield takeEvery(
+    ADD_HTML_CONSOLE_ERROR,
+    function* (action: ReturnType<typeof actions.addHtmlConsoleError>): any {
```
ianyong, line 95:
```suggestion
    function* (action: ReturnType<typeof actions.addHtmlConsoleError>) {
```

Maybe we should let TypeScript infer the type automatically instead of forcing it to be `any`?

src/commons/sideContent/SideContent.tsx
```diff
@@ -71,7 +71,7 @@ const SideContent = (props: SideContentProps) => {
    * Adds'side-content-tab-alert' style to newly spawned module tabs
```
ianyong, line 71: Header comment needs updating.

src/commons/sideContent/SideContentHtmlDisplay.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useEffect } from 'react';
+
+export interface SideContentHtmlDisplayProps {
```
ianyong, line 3: Not sure why this particular coding standard was set, but a `type` should be used here instead of an `interface`. See https://github.com/source-academy/frontend/wiki/Coding-Standard#interface-and-types.

src/commons/sideContent/SideContentHtmlDisplay.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const ERROR_MESSAGE_REGEX = /^Line [\d]+: /i;
+
+const SideContentHtmlDisplay: React.FC<SideContentHtmlDisplayProps> = props => {
+  const { content, handleAddHtmlConsoleError } = props;
+
+  useEffect(() => {
+    const handleEvent = (event: MessageEvent) => {
+      const msg: string = event.data;
+      if (msg.match(ERROR_MESSAGE_REGEX)) {
```
ianyong: The application crashes at this line if I try to run malformed HTML:
```html
<html>
  <
</html>
```

src/commons/sideContent/SideContentHtmlDisplay.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useEffect } from 'react';
+
+export interface SideContentHtmlDisplayProps {
+  content: string;
+  handleAddHtmlConsoleError: (errorMsg: string) => void;
+}
+
+const ERROR_MESSAGE_REGEX = /^Line [\d]+: /i;
```
ianyong, line 8:
```suggestion
const ERROR_MESSAGE_REGEX = /^Line \d+: /i;
```

Square brackets are unnecessary here since the character class has only one (meta)character.

src/commons/utils/IntroductionHelper.ts
```diff
@@ -25,6 +25,10 @@ const generateSourceDocsLink = (sourceChapter: Chapter, sourceVariant: Variant)
     );
   }
 
+  if (sourceChapter === Chapter.HTML) {
+    return 'However, you have chosen HTML, which renders your code directly as a HTML document.';
```
ianyong, line 29: image

I'm not very sure what this sentence is trying to tell me, or how it might help a beginner better understand what HTML is.

src/pages/playground/Playground.tsx
```diff
@@ -568,6 +579,25 @@ const Playground: React.FC<PlaygroundProps> = props => {
   const tabs = React.useMemo(() => {
     const tabs: SideContentTab[] = [playgroundIntroductionTab];
 
+    // For HTML, show only introduction tab, HTML Display tab only appears after running code
```
ianyong, line 582: Better phrasing?

src/styles/_workspace.scss
```diff
@@ -563,6 +563,12 @@ $code-color-error: #ff4444;
         }
       }
     }
+
+    .sa-html-display {
+      background-color: white;
+      width: 100%;
+      height: 60vh;
```
ianyong, line 570: There's an unnecessary scrollbar in the mobile view that can be wiggled about.

vmconnect_xpU3TMehtD

ianyong (changes requested):

src/commons/repl/Repl.tsx
zhaojj2209: The reason why the line is there is due to browser differences, e.g. Chrome adds extra warnings, Safari doesn't allow routing of console errors so the error message displayed will always be "Line 0: Script error".

Maybe it would be better for this explanation to be on the Playground Home tab, I'll shift the information there instead to avoid the output display being overly cluttered.

src/commons/utils/IntroductionHelper.ts
zhaojj2209: Updated description:
image

ianyong (approved):

PR #2239, opened 2022-10-15
Remove editor dimensions from Redux store

### LINKED ISSUE #2234: Remove editor width & height from the Redux store
# Context

As part of #2176, a sidebar was added to the workspace, resulting in a 3-column view:
image

Compared to the previous 2-column view which has only one point in which it can be resized along the x-axis, the new 3-column view has two such points:
1. Between the sidebar and the code editor.
1. Between the code editor and the side content panel.

The code responsible for handling the resizing can be found in `src/commons/workspace/Workspace.tsx`. For some reason however, the width & height of the code editor is part of the workspace state in the Redux store:
https://github.com/source-academy/frontend/blob/c4042f54ce9da83c067eb6c801f5d0bfefe64ceb/src/commons/workspace/WorkspaceTypes.ts#L82-L110

To make matters worse, the state in the Redux store is not used consistently, such as in the following instance where the width of the editor makes use of the state in the Redux store, but the height of the editor is set to `'100%'` (thus ignoring `editorHeight` in the Redux store):
https://github.com/source-academy/frontend/blob/c4042f54ce9da83c067eb6c801f5d0bfefe64ceb/src/commons/workspace/Workspace.tsx#L53-L66

The main issue of such an implementation (lack of consistency aside) is that we cannot rely on the browser's automatic calculation of width & height based on the CSS styles. What is being done now is that whenever the workspace is resized, the width of the editor is manually calculated and dispatched to the Redux store:
https://github.com/source-academy/frontend/blob/c4042f54ce9da83c067eb6c801f5d0bfefe64ceb/src/commons/workspace/Workspace.tsx#L54-L55

This is very brittle as any changes made to the layout will cause the calculation to be wrong. In fact, just the addition of the sidebar alone already makes the manual calculation wrong. (When resizing the boundary between the code editor and the side content panel, the code editor will grow/shrink beyond where the user drags it to when the sidebar is active.) On top of that, manual calculations become very hard to do once we introduce more than 1 point which can be resized along the x-axis.

# Proposed Changes

There does not seem to be any reason as to why the dimensions of the code editor are part of `WorkspaceState`. One possible reason I can think of is so as to persist the layout of the workspace (after the user has done some resizing) through page refreshes. However, this is currently not the case. Even if we were to support such a feature, it would also be better to implement it in a different way that stores the dimensions on resize and restores the dimensions on initial load (instead of having the code editor directly get its dimensions from the store) so that manual calculations are not necessary.

As such, the `editorWidth` and `editorHeight` fields should be removed from `WorkspaceState`. Then, the dimensions of the code editor can be dynamically calculated by the browser when resized.

ianyong:
### Description

See #2234 for the context behind this PR.

Fixes #2234. Part of the refactoring for #2176.

### Environment Visualizer

Note that there remains instances of `editorWidth` in `src/commons/sideContent/SideContentEnvVisualizer.tsx`:
https://github.com/source-academy/frontend/blob/40c3ea4036821a283063b5413935429fb751231a/src/commons/sideContent/SideContentEnvVisualizer.tsx#L14-L17

Searching for usages shows that whenever the `SideContentEnvVisualizer` component is used, no props are passed in:
image

Even in #2119 where the code snippet above was introduced, the `editorWidth` prop is unused. However, I opted to leave `SideContentEnvVisualizer` unchanged because there is a mess of logic dealing with `editorWidth`:
https://github.com/source-academy/frontend/blob/40c3ea4036821a283063b5413935429fb751231a/src/commons/sideContent/SideContentEnvVisualizer.tsx#L18-L91

Testing the environment visualizer on production also shows that the current implementation doesn't really work, meaning that to remove the manual calculations would require input on the expected behaviour from the team that wrote this component:
vmconnect_IdC4VtPBCP

The issue of the environment visualizer not resizing well is out of scope for this PR since none of the usages of `SideContentEnvVisualizer` even pass in the `editorWidth` prop.

### Type of change

### How to test

In any desktop workspace view (i.e., playground, mission, etc), drag the divider between the editor and the side content panel around to verify that resizing works.


zhaojj2209 (approved):
LGTM!

PR #2242, opened 2022-10-17
Hook up sidebar tabs to mobile view & add icons to sidebar tabs

ianyong:
### Description

Hook up sidebar tabs to the mobile view.
image
Sidebar tabs are only displayed in the mobile view (as normal side content tabs) if they have an `id` specified:
https://github.com/source-academy/frontend/blob/7748c8a8c6ac447c1f48128fe8cc70be23b3e247/src/commons/mobileWorkspace/MobileWorkspace.tsx#L200-L201

Also add icons to sidebar tabs (since side content tabs require them anyway).
image

In the mobile view, long presses are converted into the equivalent of right clicks on the desktop view by most browsers.

Part of #2176.

### Type of change

### How to test

Use the developer tools to force the mobile view and verify that the file system view has all of the functionality of the desktop view as described in #2236.

### DIFF
--- a/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
+++ b/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
@@ -807,15 +807,16 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
     externalLibrary: question?.library?.external?.name || 'NONE',
     replButtons: replButtons()
   };
+  const sideBarProps = {
+    tabs: []
+  };
   const workspaceProps: WorkspaceProps = {
     controlBarProps: controlBarProps(questionId),
     editorProps: editorProps,
     handleSideContentHeightChange: props.handleSideContentHeightChange,
     hasUnsavedChanges: props.hasUnsavedChanges,
     mcqProps: mcqProps,
-    sideBarProps: {
-      tabs: []
-    },
+    sideBarProps: sideBarProps,
     sideContentHeight: props.sideContentHeight,
     sideContentProps: sideContentProps(props, questionId),
     replProps: replProps
@@ -825,6 +826,7 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
     hasUnsavedChanges: props.hasUnsavedChanges,
     mcqProps: mcqProps,
     replProps: replProps,
+    sideBarProps: sideBarProps,
     mobileSideContentProps: mobileSideContentProps(questionId)
   };
 
--- a/src/commons/mobileWorkspace/MobileWorkspace.tsx
+++ b/src/commons/mobileWorkspace/MobileWorkspace.tsx
@@ -196,6 +196,9 @@ const MobileWorkspace: React.FC<MobileWorkspaceProps> = props => {
     }
   };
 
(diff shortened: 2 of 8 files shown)


zhaojj2209 (changes requested):
Unfortunately "hold-to-right-click" doesn't seem to be working on iOS, so I am unable to create any new files when testing on Apple devices. At the very least, an "add file" button needs to be added so that I can proceed to test if "hold-to-right-click" still works on files/directories. Note that the "add file" button needs to be added to both mobile and desktop view since there are tablets that use desktop view.

https://user-images.githubusercontent.com/54243224/196328332-47ee62e8-31da-4c1d-bbb3-415630a1e380.MOV

^ Tested on Chrome on iPhone

https://user-images.githubusercontent.com/54243224/196329093-05f6892c-7ef3-41b8-84e3-b4146834f491.MP4

^ Tested on Chrome on iPad

Furthermore, when there are no files, the files tab on mobile view is blank. Showing some introductory text may be beneficial for students who may be confused by the blank screen.

ianyong:
Thanks for testing!

> Unfortunately "hold-to-right-click" doesn't seem to be working on iOS, so I am unable to create any new files when testing on Apple devices.

Since all browsers on iOS are actually running WebKit under the hood, this seems to be a Safari issue.

Firefox on Android works:
https://user-images.githubusercontent.com/5585517/196336375-98ab6e9e-d4b5-4f05-9531-d418520e1575.mp4

It would seem to me that this issue of long pressing an item resulting in some other item being selected is an issue with the browser and not the code. I have no idea what can possibly be done to fix this at the Source Academy level nor do I have the means of debugging this issue.

> At the very least, an "add file" button needs to be added so that I can proceed to test if "hold-to-right-click" still works on files/directories. Note that the "add file" button needs to be added to both mobile and desktop view since there are tablets that use desktop view.

Why would it work on files/directories but not on the empty space? The context menu component that is used is the same, and the problem of long presses selecting something else is a browser bug.

Also, this is not an acceptable compromise. As more features are added, we need to be aware that the context menu might look something like this in the future:
image

How do you propose we add a button on the UI for every single action that can be performed? There is also the issue whereby actions are performed on a particular file/directory, meaning that we need both a target and an action. When the "add file" button is pressed, where should the file go? We would need to introduce the notion of a working directory somehow.

> Furthermore, when there are no files, the files tab on mobile view is blank. Showing some introductory text may be beneficial for students who may be confused by the blank screen.

This is a good point which I have intentionally held off on as discussed in the last point of https://github.com/source-academy/frontend/pull/2236#issuecomment-1272835157 because (1) things will be clearer as we progress and (2) the aim right now is to come up with a minimum viable product.

@zhaojj2209 Unfortunately, I will need your help to debug the long press issue on iOS.

zhaojj2209 (approved):
Apologies for the lack of clarification, the suggestion to add a "Add file" button was for testing purposes. I've since added a file manually through the code to test, and have confirmed that "hold-to-right-click" doesn't work even when there are files, either.

You're right in that this is probably a browser issue, in which case we will likely need to look for workarounds. We can come back to this issue at a later time; other than this, LGTM.

ianyong:
> You're right in that this is probably a browser issue, in which case we will likely need to look for workarounds. We can come back to this issue at a later time; other than this, LGTM.

Tracking in #2243.

PR #2250, opened 2022-11-04
Fix GitHub persistence feature only retrieving 30 repositories

### LINKED ISSUE #2245: List of GitHub repositories shown when using GitHub to persist files is capped at 30
Currently, a call is made to the `https://api.github.com/user/repos` endpoint to retrieve the list of repositories that a user has access to when the `Open` or `Save As` buttons are clicked:
image

However, the API only returns 30 results by default (see https://docs.github.com/en/rest/repos/repos#list-repositories-for-the-authenticated-user). As such, not all repositories are listed.

ianyong:
### Description

By default, [GitHub's API for listing the repositories for the authenticated user](https://docs.github.com/en/rest/repos/repos#list-repositories-for-the-authenticated-user) retrieves only 30 results per page, sorted by repository name. This means that GitHub accounts with more than 30 repositories will only be able to choose between a subset of the available repositories when using the GitHub persistence feature.

The Source Academy makes use of the [GitHub SDK octokit.js](https://github.com/octokit/octokit.js). As such, we can easily query for all results of any particular API endpoint using the [pagination API](https://github.com/octokit/octokit.js#pagination). Since API calls using the pagination API is sequential (i.e., the previous API response needs to be received before the pagination API knows whether it needs to query for the next page of results), we also increase the number of results per page to 100 (the maximum allowed) so as to speed up the retrieving of repositories.

Fixes #2245.

### Type of change

### How to test

1. Follow the instructions [here](https://github.com/source-academy/frontend/wiki/GitHub-Integration:-Initial-Set-up) to set up OAuth for `localhost`.
1. Login to a GitHub account with more than 30 repositories.
1. Verify that all repositories appear in the user's repository lists.

Example network log (with the results per page temporarily set to 30 instead of 100 to show that the pagination is working):
image

### DIFF
--- a/src/commons/sagas/GitHubPersistenceSaga.ts
+++ b/src/commons/sagas/GitHubPersistenceSaga.ts
@@ -77,6 +77,7 @@ function* githubOpenFile(): any {
   const userRepos: ListForAuthenticatedUserData = yield call(
     async () =>
       await octokit.paginate(octokit.repos.listForAuthenticatedUser, {
+        // 100 is the maximum number of results that can be retrieved per page.
         per_page: 100
       })
   );
@@ -146,6 +147,7 @@ function* githubSaveFileAs(): any {
   const userRepos: ListForAuthenticatedUserData = yield call(
     async () =>
       await octokit.paginate(octokit.repos.listForAuthenticatedUser, {
+        // 100 is the maximum number of results that can be retrieved per page.
         per_page: 100
       })
   );
--- a/src/commons/sagas/GitHubPersistenceSaga.ts
+++ b/src/commons/sagas/GitHubPersistenceSaga.ts
@@ -66,10 +66,10 @@ function* githubLogoutSaga() {
 }
 
 function* githubOpenFile(): any {
-  const octokit = GitHubUtils.getGitHubOctokitInstance() || {
-    users: { getAuthenticated: () => {} },
-    repos: { listForAuthenticatedUser: () => {} }
-  };
+  const octokit = GitHubUtils.getGitHubOctokitInstance();
+  if (octokit === undefined) {
+    return;
+  }
 
   type ListForAuthenticatedUserData = GetResponseDataTypeFromEndpointMethod<
     typeof octokit.repos.listForAuthenticatedUser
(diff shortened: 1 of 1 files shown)


ianyong:
@chekjun Is there a reason why the methods on the Octokit object are mocked to be nullary functions that do nothing?

https://github.com/source-academy/frontend/blob/2830bbe78b9d45d09002941de45c49cdb08fb987/src/commons/sagas/GitHubPersistenceSaga.ts#L69-L72

As opposed to directly failing if the Octokit object is `undefined` as seen in a related but different function:

https://github.com/source-academy/frontend/blob/2830bbe78b9d45d09002941de45c49cdb08fb987/src/commons/sagas/GitHubPersistenceSaga.ts#L111-L112

The issue is that when the Octokit object is not loaded, the `paginate` method will also not be available. I could follow the current implementation and add a mock `paginate` function that does nothing, but it leads to messier code. As such, I would like your input on why the code was originally written like that.

chekjun:
> @chekjun Is there a reason why the methods on the Octokit object are mocked to be nullary functions that do nothing?
> 
> https://github.com/source-academy/frontend/blob/2830bbe78b9d45d09002941de45c49cdb08fb987/src/commons/sagas/GitHubPersistenceSaga.ts#L69-L72
> 
> As opposed to directly failing if the Octokit object is `undefined` as seen in a related but different function:
> 
> https://github.com/source-academy/frontend/blob/2830bbe78b9d45d09002941de45c49cdb08fb987/src/commons/sagas/GitHubPersistenceSaga.ts#L111-L112
> 
> The issue is that when the Octokit object is not loaded, the `paginate` method will also not be available. I could follow the current implementation and add a mock `paginate` function that does nothing, but it leads to messier code. As such, I would like your input on why the code was originally written like that.

Returning if octokit is undefined should work. I think this was an oversight on my part as I was new to React then. If I recall yarn run start would give me an error since octokit could be undefined and I forgot to check it in the similar way to Line 111 to Line 112.

ianyong:
> Returning if octokit is undefined should work. I think this was an oversight on my part as I was new to React then. If I recall yarn run start would give me an error since octokit could be undefined and I forgot to check it in the similar way to Line 111 and Line 112.

I see, thank you!

src/commons/sagas/GitHubPersistenceSaga.ts
```diff
@@ -66,22 +66,20 @@ function* githubLogoutSaga() {
...
-    repos: { listForAuthenticatedUser: () => {} }
-  };
-
-  type ListForAuthenticatedUserResponse = GetResponseTypeFromEndpointMethod<
-    typeof octokit.repos.listForAuthenticatedUser
-  >;
-  const results: ListForAuthenticatedUserResponse = yield call(
-    octokit.repos.listForAuthenticatedUser
-  );
+  const octokit = GitHubUtils.getGitHubOctokitInstance();
```
ianyong, line 69: Ideally, the return type of `GitHubUtils.getGitHubOctokitInstance()` should be `Octokit | undefined` instead of `any` but changing the type will result in having to fix a lot of type errors, which I deem to be out of scope of this PR.

zhaojj2209 (approved):
LGTM, tested also by manually lowering the results per page value.

PR #2262, opened 2022-12-01
Fix REPL representation bug

shenyih0ng:
## Description

Closes https://github.com/source-academy/js-slang/issues/1309

### DIFF
--- a/src/commons/repl/Repl.tsx
+++ b/src/commons/repl/Repl.tsx
@@ -2,14 +2,12 @@ import { Card, Classes, Pre } from '@blueprintjs/core';
 import classNames from 'classnames';
 import { parseError } from 'js-slang';
 import { Chapter, Variant } from 'js-slang/dist/types';
-import { stringify } from 'js-slang/dist/utils/stringify';
 import * as React from 'react';
 import AceEditor from 'react-ace';
 import { HotKeys } from 'react-hotkeys';
 
 import { InterpreterOutput } from '../application/ApplicationTypes';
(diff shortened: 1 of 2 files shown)


src/commons/repl/Repl.tsx
```diff
@@ -127,16 +125,6 @@ export const Output: React.FC<OutputProps> = (props: OutputProps) => {
   }
 };
 
-const renderResult = (value: any) => {
-  /** A class which is the output of the show() function */
-  const ShapeDrawn = (window as any).ShapeDrawn;
-  if (typeof ShapeDrawn !== 'undefined' && value instanceof ShapeDrawn) {
-    return <SideContentCanvasOutput canvas={value.$canvas} />;
```
shenyih0ng, line 134: @chownces wanted to check if `ShapeDrawn` is still used? Found it being used in legacy `externLibs/.../webGL(rune|graphics).js`

> there are a couple of duplicated `renderResult` functions lying around as well, if `ShapeDrawn` is deprecated we can clean those up as well

chownces: sry for the delayed response! i believe we can deprecate this now that runes has been migrated to the modules repo

shenyih0ng: removed the use of the `ShapeDrawn` type, as for the deprecation of modules in `externalLibs`, i will batch it together in another PR

chownces (approved):
LGTM! Thanks :)

PR #2263, opened 2022-12-04
Make editor value non-nullable

ianyong:
### Description

There is no reason for the editor value to be nullable. Semantically, there is no difference between an editor value that is `null` and the empty string because the editor is always present. In future, to have multiple-files support, there can be any number of editors (including 0), but there should only ever be one and only one editor value for every editor.

The very fact that non-null assertions are placed in all of the workspace containers is a huge code smell that indicates that the `editorValue` field should not be nullable.

### Type of change

### How to test

There should not be any change in behaviour.

### DIFF
--- a/src/commons/assessment/AssessmentTypes.ts
+++ b/src/commons/assessment/AssessmentTypes.ts
@@ -129,7 +129,7 @@ export type BaseQuestion = {
   answer: string | number | ContestEntry[] | null;
   comments?: string;
   content: string;
-  editorValue?: string | null;
+  editorValue?: string;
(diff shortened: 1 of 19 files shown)


zhaojj2209:
There is one instance of `editorValue || ''` in the codebase (in GitHubPersistenceSaga.ts), it can be safely removed.

There's also one more instance of optional `editorValue` in `GradingEditor.tsx`, do look into whether it can be safely made non-optional as well.

zhaojj2209 (approved):
LGTM!

PR #2265, opened 2022-12-07
Refactor workspace Redux state in preparation for multiple editor tabs

ianyong:
### Description

Refactor the state of the workspace in the Redux store as such:
```ts
export type EditorTabState = {
  readonly value: string;
  readonly prependValue: string;
  readonly postpendValue: string;
  readonly highlightedLines: HighlightedLines[];
  readonly newCursorPosition?: Position;
};

export type WorkspaceState = {
  // ...
  readonly activeEditorTabIndex: number | null;
  readonly editorTabs: EditorTabState[];
  // ...
}
```

All state which should be tied to an editor tab instead of the workspace have been moved into `EditorTabState`. The `WorkspaceState` keeps track of an array of `EditorTabState`s, as well as the active editor tab index.

Note that the following properties were renamed to better fit the new `EditorTabState`:
* `editorValue` -> `value`
* `editorPrepend` -> `prependValue`
* `editorPostpend` -> `postpendValue`
  * Strictly speaking, "postpend" should be "append", but I decided to keep the current naming to avoid adding more complexity to this PR.

All other properties in `WorkspaceState` should be tied to the workspace and not any individual editor tab, except for `breakpoints`, which will be dealt with in a follow-up PR because it is apparently unused. The removal of `breakpoints` will be dealt with in a separate PR so that the change will be more visible in the Git history.

Note that some instances of `Partial` were updated to `DeepPartial`. This is because `Partial` only makes the top-level properties of any object optional, meaning that nested properties are still required. `DeepPartial` from the `redux` library solves this issue by recursively making all nested properties optional.

Even though the shape of the workspace state has been updated, there should be no change in functionality. Currently, instances in which the first editor tab is hardcoded is marked with the following `TODO`:
```ts
// TODO: Hardcoded to make use of the first editor tab. Rewrite after editor tabs are added.
```

All instances of such `TODO`s will be incrementally addressed before the feature branch is merged into `master`.

Part of the refactoring for https://github.com/source-academy/frontend/issues/2176.

### Type of change

### How to test

There should not be any change in behaviour.

### DIFF
--- a/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
+++ b/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
@@ -334,6 +334,7 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
     props.handleUpdateCurrentAssessmentId(assessmentId, questionId);
     props.handleResetWorkspace({
       autogradingResults,
+      // TODO: Hardcoded to make use of the first editor tab. Rewrite after editor tabs are added.
       editorTabs: [
         { value: editorValue, prependValue: editorPrepend, postpendValue: editorPostpend }
       ],
--- a/src/commons/sagas/PersistenceSaga.tsx
+++ b/src/commons/sagas/PersistenceSaga.tsx
@@ -237,6 +237,7 @@ export function* persistenceSaga(): SagaIterator {
         yield call(ensureInitialisedAndAuthorised);
 
         const [code, chapter, variant, external] = yield select((state: OverallState) => [
+          // TODO: Hardcoded to make use of the first editor tab. Rewrite after editor tabs are added.
           state.workspaces.playground.editorTabs[0].value,
           state.workspaces.playground.context.chapter,
           state.workspaces.playground.context.variant,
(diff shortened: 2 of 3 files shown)


martin-henz:
Looks like CI is failing: https://github.com/source-academy/frontend/actions/runs/3640005248/jobs/6144106133
Is this still "in progress"?

ianyong:
> Looks like CI is failing: https://github.com/source-academy/frontend/actions/runs/3640005248/jobs/6144106133 Is this still "in progress"?

It seems that the Git pre-push hook does not type check tests vs the CI pipeline which type checks all TypeScript files:
https://github.com/source-academy/frontend/blob/91aaaffce1fd34fd10d4b5cefd0b1b7794f79a00/scripts/test.sh#L10-L18

martin-henz:
> > It seems that the Git pre-push hook does not type check tests vs the CI pipeline which type checks all TypeScript files
> 
> @martin-henz Created #2266; shouldn't be too difficult a change.

Great. Thanks for looking into this.

ianyong:
> LGTM, just pointing out some places that I think needs TODOs so as to make things easier down the line.

Thanks for catching these instances!

PR #2269, opened 2022-12-09
Make updating of breakpoints go through the Redux store & refactor in preparation for multiple editor tabs

### LINKED ISSUE #2268: All workspaces make use of the same breakpoints state
When the user changes workspace, the state of the breakpoints set remains the same as the last workspace in which the user added or removed a breakpoint. This can be confusing to the user if they have not set any breakpoints in the new workspace as the breakpoints are not displayed in the editor of the new workspace, yet are hit when run.

An example to replicate this is:
1. Make use of a version of the Source Academy with access to at least 2 workspaces. The NUS version of the Source Academy is ideal for this.
1. Write a Source 3+ program in any workspace (say the playground).
1. Write another Source 3+ program in another workspace (say Sourcecast). This program should have at least as many lines as the previous program. For simplicity, we can use the same program for both workspaces.
1. In either of the workspaces, set at least one breakpoint.
1. Switch over to the other workspace **without leaving the browser page**. This is because the breakpoints state will be cleared on refresh.
1. Set the language to Source 3+.
1. Run the program. The program will be evaluated as if the breakpoints set in the previous workspace exist in the current workspace.

ianyong:
### Description

While working on #2265, I realised that the `breakpoints` property in `WorkspaceState` is never updated:
https://github.com/source-academy/frontend/blob/78ff16511385a31e8e66fbfd72a3e201cbbbe8bc/src/commons/workspace/WorkspaceTypes.ts#L81-L83

After tracing the code, the reason for this is that the `WorkspaceReducer` does not have any logic for doing so. This was rather perplexing as the breakpoints clearly work (when using the environment visualiser in Source 3 and above - thanks to @zhaojj2209 for explaining this to me when I was trying to figure out what breakpoints even do in Source 1).

It turns out the `breakpoints` state is not actually stored in the Redux store, but as a local variable in `src/commons/sagas/WorkspaceSaga.ts`:
https://github.com/source-academy/frontend/blob/78ff16511385a31e8e66fbfd72a3e201cbbbe8bc/src/commons/sagas/WorkspaceSaga.ts#L89

When the `UPDATE_EDITOR_BREAKPOINTS` action is dispatched, there is no corresponding logic in `WorkspaceReducer` to handle it. Instead, a Saga responds to it and updates the local `breakpoints` variable:
https://github.com/source-academy/frontend/blob/78ff16511385a31e8e66fbfd72a3e201cbbbe8bc/src/commons/sagas/WorkspaceSaga.ts#L237-L243

Subsequently, when `evalEditor` is run (to evaluate the program), the local `breakpoints` variable is accessed instead of the `breakpoints` property in `WorkspaceState`.
https://github.com/source-academy/frontend/blob/78ff16511385a31e8e66fbfd72a3e201cbbbe8bc/src/commons/sagas/WorkspaceSaga.ts#L537-L602

I have no idea why the original author wrote the code like this, but it is clearly wrong to have this one particular state not go through the Redux store as it is not separated by workspace (see #2268). On top of that, we will need to maintain a separate `breakpoints` state for each editor tab in the future. As such, the code to update breakpoints has been refactored to go through the Redux store, as well as the `WorkspaceState` updated such that each `EditorTabState` has its own `breakpoints` state.

Fixes #2268.

### Type of change

### How to test

There should not be any change in behaviour other than fixing the issue described in #2268 (instructions on how to replicate can be found within).


zhaojj2209 (approved):
LGTM!

PR #2272, opened 2022-12-12
Decouple highlighted lines state from EVAL_INTERPRETER_SUCCESS action & make updating of highlighted lines go through Redux store

ianyong:
### Description

While refactoring the `EVAL_INTERPRETER_SUCCESS` action in the `WorkspaceReducer` to not make use of a hardcoded editor tab state, I had to figure out why there was a need to reset the `highlightedLines` state. As it turns out, there is no reason for that.

The `highlightedLines` state is necessary for the painting of the blue lines whenever a breakpoint is hit in Source 3+. From tracing the code, it is correctly reset when the debugger resumes normal evaluation after all breakpoints have been hit within the `DEBUG_RESUME` saga in line 214:
https://github.com/source-academy/frontend/blob/78ff16511385a31e8e66fbfd72a3e201cbbbe8bc/src/commons/sagas/WorkspaceSaga.ts#L202-L216

image
However, when the debugger is stopped prematurely via the "Stop Debugger" button instead, the `highlightedLines` state is not correctly reset. This is because whoever wrote the `DEBUG_RESET` saga decided to bypass the Redux store and call the code that handles the painting and clearing of lines in `public/externalLibs/inspector/inspector.js` directly as seen in line 222 below:
https://github.com/source-academy/frontend/blob/78ff16511385a31e8e66fbfd72a3e201cbbbe8bc/src/commons/sagas/WorkspaceSaga.ts#L218-L226

Not to mention, lines 221 and 223 perform the same action twice...

### Changes made

All changes to the highlighted lines now go through the Redux store. The `highlightLine` function in `src/commons/utils/JsSlangHelper.ts` was also split up into a `highlightLine` and a `highlightClean` function so as to enforce single responsibility, as well as have safer typing & more readable code through the omission of `undefined`. Finally, the highlighted lines state is no longer changed by the `EVAL_INTERPRETER_SUCCESS` action because there is no reason for this coupling.

### Type of change

### DIFF
--- a/src/commons/utils/JsSlangHelper.ts
+++ b/src/commons/utils/JsSlangHelper.ts
@@ -98,10 +98,16 @@ export function visualizeEnv({ context }: { context: Context }) {
   }
 }
 
-export function highlightLine(line: number | undefined) {
+export function highlightClean() {
   if ((window as any).Inspector) {
     (window as any).Inspector.highlightClean();
-    // if number is undefined it just clears the highlighting.
+  } else {
+    throw new Error('Inspector not loaded');
+  }
+}
+
+export function highlightLine(line: number) {
+  if ((window as any).Inspector) {
     (window as any).Inspector.highlightLine(line);
   } else {
     throw new Error('Inspector not loaded');
--- a/src/commons/sagas/WorkspaceSaga.ts
+++ b/src/commons/sagas/WorkspaceSaga.ts
@@ -221,7 +221,6 @@ export default function* WorkspaceSaga(): SagaIterator {
     yield put(actions.clearReplOutput(workspaceLocation));
     yield put(actions.highlightEditorLine([], workspaceLocation));
     highlightLine(undefined);
-    yield put(actions.clearReplOutput(workspaceLocation));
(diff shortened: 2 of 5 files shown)


src/commons/utils/JsSlangHelper.ts
```diff
@@ -98,10 +98,16 @@ export function visualizeEnv({ context }: { context: Context }) {
...
   if ((window as any).Inspector) {
     (window as any).Inspector.highlightClean();
-    // if number is undefined it just clears the highlighting.
+  } else {
+    throw new Error('Inspector not loaded');
+  }
+}
+
+export function highlightLine(line: number) {
+  if ((window as any).Inspector) {
```
ianyong, line 110: Ideally, I would type the `window` object so that there is no need to cast to `any`. However, this PR itself is already far beyond the scope of my work and I have deadlines to meet.

ianyong: Created a follow-up issue #2273.

zhaojj2209 (approved):

Issue #2276, opened 2022-12-13
Mobile keyboard steals focus from underlying text container


ianyong:
The `MobileWorkspace` component has a `MobileKeyboard` for inputting commonly used character(s) in Source. However, clicking on the mobile keyboard makes the underlying text container (be it the editor or the REPL) lose focus because the mobile keyboard gains focus from being clicked on.

![vmconnect_sXn9Et2Tyq](https://user-images.githubusercontent.com/5585517/207343224-c6103c7a-fc1e-43f0-98eb-71442175a3a5.gif)

In the screen capture above, it can be seen that the editor loses focus when the mobile keyboard is clicked on (as indicated by the cursor no longer blinking).

martin-henz:
Bug seems to be gone now. Checked by @martin-henz on Android (Samsung Galaxy) and Joven (iPhone 13).

PR #2278, opened 2022-12-13
Remove editor ref forwarding in favour of callbacks

ianyong:
### Description

The mobile workspace comes with a mobile keyboard for inputting characters common to Source that might be hard to type on typical phone keyboards. To allow the mobile keyboard to manipulate the editors (which are child components of the `MobileWorkspace` component), the current codebase makes use of React refs to track which editor has focus:
https://github.com/source-academy/frontend/blob/78ff16511385a31e8e66fbfd72a3e201cbbbe8bc/src/commons/mobileWorkspace/MobileWorkspace.tsx#L95-L112

This however results in the need to forward refs through the component hierarchy from the child editor components to the `MobileWorkspace` component which is rather messy. In two separate locations, the merging of refs is used, which further complicates the codebase.

Instead of having the parent `MobileWorkspace` component be responsible for setting the event listeners on the child editor components (as seen in the code snippet above), we can make use of callbacks which are passed to the child editor components. This shifts the responsibility of maintaining event listeners to the child editor components, resulting in more cohesive code. On top of that, this also removes the need to merge refs, leading to more readable code.

Part of the refactoring for #2176 in preparation for adding multiple editor tabs.

### Type of change

### How to test

Go into the mobile view and test that the mobile keyboard is able to input characters into the playground editor, the REPL, and the Sourcecast editor (the playground editor & Sourcecast editor are different components). Note that there is a separate issue where the mobile keyboard steals focus from the underlying editor (see #2276).


zhaojj2209 (approved):
LGTM!

PR #2281, opened 2022-12-14
:sparkles: feat: add total xp table

ShenyiCui:
### Description
Add a total XP table so that admins can view the `total xp` of all students and download a CSV of all their total XP.

### Type of change

### Checklist
- [x] I have tested this code

### Screenshots
Screenshot 2022-12-14 at 9 52 45 AM


src/pages/academy/xpCalculation/XpCalculation.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+type PageState = {
+  currPage: number;
+  maxPages: number;
+  rowCountString: string;
+  isBackDisabled: boolean;
+  isForwardDisabled: boolean;
+};
+
+const XpCalculation = ({ allUserXp, handleAllUserXpFetch }: XpCalculationProps) => {
```
RichDom2185, line 36: I think this typing might be better:

```suggestion
const XpCalculation: React.FC<XpCalculationProps> = ({ allUserXp, handleAllUserXpFetch }) => {
```

src/pages/academy/xpCalculation/XpCalculation.tsx
```diff
@@ -0,0 +1,249 @@
...
+    { field: 'name' },
+    { field: 'NUS Net ID' },
+    { field: 'Assessment Xp' },
+    { field: 'Achievement Xp' }
+  ];
+
+  const defaultColDef = {
+    resizable: true,
+    sortable: true
+  };
```
RichDom2185, line 73: These constants don't depend on anything; moving them out of the component (so that a shared copy is used instead of creating a new object each time) may improve performance.

RichDom2185: Moreover, marking them as read-only using `as const` may also be appropriate.

src/pages/academy/xpCalculation/XpCalculation.tsx
```diff
@@ -0,0 +1,249 @@
...
+  //   if (gridApi) {
+  //     gridApi.setQuickFilter(filterText);
+  //   }
+  // };
+
+  // const handleFilterKeypress = (event: React.KeyboardEvent<HTMLInputElement>) => {
+  //   if (event.key === 'Enter') {
+  //     handleApplyFilter();
+  //   }
+  // };
```
RichDom2185, line 158: What is the purpose of this commented out code?

src/pages/academy/xpCalculation/XpCalculation.tsx
```diff
@@ -0,0 +1,249 @@
...
+            return gridApi.paginationGoToPreviousPage();
+          case 'next':
+            return gridApi.paginationGoToNextPage();
+          case 'last':
+            return gridApi.paginationGoToLastPage();
+          default:
+        }
+      }
+    };
+  };
```
RichDom2185, line 109: Why not make `changePaginationView` take in two arguments, `gridApi` and `type`? That way, it can be memoized for performance.

src/pages/academy/xpCalculation/XpCalculation.tsx
```diff
@@ -0,0 +1,249 @@
...
+          </div>
+          <div className="right-controls">
+            <Button icon={IconNames.EXPORT} onClick={exportCSV}>
+              <span className="hidden-xs">Export to CSV</span>
+            </Button>
+          </div>
+        </div>
+      </div>
+    );
+  }; /* Display either a loading screen or a table with overviews. */
```
RichDom2185, line 208: Would it be better if this component is moved out of `XpCalculation`? This component looks very similar to the ones used in `src/pages/academy/grading/Grading.tsx`. I think it may be reasonable to abstract this further to reduce the code duplication.

src/pages/academy/xpCalculation/XpCalculation.tsx
```diff
@@ -0,0 +1,249 @@
...
+        rowData={rowData}
+        rowHeight={30}
+        pagination={true}
+        paginationPageSize={20}
+        suppressCellSelection={true}
+        suppressMovableColumns={true}
+        suppressPaginationPanel={true}
+      />
+    </div>
+  );
```
RichDom2185, line 238: Ditto on abstraction and possible code duplication.

RichDom2185 (commented):
Understand that there were time constraints involved hence the priority was to get things working at the time, but if you don't mind, could you take another look at the potential code quality improvements below?

src/pages/academy/xpCalculation/XpCalculationContainer.ts
```diff
@@ -0,0 +1,22 @@
+import { connect, MapDispatchToProps, MapStateToProps } from 'react-redux';
```
chownces, line 1: let's shift to useDispatch instead

src/pages/academy/xpCalculation/XpCalculation.tsx
chownces: let's KIV for the future and get this merged in first to prepare for the sem

src/pages/academy/xpCalculation/XpCalculation.tsx
chownces: let's KIV for the future and get this merged in first to prepare for the sem

chownces (approved):

RichDom2185:
As discussed offline, we will revert this. Unfortunately, this is hundreds of commits behind the main branch so we'll have to create the revert commit manually.

PR #2257, opened 2022-11-19
Simplify Playground component

RichDom2185:
### Description

Move out a bunch of handlers to utilise the `useDispatch` hook instead to make the code more readable and also be in line with the components being functional, not class-based. Further refactoring awaits in a future PR.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Code quality improvements

### How to test

All features should still work as per usual.

### Checklist

- [x] I have tested this code

PR #2282, opened 2022-12-14
Multiple files and directories checkpoint 2

ianyong:
### Description

Merging the (disabled) feature into `master` so as to avoid a single big bang integration at the end which might have a lot of conflicts. Please see the squashed commits and their respective PRs for more information on the changes.

I initially planned to only merge this feature branch into `master` after the file system view was added. However, it seems that there are other large-scale refactoring taking place which resulted in quite a few merge conflicts. (I have since rebased the branch so as to have a linear Git history when merged.) Currently, since the sidebar is disabled, and since all of the Redux store logic is hardcoded to only use the first editor tab despite the altered shape of the workspace state, there should be no change in behaviour on production aside from fixing the global breakpoints state bug.

Part of #2176.

### Type of change

### How to test

There should not be any change in behaviour (aside from fixing the global breakpoints state bug) since the feature is disabled on production.


ianyong:
@chownces As discussed, to be merged in without squashing (so as to preserve the git history of the feature branch), thank you!

src/pages/playground/PlaygroundContainer.ts
```diff
@@ -0,0 +1,145 @@
+import { Chapter, Variant } from 'js-slang/dist/types';
```
RichDom2185, line 1: Oops, I think this may have been missed out in the rebase due to the squashed commits; sorry about the conflicts. In #2257, I renamed `PlaygroundContainer.ts` to `PlaygroundContainer.tsx` as I needed to use JSX inside, so this file shouldn't exist anymore.

chownces: maybe @RichDom2185 could you readd the changes to this PR?

chownces: maybe @RichDom2185 could you readd these changes to this PR? LGTM after this is resolved

ianyong: @RichDom2185 Thanks for catching this! I've fixed up the earliest commit which re-added `PlaygroundContainer.ts` via an interactive rebase, and manually checked through all commits that had merge conflicts while rebasing such that each commit is in a valid state. `PlaygroundContainer.ts` should be entirely purged from history now.

src/commons/mobileWorkspace/MobileWorkspace.tsx
```diff
@@ -92,30 +97,68 @@ const MobileWorkspace: React.FC<MobileWorkspaceProps> = props => {
...
-    replRef.current?.editor.on('focus', () => {
-      setKeyboardInputRef(replRef);
-    });
-    const clearRef = () => setKeyboardInputRef(emptyRef);
-    editorRef.current?.editor.on('blur', clearRef);
-    replRef.current?.editor.on('blur', clearRef);
-  }, []);
+  const clearTargetKeyboardInput = () => setTargetKeyboardInput(null);
+
+  const enableMobileKeyboardForEditor = (props: EditorProps): EditorProps => {
```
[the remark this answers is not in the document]
ianyong: Assuming the components only re-render when the state of `MobileWorkspace` changes (when the `targetKeyboardInput` is changed), this shouldn't cause any more re-renders than the previous method if I'm not wrong because the previous method also requires `MobileWorkspace` to keep track of which ref to use (as opposed to using a single ref and updating its `current` value which would not cause re-renders).

chownces (approved):
LGTM. Thanks for looking into this!

PR #2285, opened 2022-12-15
Add action for partially updating the state of the active editor tab

### LINKED ISSUE #2284: Assessment Workspace line highlighting crashes frontend
Bug introduced in #2282. Only occurs in assessments where the AceEditor is being used

image

ianyong:
### Description

In #2265, the actions for resetting and updating the workspace state had their type signatures changed as such:
> Note that some instances of `Partial` were updated to `DeepPartial`. This is because `Partial` only makes the top-level properties of any object optional, meaning that nested properties are still required. `DeepPartial` from the redux library solves this issue by recursively making all nested properties optional.

Upon closer inspection into how the workspace state is partially updated, this change was incorrect and should not have been made. This is because nested objects in an array cannot be partially updated. As such, this change was reverted.

However, there are instances which require for partial updating of the active editor tab, such as in mission control, when adding `prependValue` and `postpendValue` to programming questions. To satisfy this need, a new `UPDATE_ACTIVE_EDITOR_TAB` action was added which partially updates the active editor tab.

Fixes #2284.

### Type of change

### How to test

Navigate to the assessment workspace and verify that it is able to load programming questions without crashing. Additionally, under mission control, try to prepend/postpend values to a programming question and verify that the functionality is working (such as by writing a program that depends on something defined in the prepended value).


chownces (approved):
Thanks for taking a look at this so quickly!

PR #2289, opened 2022-12-15
Fix Playground test and add type guard

### LINKED ISSUE #2288: Invalid Playground test props
Following #2257, there are additional props that are no longer used by the Playground. However, the test file failed to remove these additional props.

The props (keys) should be removed.

Also, ideally, this should have given a compile time error in order to enhance the developer experience.

RichDom2185:
### Description

Closes #2288. Also adds a type guard in the test file as mentioned in the ideal case for the abovementioned issue.

Creates a new function to act as a "middleware" and guard against invalid keys, while still retaining the original type definitions (unlike explicit typing, which prevents extension to "incrementally" build the final object, or `Partial` types and explicit type assignments, which cause issues with "potential" `undefined`s).

This function should be ideally adopted into the source code incrementally (issue for adoption to be created after this PR is merged).

### Type of change

### How to test

NIL

### Checklist

- [x] I have tested this code


RichDom2185:
@chownces I honestly doubt the usefulness of the test snapshot that I updated. It does not seem to check the actual rendering behaviour, and only the props? Not too sure how this works.

chownces (approved):

chownces:
> @chownces I honestly doubt the usefulness of the test snapshot that I updated. It does not seem to check the actual rendering behaviour, and only the props? Not too sure how this works.

agree. our frontend is lacking proper UI testing for actual rendering behaviour/ user interactions. the snapshots are not useful in this sense

it might be worth to look into testing with mocked interactions for more comprehensive automated testing, but most projects do not have enough time to do this. right now we are getting by with a simple QA before every sem

perhaps we can raise this at our next meeting? and see whether this is worth the engineering effort

PR #2291, opened 2022-12-17
Add editor container in preparation of multiple editor tabs

ianyong:
### Description

With #2278 merged in, it is now possible to add a container component for instances of the `Editor`. This container component will be responsible for all tab switching logic in the future. Currently, it does nothing.

Note that Sourcecast & Sourcereel make use of their own custom editor but pass in the custom editor itself as prop to `Workspace`/`MobileWorkspace`, instead of passing in the props themselves. I will need to refactor these in order to make them work with the `EditorContainer`.

### Type of change

### How to test


ianyong:
Since there is no change in behaviour, I have targeted the `master` branch for this PR. This is so as to reduce the probability of merge conflicts as much as possible.

src/commons/editingWorkspace/EditingWorkspace.tsx
```diff
@@ -166,22 +167,29 @@ class EditingWorkspace extends React.Component<EditingWorkspaceProps, State> {
...
+                      question.editorValue ||
+                      (question as IProgrammingQuestion).solutionTemplate
+                  };
+                }),
               editorSessionId: '',
-              editorValue:
-                // TODO: Hardcoded to make use of the first editor tab. Rewrite after editor tabs are added.
-                this.props.editorTabs[0].value ||
-                question.editorValue ||
-                (question as IProgrammingQuestion).solutionTemplate,
```
ianyong, line 177: Note: the extra `map` is simply to maintain the current behaviour of only replacing the first editor tab.

zhaojj2209 (approved):

PR #2292, opened 2022-12-17
Handle editor variants inside EditorContainer & align editor event listener types

ianyong:
### Description

The `Sourcecast` & `Sourcereel` workspaces make use of the `SourcecastEditor` instead of the `Editor`. To allow for this, both `Workspace` and `MobileWorkspace` take in an optional `customEditor` prop of type `JSX.Element`, which is rendered instead of the normal editor if the prop is present.

This is however quite problematic as the creation of `SourcecastEditor` is already done before the `Workspace`/`MobileWorkspace` component is called, as opposed to the `Editor` which is created in `Workspace`/`MobileWorkspace`. In order to support multiple editor tabs for both the `Editor` and the `SourcecastEditor` without having to duplicate similar logic, `EditorContainer` was updated to make use of a discriminated union as its prop. I initially tried playing around with generics to get something like this:
```ts
export type EditorContainerProps<T> = Omit<T, keyof EditorTabStateProps> & {
  renderEditor: (editorProps: T) => void;
  editorTabs: EditorTabStateProps[];
};
....
const EditorContainer = <T extends EditorTabStateProps>(props: EditorContainerProps<T>) => {
  const { editorVariant, editorTabs, ...editorProps } = props;
  // TODO: Implement editor tabs.
  const editorTabProps: T = {
    ...editorProps,
    ...editorTabs[0]
  };
  return renderEditor(editorTabProps);
```
Unfortunately, TypeScript is unable to infer the type of `editorTabProps` correctly here unless I perform type casts (which are evil). On top of that, using generics results in all of the parent states also becoming generic, which is rather ugly as even the top-level `WorkspaceState` requires a type variable to represent the editor variant.

By making use of discriminated unions, we can remove the `customEditor` prop in `Workspace` and `MobileWorkspace`. Regardless of which editor variant should be used, `EditorContainer` will be rendered. This allows the future multiple editor tabs logic to work for both kinds of editors, as well as makes the handling of both editor variants consistent.

Additionally, I also aligned the event listener prop type signatures between `SourcecastEditor` and `Editor` so that `enableMobileKeyboardForEditor` can be used for both editor variants in `MobileWorkspace`.

### Type of change

### How to test

### DIFF
--- a/src/commons/editor/EditorContainer.tsx
+++ b/src/commons/editor/EditorContainer.tsx
@@ -1,3 +1,7 @@
+// Necessary to prevent "ReferenceError: ace is not defined" error.
+// See https://github.com/securingsincity/react-ace/issues/1233 (although there is no explanation).
+import 'ace-builds/src-noconflict/ace';
+
 import _ from 'lodash';
 import React from 'react';
 
--- a/src/commons/editor/EditorContainer.tsx
+++ b/src/commons/editor/EditorContainer.tsx
@@ -1,12 +1,26 @@
 import _ from 'lodash';
 import React from 'react';
 
+import SourcecastEditor, {
+  SourceRecorderEditorProps
+} from '../sourceRecorder/SourceRecorderEditor';
 import { EditorTabState } from '../workspace/WorkspaceTypes';
 import Editor, { EditorProps, EditorTabStateProps } from './Editor';
 
-export type EditorContainerProps = Omit<EditorProps, keyof EditorTabStateProps> & {
+export type NormalEditorContainerProps = Omit<EditorProps, keyof EditorTabStateProps> & {
+  editorVariant: 'normal';
   editorTabs: EditorTabStateProps[];
 };
+
+export type SourcecastEditorContainerProps = Omit<
+  SourceRecorderEditorProps,
+  keyof EditorTabStateProps
+> & {
+  editorVariant: 'sourcecast';
+  editorTabs: EditorTabStateProps[];
+};
+
+export type EditorContainerProps = NormalEditorContainerProps | SourcecastEditorContainerProps;
 
 export const convertEditorTabStateToProps = (editorTab: EditorTabState): EditorTabStateProps => {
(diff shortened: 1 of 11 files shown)


ianyong:
Since there is no change in behaviour, I have targeted the `master` branch for this PR. This is so as to reduce the probability of merge conflicts as much as possible.

src/commons/editor/EditorContainer.tsx
```diff
@@ -1,24 +1,61 @@
+// Necessary to prevent "ReferenceError: ace is not defined" error.
+// See https://github.com/securingsincity/react-ace/issues/1233 (although there is no explanation).
+import 'ace-builds/src-noconflict/ace';
```
ianyong, line 3: To note: this thing is very weird and I don't actually know the reason why this import statement is needed for the error to not appear. My guess is that there's some weird load order issue going on, although that doesn't explain why this import fixes it.

src/commons/editor/EditorContainer.tsx
```diff
@@ -1,24 +1,61 @@
...
 const EditorContainer: React.FC<EditorContainerProps> = (props: EditorContainerProps) => {
-  const { editorTabs, ...editorProps } = props;
+  let createEditorTab;
+  if (props.editorVariant === 'sourcecast') {
+    const { editorVariant, editorTabs, ...editorProps } = props;
+    createEditorTab = createSourcecastEditorTab(editorProps);
+  } else {
+    const { editorVariant, editorTabs, ...editorProps } = props;
+    createEditorTab = createNormalEditorTab(editorProps);
+  }
```
ianyong, line 56: The code is written as such so that TypeScript can automatically infer and narrow the types for us. I've tried to write it more succinctly, but TypeScript was unable to narrow the types automatically (meaning that type casting would be needed). If you can find a more succinct way of writing this, do let me know.

zhaojj2209 (approved):
LGTM!

PR #2302, opened 2023-01-06
Remove `externalLibs`

shenyih0ng:
### Description

This PR removes deprecated scripts under `public/externalLib` and the notion of `ExternalLibrary` in favor of the current modules systems.

### Context

Scripts under `externalLib` were used to load libraries (e.g. `rune`, `pix_n_flix`) before the introduction of modules. Library functions will loaded directly into the browser global context ('window'), allowing `js-slang` to reference it directly, as opposed to the current module import system which adds to a local `js-slang` context.

### What has been done

- External libs removed
    - [x] rune
    - [x] video (aka pix_n_flix)
    - [x] sound
        - **except `ToneMatrix` UI component (`/externalLib/sound/soundToneMatrix.js`)** 
    - [x] binary_tree
    - [x] list
    - [x] face_api (machine learning)

- Keep EV3 as `ExternalLibraryName`
    - EV3 has no existing scripts under `/externalLib` but it still remains as a `ExternalLibraryName` which remote execution depends on

    https://github.com/source-academy/frontend/blob/dd28170d2f87fe3b74ead04c4c170e04e9c8ba21/src/features/remoteExecution/RemoteExecutionTypes.ts#L98-L102

- Removed deprecated `SideContent` UI components
    - `SideContentVideoDisplay` for legacy `pix_n_flix`

- Removed external library selector UI component (`ControlBarExternalLibrarySelect`)

### To be done

- [X] Test `ToneMatrix` sound mission
    - Had to keep `SideContentToneMatrix` UI component since tone matrix still depends on it, will migrate over to `/modules` in a new PR
- [X] Remove redundant mocks


martin-henz:
If I remember right, there is one use case of external libs left: tone matrix. @shenyih0ng did you come across the tone matrix in your work on this? Must be hardwired somewhere.

Also: @boydanderson @Sanka-R

shenyih0ng:
> If I remember right, there is one use case of external libs left: tone matrix. @shenyih0ng did you come across the tone matrix in your work on this? Must be hardwired somewhere.
> 
> Also: @boydanderson @Sanka-R

@martin-henz initially i removed it as i found `tabs/SoundMatrix` in our modules system but after taking another look, it still depends on external libs using reference to `window`

i will restore the `ToneMatrix` for now but i believe its possible to migrate the tone matrix over to modules as well. I will make a PR for that and link it here

src/commons/mocks/AssessmentMocks.ts
```diff
@@ -296,17 +296,17 @@ const mockSoundLibrary: Library = {
...
 
 const mockBinaryTreeLibrary: Library = {
   chapter: Chapter.SOURCE_4,
   external: {
-    name: ExternalLibraryName.BINARYTREES,
-    symbols: externalLibraries.get(ExternalLibraryName.BINARYTREES)!
+    name: ExternalLibraryName.NONE,
+    symbols: externalLibraries.get(ExternalLibraryName.NONE)!
   },
   globals: mockGlobals
```
RichDom2185, line 311: Would this result in redundant test cases? Also, the names of the objects can be changed accordingly.

RichDom2185 (commented):
Looks good! Just a minor comment about the test cases. Otherwise, I don't have access to a tone matrix quest to test it with, but I did test on EV3 and everything still works! Thanks for this!

src/commons/mocks/AssessmentMocks.ts
shenyih0ng: yes these will become redundant test cases, im going to remove them

RichDom2185 (commented):
Besides the unused import causing a CI error, looks good! Thanks!

RichDom2185 (approved):
LGTM, thanks!

PR #2304, opened 2023-01-10
Revert "Fix REPL representation bug"

chownces:
Reverts source-academy/frontend#2262

This PR introduced a bug in the stepper, let's revert this commit first and reopen the issue. Perhaps the next stepper team can take a look at that.


chownces:
Closing in favor of #2305

PR #2312, opened 2023-01-17
Update to latest Node.js compatible version

Ruizhi2001:


chownces:
Let's take a closer look at this. We need to address our dependency issues/ upgrading asap

`yarn start` issue:
#2313

PR #2328, opened 2023-01-21
Modify readme

Ella-e:
### Description

Add information about current working version of node for building frontend locally. The latest LTS version v18.13.0 doesn't work with my local environment. The suggestion is to switch to a lower LTS version, such as v16.16.0

### Type of change

Update the README.md document .

### How to test

Check the README.md file.


martin-henz (approved):
Great. Nice!

PR #2331, opened 2023-01-29
Fix unnecessary rerendering of side content in desktop workspace

ianyong:
### Description

Passing object literals as props to React components causes unnecessary re-renders due to objects being compared by reference. As such, even if the object prop being passed in is semantically the same as the previous prop, it will trigger a re-render if it is not referentially the same object.

We prevent unnecessary re-renders by performing a deep comparison between the previous & next props state when determining whether a component should be re-rendered or not. This works for the desktop `Workspace` component, **but not for the `MobileWorkspace` component as the `mobileControlBarProps` that it takes in is actually changing semantically**, but I'm not sure why.

Even though memoization does not fully resolve the issue for the `MobileWorkspace` component, it's memoized anyway because it helps a bit (specifically, does not re-render due to the `tabs` object literal that is being passed as prop) and also for consistency with the `Workspace` component, which can be seen as its equivalent component.

Partially resolves https://github.com/source-academy/modules/issues/168.

### Type of change

### How to test

Replicate the steps in the linked issue.


martin-henz (approved):
Looks good and resolves the desktop issue. Will change the Issue to be "mobile" specific.

RichDom2185 (approved):
Personally, I think it would be better if we keep the `SideContent` and `MobileSideContent` components as they are, and just `export default React.memo(SideContent, propsAreEqual);` instead, but this is a stylistic thing, I'll leave it to you to decide. @chownces @shenyih0ng what do you think? I think it would be nicer the components themselves not have any logic related to re-rendering

Otherwise, LGTM! Thanks for the quick fix!

ianyong:
> Personally, I think it would be better if we keep the `SideContent` and `MobileSideContent` components as they are, and just `export default React.memo(SideContent, propsAreEqual);` instead, but this is a stylistic thing, I'll leave it to you to decide. @chownces @shenyih0ng what do you think? I think it would be nicer the components themselves not have any logic related to re-rendering

This makes sense, thanks for the suggestion! Possibly can be included in the fix for the mobile workspace.

martin-henz:
> This makes sense, thanks for the suggestion! Possibly can be included in the fix for the mobile workspace.

Great. Who can do the mobile fix?

PR #2333, opened 2023-01-31
Remove infinite loop telemetry

ianyong:
### Description

Discussed with Prof @martin-henz. The infinite loop telemetry no longer works as js-slang now operates on ASTs instead of code. Since the research project that the data being collected for is now over, the infinite loop telemetry should be removed. However, the agreement to participate in educational research feature is left entirely intact because it might be used again in the future (and is worded to include all educational research, not just infinite loop detection).

Necessary for the frontend to work with js-slang after https://github.com/source-academy/js-slang/pull/1343.

Part of #2176.

### Type of change

This PR does not fit any of the below...

### How to test

No change in behaviour except there is no longer any infinite loop telemetry.


ianyong:
@martin-henz This PR does not require bumping the js-slang version, but bumping the js-slang version requires this PR.

martin-henz (approved):
As discussed. Good to remove infinite loop detection until anyone wants to get back to this research topic.

PR #2334, opened 2023-01-31
Improve performance when applying filter over a large dataset

zhyuhan:
### Description

When trying to apply a quick filter through the filter text field, each change in the input value will cause a render of the `Grading` component, since the change will update the `filterValue` field in the state.

Since all calls to `gridApi.setQuickFilter` are from the input field itself, there is actually no need to keep `filterValue` in the component state, and the filter text can be passed from `event.target.value` directly.

This improves performance greatly, especially when the set of grading data is very large (often the case for lecturers).

### DIFF
--- a/src/pages/academy/grading/Grading.tsx
+++ b/src/pages/academy/grading/Grading.tsx
@@ -388,7 +388,6 @@ class Grading extends React.Component<GradingProps, State> {
   };
 
   private applyFilter = (filter: string) => {
-    console.log(filter);
     this.gridApi?.setQuickFilter(filter);
   };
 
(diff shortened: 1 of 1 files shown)


src/pages/academy/grading/Grading.tsx
```diff
@@ -398,6 +387,11 @@ class Grading extends React.Component<GradingProps, State> {
     }
   };
 
+  private applyFilter = (filter: string) => {
+    console.log(filter);
```
shenyih0ng, line 391: nit: remove console log

shenyih0ng (changes requested):
Thanks for this, the rest looks good!

Tested it on locally and it works well

src/pages/academy/grading/Grading.tsx
shenyih0ng, line 391:
```suggestion

```

shenyih0ng (approved):

PR #2338, opened 2023-02-01
Fix unnecessary rerendering of side content in mobile workspace

### LINKED ISSUE #2332: Assessment Workspace: Module tab reloading on each keystroke in editor
# Expected Behavior

Editor keystrokes should not affect module loading.

# Current Behavior

Currently each editor keystroke reloads the currently loaded modules in mobile workspace.

## Steps to Reproduce

1) Load: https://share.sourceacademy.org/oxhil
2) shrink screen to mobile mode
2) press run 
3) open the runes tab
4) start typing in editor

you will see that each keystroke restarts the animation and thus the keyboard entry will be slow.

## Context

This is a critical issue in modules. Fixing it is necessary for current CS1101S.

ianyong:
### Description

Played whack-a-mole with the React profiler until everything that is causing unnecessary re-rending of the side content tabs in the mobile view is memoized in some form or another. **Note that unnecessary re-rendering here is only in the context of typing in the editor.** The side content in the mobile view will definitely still re-render more than is necessary (i.e., more than the desktop view) by virtue of the React component hierarchy described in the paragraph below.

The main difficulty encountered here is that in the desktop workspace, the control bar is not nested in the side content component. As such, both of these components can re-render separately. In the mobile workspace however, the control bar is a child of the side content component. Whenever the control bar has to re-render, so does the side content. An alternative to all of these memoization would be to redesign the mobile workspace such that the control bar is not nested inside the side content component. This would also help prevent future changes from reintroducing this unnecessary re-render issue. However, this is also not an easy refactor to carry out.

Also implements @RichDom2185's [suggestion to memoize components when exporting as default](https://github.com/source-academy/frontend/pull/2331#pullrequestreview-1274097301).

See each individual commit for the context behind every change.

Fixes #2332. Related to #2331.

### Type of change

### How to test

Replicate the steps in the linked issue.


src/pages/playground/Playground.tsx
```diff
@@ -512,16 +549,23 @@ const Playground: React.FC<PlaygroundProps> = ({ workspaceLocation = 'playground
 
   const { handleEditorValueChange } = props;
 
-  // No point memoing this, it uses props.editorValue
```
ianyong, line 515: This turned out to be untrue...

src/pages/playground/Playground.tsx
```diff
@@ -512,16 +549,23 @@ const Playground: React.FC<PlaygroundProps> = ({ workspaceLocation = 'playground
...
-      // TODO: Hardcoded to make use of the first editor tab. Rewrite after editor tabs are added.
-      editorValue={props.editorTabs[0].value}
-      handleSetEditorSessionId={id => dispatch(setEditorSessionId(workspaceLocation, id))}
-      sharedbConnected={props.sharedbConnected}
-      key="session"
-    />
+  const getEditorValue = React.useCallback(
+    () => store.getState().workspaces.playground.editorTabs[0].value,
+    [store]
+  );
```
ianyong, line 555: We do this so that we only access the editor value when the session button is pressed, instead of constantly re-rendering the session button whenever the editor value changes. As a result, we need the `store` object using the `useStore` hook which does not listen to changes (as compared to the `useSelector` hook which does and will force re-renders every time the listened value updates).

src/pages/playground/Playground.tsx
```diff
@@ -512,16 +549,23 @@ const Playground: React.FC<PlaygroundProps> = ({ workspaceLocation = 'playground
...
-    <ControlBarSessionButtons
-      editorSessionId={props.editorSessionId}
-      // TODO: Hardcoded to make use of the first editor tab. Rewrite after editor tabs are added.
-      editorValue={props.editorTabs[0].value}
-      handleSetEditorSessionId={id => dispatch(setEditorSessionId(workspaceLocation, id))}
-      sharedbConnected={props.sharedbConnected}
-      key="session"
-    />
+  const getEditorValue = React.useCallback(
+    () => store.getState().workspaces.playground.editorTabs[0].value,
```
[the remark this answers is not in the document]
ianyong: Thanks for the catch!

RichDom2185 (approved):
LGTM, thanks!

PR #2340, opened 2023-02-07
Refactor & document insertion of debugger statements into code

ianyong:
### Description

Extracts out the logic for inserting debugger statements into code as a separate function & documents the context behind why this step is being done. This change stemmed from an offline discussion with Prof @martin-henz on why the frontend is calling the `parse` function from js-slang (relevant issue: https://github.com/source-academy/js-slang/issues/1349#issuecomment-1413186253).

Related to #2339 and #2176.

### Type of change

### How to test

Check that breakpoints still function correctly. There should be no change in behaviour.


martin-henz (approved):
Nice and clear description: good that this is now properly documented

PR #2341, opened 2023-02-10
stepper's step limit restricted to even number

Ella-e:
### Description

Summary of change:
1. StepSize of stepper's step limit changed to 2.
2. auto-scale the odd step limit into the next even integer when the input box lost focus.

Motivation:
From the discussion in issue #1345, we decided that it would be more consistent with the current definition of stepper and makes the user experience less confusing to restrict the change of step limit to even numbers at the frontend.

### DIFF
--- a/src/commons/controlBar/ControlBarStepLimit.tsx
+++ b/src/commons/controlBar/ControlBarStepLimit.tsx
@@ -16,14 +16,14 @@ type StateProps = {
 };
 
 export const ControlBarStepLimit: React.FC<ControlBarStepLimitProps> = props => {
-   /**
-   * Scale the odd stepLimit into the next even integer. 
+  /**
+   * Scale the odd stepLimit into the next even integer.
    * Triggered when the input box lost focus.
    */
-   const onBlurAutoScale = ()=>{
+  const onBlurAutoScale = () => {
     props.handleOnBlurAutoScale != undefined
-    ? props.handleOnBlurAutoScale(Number(props.stepLimit))
-    : console.log('undefined');
+      ? props.handleOnBlurAutoScale(Number(props.stepLimit))
+      : console.log('undefined');
   };
 
   return (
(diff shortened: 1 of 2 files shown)


src/commons/controlBar/ControlBarStepLimit.tsx
```diff
@@ -15,6 +16,16 @@ type StateProps = {
...
 
 export const ControlBarStepLimit: React.FC<ControlBarStepLimitProps> = props => {
+  /**
+   * Scale the odd stepLimit into the next even integer.
+   * Triggered when the input box lost focus.
+   */
+  const onBlurAutoScale = () => {
+    props.handleOnBlurAutoScale !== undefined
+      ? props.handleOnBlurAutoScale(Number(props.stepLimit))
+      : console.log('undefined');
```
martin-henz, line 26: Not sure about this console.log: Is it for your own debugging? The console log "undefined" isn't helpful for the user (or developer). Can you take another look?

martin-henz (changes requested):

src/commons/controlBar/ControlBarStepLimit.tsx
Ella-e: image
I changed it to use if statement. Is it okey to write like that?

shenyih0ng: a simpler way to do this could be via [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

```javascript
props.handleOnBlurAutoScale?.(...)
```

Ella-e: Okey! Thank you very much. I will change it to optional chaining.

Ella-e: image
changes to optional chaining committed.

martin-henz (approved):

PR #2346, opened 2023-02-16
Shift breakpoints with code insertions & deletions

ianyong:
### Description

Modelled the breakpoint shifting behaviour after popular code editors & IDEs available such as VS Code & JetBrains IDEs. Whenever a change event is fired on the editor, the breakpoints are shifted around based on the delta.

vmconnect_MXqbkeLDWP

Resolves #2345. Fixes #1983. Related to #2176 (editor improvement).

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### How to test

Play around with it and try to break it.

PR #2348, opened 2023-02-18
Fix playground editor bugging out due to unnecessary updates

### LINKED ISSUE #2337: Refactoring of names broken - editor's previous state not accounted for with multicursor/undo/redo
The playgrounds in both the SA and NUS deployments currently have the issues described below. I am unsure if it is related to the merging of PR #2331, which quickly addressed a critical bug with editor keystrokes and state, as I do not know what version each deployment is currently running (ie when the issues started).

Certain actions performed in the editor cause the cursor position(s) and selected text to bug out; I have encountered multicursor behaviour being broken, which in turn breaks the refactoring of multiple name occurrences. Undoing/redoing also does not behave as expected.

I suspect that the editor is unable to properly make use of its previous state from before an affected action was performed. To correctly perform the action, it needs to take the previous cursor(s) and selection(s) into account, thus knowing where the cursor(s) should then be and what text it should then be selecting. The current behaviour is as if the initial application of the action "deletes" all previous text, and replaces it with a "paste" of the new displayed text.

Interestingly, these issues do not occur with the homework editor as opposed to the playground editor. Source/JavaScript language settings also do not appear to affect this. In relation to #2332, this was all tested on desktop.

---

When using multiple cursors and pressing a keystroke, the keystroke is only correctly applied to the last instance of selected text or cursor position in the file. The cursor then gets placed at the end of the file, which is where subsequent applications take place. After that, further keystrokes cause 2 applications at the end of the file.

Refactoring variable name occurrences (Ctrl+M) makes use of multicursor and is thus similarly impacted.

To demonstrate, in the images below, I refactor `variable` by having all 6 occurrences selected with a cursor at the end of each (via Ctrl+M):

image

I then type `name`. The last instance of selected text is replaced by `n`, the cursor gets placed at the end of the file, and the remaining 5 `n`s are placed there. The rest of the letter keystrokes are applied only twice, in the same place:

image

It is instead expected that the cursors remain in place, and the keystrokes replace the selected text or are inserted at the cursor(s). Here is a depiction from the NUS deployment's homework editor:

image

---

When undoing (Ctrl+Z) or redoing (Ctrl+Shift+Z), all the text in the editor gets selected and the cursor is placed at the end of the file. This behaviour bears resemblance to how multicursor is being impacted.

Initial cursor position and text:

image

After undo:

image

After redo:

image

It is instead expected that the cursor jumps to the specific place where text was removed, or selects the inserted text with the cursor placed at its end, as the case may be:

image

image

ianyong:
### Description

This bug's existence negatively affects the CS4215 homework experience because undoing/redoing sends you to the bottom of a 1000+ line program, among other annoying issues...

kGqRRF8OcW

I was unable to figure out what was causing the bug from just the profiler alone, so I had to binary search the git history to find the offending PR. Then, I binary searched the git history of the unsquashed branch to find that the offending commit is https://github.com/source-academy/frontend/commit/037d9c957e22a8097660b5f76a214553fddebf04.

Because some of these callback functions being passed into child components used to be declared in the `PlaygroundContainer`, the `Playground` re-rendering would not cause these functions to be recreated. After the refactoring in #2257, this was no longer the case and every re-render of the `Playground` meant that these callback functions would be recreated. To fix this issue, I memoize these callback functions using the `useCallback` hook.

Fixes #2337.

### Type of change

### How to test

See the linked issue for steps to replicate. Alternatively, see https://github.com/source-academy/frontend/pull/2346#issuecomment-1433319375 to test with the new shifting breakpoints.

### DIFF
--- a/src/pages/playground/Playground.tsx
+++ b/src/pages/playground/Playground.tsx
@@ -780,15 +780,25 @@ const Playground: React.FC<PlaygroundProps> = ({ workspaceLocation = 'playground
     ..._.pick(props, 'editorSessionId', 'isEditorAutorun'),
     editorVariant: 'normal',
     editorTabs: props.editorTabs.map(convertEditorTabStateToProps),
-    handleDeclarationNavigate: (cursorPosition: Position) =>
-      dispatch(navigateToDeclaration(workspaceLocation, cursorPosition)),
+    handleDeclarationNavigate: React.useCallback(
+      (cursorPosition: Position) =>
+        dispatch(navigateToDeclaration(workspaceLocation, cursorPosition)),
+      [dispatch, workspaceLocation]
+    ),
     handleEditorEval,
-    handlePromptAutocomplete: (row: number, col: number, callback: any) =>
-      dispatch(promptAutocomplete(workspaceLocation, row, col, callback)),
-    handleSendReplInputToOutput: (code: string) =>
-      dispatch(sendReplInputToOutput(code, workspaceLocation)),
-    handleSetSharedbConnected: (connected: boolean) =>
-      dispatch(setSharedbConnected(workspaceLocation, connected)),
+    handlePromptAutocomplete: React.useCallback(
+      (row: number, col: number, callback: any) =>
+        dispatch(promptAutocomplete(workspaceLocation, row, col, callback)),
+      [dispatch, workspaceLocation]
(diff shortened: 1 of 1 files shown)


martin-henz (approved):
Looks great!

PR #2355, opened 2023-02-23
bumping version of js-slang

martin-henz:
bumping js-slang to 1.0.13

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "classnames": "^2.3.2",
     "flexboxgrid": "^6.3.1",
     "flexboxgrid-helpers": "^1.1.3",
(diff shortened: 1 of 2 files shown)


martin-henz:
As discussed with the Env Viz team, we currently have some test cases failing, and therefore we cannot merge this PR.

@zhaojj2209 this means that the js-slang-to-frontend pipeline is a bit blocked right now. The Env Viz team will try to take a look at the failing test cases. We are hopeful that we can unblock the pipeline (even if we need to disable some Env Viz test cases temporarily).

martin-henz:
obsolete; bumping to 1.0.14

PR #2360, opened 2023-02-25
Remove extraneous dependencies

RichDom2185:
### Type of change

### Checklist

- [x] I have tested this code


package.json
```diff
@@ -89,33 +87,28 @@
     "@testing-library/react": "^12.1.2",
     "@testing-library/user-event": "^14.3.0",
     "@types/acorn": "^6.0.0",
-    "@types/common-tags": "^1.8.1",
-    "@types/draft-js": "^0.11.9",
     "@types/enzyme": "3.10.12",
-    "@types/enzyme-adapter-react-16": "^1.0.6",
-    "@types/estree": "1.0.0",
```
ianyong, line 96: I'm not sure if this should be removed. The ESTree types are used in `src/commons/mocks/ContextMocks.ts`. The reason why everything still works is because the package is installed anyway as a dependency of other dependencies.

martin-henz (approved):
Build successful so obviously these are unnecessary dependencies

package.json
RichDom2185: Hmm I'm not sure if we really should be keeping track of these explicitly, since we are not directly referencing them. I'm thinking that we should leave it to the other dependencies to manage which version of their dependencies are used.

Since removing node_modules and reinstalling all the dependencies does not cause issues, I think they're still safe to remove. I guess, since it has already been merged, we can just leave this here and can always revert later if it turns out to cause issues.

ianyong: > Hmm I'm not sure if we really should be keeping track of these explicitly, since we are not directly referencing them. I'm thinking that we should leave it to the other dependencies to manage which version of their dependencies are used.

We are!

https://github.com/source-academy/frontend/blob/2437ca6720dc9c3624b69e9abf380100d98f86ed/src/commons/mocks/ContextMocks.ts#L2

> Since removing node_modules and reinstalling all the dependencies does not cause issues, I think they're still safe to remove. I guess, since it has already been merged, we can just leave this here and can always revert later if it turns out to cause issues.

The main reason why I think dependencies that we directly use should be listed in our `package.json` is that if we rely on them being available transitively through other direct dependencies, we face the following issues:
1. If the direct dependency updates our transitive dependency to a version with breaking changes; or
1. If the direct dependency drops our transitive dependency,

our codebase will break. Since the breaking changes happened indirectly, it would be harder to establish cause and effect than if we managed our direct dependencies directly.

The silver lining here is that `@types/estree` is heavily using in js-slang and does not update often, if at all. Since the frontend will probably always depend on js-slang, there is almost no chance of any issues arising here - my concern is more of an ideological one (i.e., what good practice should be).

RichDom2185: I see, that makes sense, thanks for the input! I will add this line back in a future PR 😄

PR #2364, opened 2023-02-27
Make the retrieval of breakpoint line numbers more robust

### LINKED ISSUE #2362: Breakpoints in the playground editor not restored properly from localstorage
Steps to replicate:
1. Go to the playground.
1. Set a breakpoint.
1. Refresh the page.
1. All lines up to the line with the greatest line number that had a breakpoint added will have breakpoints set.

Most likely broken by #2356.

ianyong:
### Description

The JSON-serialised representation of the breakpoints array in the localstorage is unable to distinguish between empty slots, and slots where the value is `null`. As a result, after serialising into JSON and deserialising back to JavaScript, the empty slots of an array are now filled with `null`. This breaks the invariant that only line numbers with a breakpoint are non-empty slots that was previously relied on in `getBreakpointLineNumbers`.

~~To resolve this issue, we now check that the value corresponding to the line number is the string `'ace_breakpoint'` to determine whether the line number has a breakpoint set or not.~~
To resolve this issue, we now define a breakpoint as follows:
> Breakpoints are elements in the array with the value 'ace_breakpoint' where the associated line number is the index of the element in the array.

By making use of array semantics instead of object semantics, we have a cleaner definition & implementation.

Fixes #2362.

### Type of change

### How to test

Follow the steps to replicate in the linked issue.

### DIFF
--- a/src/commons/editor/Editor.tsx
+++ b/src/commons/editor/Editor.tsx
@@ -145,27 +145,18 @@ const makeHandleGutterClick =
  * Returns an array of breakpoint line numbers from the Ace Editor's breakpoint
  * array representation.
  *
- * In JavaScript, arrays are just objects. As a result, the Ace Editor's breakpoint
- * array has empty slots (i.e., array indices which are not defined at all). Breakpoints
- * are represented in the Ace Editor's breakpoints array as key-value pairs where the
- * key is a string representation of the line number it is on, and the value is
- * 'ace_breakpoint'.
- *
- * However, because the JSON-serialised representation of empty slots in a JavaScript
- * array is the same as that of the 'null', we cannot simply parse all keys of the object
- * as integers to get the breakpoint line numbers. Instead, we need to:
- * (1) Find all entries in the object which has 'ace_breakpoint' as its value; then
- * (2) Parse the keys of all such entries to integers to get our breakpoint line numbers.
+ * Breakpoints are elements in the array with the value 'ace_breakpoint' where
+ * the associated line number is the index of the element in the array.
  *
  * @param breakpoints The Ace Editor's breakpoint representation.
  */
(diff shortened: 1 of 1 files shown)


src/commons/editor/Editor.tsx
```diff
@@ -149,13 +149,26 @@ const makeHandleGutterClick =
  * array has empty slots (i.e., array indices which are not defined at all). Breakpoints
  * are represented in the Ace Editor's breakpoints array as key-value pairs where the
  * key is a string representation of the line number it is on, and the value is
- * 'ace_breakpoint'. As such, to get an array of breakpoint line numbers, we need to
- * get the keys of the Ace Editor's breakpoint array/object and parse them as integers.
+ * 'ace_breakpoint'.
+ *
+ * However, because the JSON-serialised representation of empty slots in a JavaScript
+ * array is the same as that of the 'null', we cannot simply parse all keys of the object
```
martin-henz, line 155: Actually, accessing the value of an empty array slot results in undefined, not null. The fix is still valid I think, but the comment should change to avoid confusion.

martin-henz (changes requested):

src/commons/editor/Editor.tsx
ianyong: I was previously accessing the array as an object based on the observation that empty slots in an array don't even exist as key-value pairs in the underlying object. After thinking about it more from your comment though, I realised that since we now have to check the value of elements in the array anyway, it no longer makes sense to think in terms of object semantics when we can just make use of array semantics for a cleaner implementation. This also means that most of the comment can be removed.

Thanks for bringing this up! I'll make the change (for both code and comment).

martin-henz (approved):
Great.

PR #2375, opened 2023-03-12
Expand sidebar by default

ianyong:
### Description

As suggested by Prof @martin-henz, it would be sensible to expand the sidebar by default. This is so that when users enable the multiple files mode, they do not have to click two buttons (the multiple files mode button & the files sidebar tab).

In the process, I also fixed a latent bug in the sidebar width logic that surfaced due to this change. This made me realise that I could simplify the logic to make it both more readable and robust. Instead of setting the width of the sidebar whenever the `expandSideBar` and `collapseSideBar` functions are called, the responsibility of setting the width is now part of the sidebar's `Resizable` component.

Part of #2176.

### Type of change

### How to test

The multiple files mode is not yet ready for production and is disabled programmatically. Enable multiple files mode for testing by changing the condition to `false`:
https://github.com/source-academy/frontend/blob/9bc11408b104e29679bfdc025992b241975ea6d7/src/pages/playground/Playground.tsx#L603-L615

In the playground, enable multiple files mode and verify that the sidebar is expanded by default.


zhaojj2209 (approved):
LGTM!

PR #2376, opened 2023-03-12
Upgrade CRACO to v7.1.0 & react-scripts to v5.0.1

### LINKED ISSUE #2213: [Development Only] Error thrown & iframe injected when the application is hot reloaded

### LINKED ISSUE #2313: Upgrade react-scripts
- for Node v18 compatibility (https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported/71334532#71334532)

Alternatively,:
- consider shifting out of CRA
  -  e.g. migrating to vitejs
  - CRA is not as actively maintained anymore

ianyong:
### Description

We are upgrading `react-scripts` to v5 so that it can be used with Node.js 18. This is because [Node.js 16 is reaching end of life this September](https://nodejs.org/ar/blog/announcements/nodejs16-eol/).

Changes:
* Updated CRACO from v6.4.5 to v7.1.0.
* Updated `react-scripts` from v4.0.3 to v5.0.1.
  * This also updates Webpack to v5 which is the cause for many of the changes below...
* Removed selective dependency resolution of `react-error-overlay`.
  * This was a temporary hack that can now be removed due to `react-scripts` being updated. See #2214 for more information.
* Removed duplicate ESLint installation.
  * We need to use `react-script`'s version to prevent conflicts in configuration.
*  Enabled `syncWebAssembly` option for Webpack 5.
  * This is essentially [reverting to the behaviour of Webpack 4](https://webpack.js.org/configuration/experiments/#experiment).
* Regenerated dependency tree (i.e., deleted `yarn.lock` and regenerated it).
  * Upgrading `react-scripts` caused a lot of changes in dependencies. Some of these dependencies need to be hoisted to the top-level of the dependency tree to work, but were not.
  * An undesirable side effect is that the minor versions of other dependencies might have been bumped while doing the regeneration. However, everything still seems to work.
* Forced `@types/react` to use v17 instead of v18.
  * This is because the React 18 types [introduced a lot of breaking changes](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210).
  * The forcing of v17 was done using [selective dependency resolution](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/).
  * We will need to update React (and its types) to v18 in a separate PR. I will create an issue after this PR is merged.
* Polyfill several Node.js core modules.
  * > BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default. This is no longer the case. Verify if you need this module and configure a polyfill for it.
  * Added these polyfills as dev dependencies since they are only needed for Webpack to compile.
* Ignore missing source map warnings for dependencies.
  * Some of our dependencies do not ship with source maps. There's not much we can do about this, so we just disable the warnings if it comes from the `node_modules` directory.
* Specify the `process` and `buffer` polyfills as plugins.
  * `process` is needed for environment variables to work as if they were being used in a Node.js environment (i.e., all of the `process.env.ENV_VAR` that are being used in the codebase).
  * `buffer` is needed for some of our dependencies.

Resolves #2313. Fully fixes #2213.

### Type of change

### How to test


craco.config.js
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
ianyong, line 1: We cannot use ES6 imports for config files.

craco.config.js
```diff
@@ -19,26 +22,61 @@ const cracoConfig = (module.exports = {
...
+      webpackConfig.resolve.fallback = {
+        'child_process': false,
+        'constants': require.resolve('constants-browserify'),
+        'fs': false,
+        'http': require.resolve('stream-http'),
+        'https': require.resolve('https-browserify'),
+        'os': require.resolve('os-browserify/browser'),
+        'stream': require.resolve('stream-browserify'),
+        'timers': require.resolve('timers-browserify'),
+        'url': require.resolve('url/')
```
ianyong, line 51: The trailing slash was part of the instructions given by Webpack. I'm not very sure if removing it will cause any errors.

martin-henz (approved):
Great!

PR #2378, opened 2023-03-19
Add editor tabs

ianyong:
### Description

Note that only the minimum viable features are part of this PR. For example, the ability to shift tabs around as well as to only display the parts of the file path that uniquely identify open files will be added in a future PR. Also, the evaluation of programs is still single file and is hardcoded to make use of the first open editor tab.

image

Part of #2176.

### Type of change

### How to test

The multiple files mode is not yet ready for production and is disabled programmatically. Enable multiple files mode for testing by changing the condition to `false`:
https://github.com/source-academy/frontend/blob/9bc11408b104e29679bfdc025992b241975ea6d7/src/pages/playground/Playground.tsx#L603-L615


zhaojj2209:
Some comments:

- Seems like it might work better visually if the tab colours were to be inverted (so the selected tab would be same colour as the editor, the rest will be darkened).
telegram-cloud-photo-size-5-6104838078012175331-y

- If a file tab is open, the tab will remain even if the file is deleted. Note below that the `/test folder/test.js` tab still exists even though the file no longer exists.
telegram-cloud-photo-size-5-6104838078012175332-y

- When first turning on multiple files mode, `program.js` is not visible in the sidebar. The file will only appear in the sidebar after typing in some code and then reloading the page.
telegram-cloud-photo-size-5-6104838078012175334-y

- Seems that this PR is not backward compatible with existing local storage, resulting in the below error when accessing with a browser that contains the old SA local storage: 
telegram-cloud-photo-size-5-6104838078012175315-y
The site is able to load on incognito, or after clearing all cache and cookies.

ianyong:
Thanks for the detailed review!

> Seems like it might work better visually if the tab colours were to be inverted (so the selected tab would be same colour as the editor, the rest will be darkened).

Okay, I'll make the change.

> If a file tab is open, the tab will remain even if the file is deleted. Note below that the `/test folder/test.js` tab still exists even though the file no longer exists.

This is rather problematic since it's possible that a directory containing many nested directories and files is deleted. All open tabs corresponding to any of the removed files should also be removed. I'll handle this in a separate PR since this feature is disabled on `master` anyway.

> When first turning on multiple files mode, `program.js` is not visible in the sidebar. The file will only appear in the sidebar after typing in some code and then reloading the page.

This would require refactoring the defaults to also bootstrap files in the file system as necessary. I'll resolve this in a separate PR too.

> Seems that this PR is not backward compatible with existing local storage, resulting in the below error when accessing with a browser that contains the old SA local storage:

Unfortunately, deserialising JSON gives `null` for both keys which are missing and keys where the value is `null`. To workaround this issue, I will write an abstraction to store possibly `null` values as an object instead:
```ts
// In JSON, missing keys & keys with the value 'null' are both
// deserialised into 'null'. In order to differentiate between
// the two cases, we wrap nullable values in an object. Missing
// keys would then be deserialised as 'null' while keys with
// the value 'null' would be deserialised as { value: null }.
export type NullableValue<T> = {
  value: T | null;
};
```

zhaojj2209 (approved):
LGTM

PR #2385, opened 2023-03-24
Fix webpack configuration

zhaojj2209:
### Description

SA frontend is currently encountering `Uncaught ReferenceError: webpack_public_path is not defined`. This occurred because the `noParse` parameter was set to be too loose. Fix is to change the attribute to the exact path.

### Type of change

### How to test

Frontend should load properly now.


ianyong (approved):
Works locally now where previously it did not.

PR #2389, opened 2023-03-26
Enable evaluation of multiple file programs

ianyong:
### Description

Refactors the code responsible for evaluating programs to make use of the multiple file-equivalent in js-slang.

Say we have the following function `square` in `/playground/a.js`:
image

We can import it in `/playground/program.js` and invoke it:
image

The output is 25 as expected.

**Note that this PR requires https://github.com/source-academy/js-slang/pull/1378 which removes sloppy uses of the non-null assertion operator on the location of nodes. Without it, any imports will result in an error being thrown.**

Part of #2176.

### Type of change

### How to test

If js-slang has not been bumped to include the changes in https://github.com/source-academy/js-slang/pull/1378, you will also need to link js-slang locally to test this PR.

Note also that local imports/exports are only enabled in Source 2+. I will need to devise some mechanism to deal with potential confusion when people try to run multiple file programs in Source 1.


zhaojj2209:
We will likely need another PR on the js-slang side to modify the Ace editor theme. Specifically, `export` should now be an allowed keyword in Source 2+ (the colour should be orange instead of pink).

Will test more thoroughly once js-slang has been bumped.

ianyong:
> We will likely need another PR on the js-slang side to modify the Ace editor theme. Specifically, `export` should now be an allowed keyword in Source 2+ (the colour should be orange instead of pink).

Oh, that's a good point. Thanks! Let me add it now so that it can be part of the changes when js-slang gets bumped.

zhaojj2209 (approved):
LGTM, tested and works!

PR #2390, opened 2023-03-26
Standardise Acorn version to v8.8.2

ianyong:
### Description

It seems that when linking js-slang to the frontend, the mismatch in the version of Acorn results in the following error (courtesy of @zhaojj2209):
image

More specifically, it is due to the following change: https://github.com/acornjs/acorn/pull/1140.

To resolve this issue, we need to ensure that both js-slang and the frontend use the same version of Acorn.

### Type of change

### How to test

Verify that #2389 works when linking to js-slang with https://github.com/source-academy/js-slang/pull/1380 checked out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@sourceacademy/sling-client": "^0.1.0",
     "@szhsin/react-menu": "^3.2.0",
     "ace-builds": "^1.4.14",
-    "acorn": "^8.8.0",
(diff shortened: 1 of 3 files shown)


ianyong:
@martin-henz We will need to bump js-slang in the frontend together with this PR because the test snapshots will change once that happens.

zhaojj2209 (approved):
LGTM

PR #2391, opened 2023-03-28
bumping js-slang

martin-henz:


ianyong:
Will need to be bumped together with the changes in #2390.

PR #2396, opened 2023-03-30
Add sharing of multi-file programs

ianyong:
### Description

Share links now have the following additional information in order to recreate multi-file programs:
* `isFolder` - whether Folder mode is enabled
* `files` - a compressed, encoded string of a mapping from all file paths to their file contents
* `tabs` - an array of compressed, encoded file paths
* `tabIdx` - the index of the active editor tab

Most importantly, `handleHash` was written in such a way that old share links still work correctly for backward compatibility.

Part of #2176.

### Type of change

### How to test

**Check that both old and new share links work.**


zhaojj2209:
Currently, all files and folders are included in the share link even if the share link is generated while folder mode is disabled. Is this the intended behaviour?

ianyong:
> Currently, all files and folders are included in the share link even if the share link is generated while folder mode is disabled. Is this the intended behaviour?

Yes. The reason is that imports & exports are handled by js-slang, meaning that in the frontend, we do not actually know which subset of files are evaluated as part of the program. It is possible that the user is importing from another local file while in Folder mode, but then disabled Folder mode. In such cases, the program would still work as the other files are merely hidden (instead of being removed). The other reason is that by always synchronising the states that matter (editor tabs, files, etc.), the link sharing behaviour is easier to reason with.

zhaojj2209:
> Yes. The reason is that imports & exports are handled by js-slang, meaning that in the frontend, we do not actually know which subset of files are evaluated as part of the program. It is possible that the user is importing from another local file while in Folder mode, but then disabled Folder mode. In such cases, the program would still work as the other files are merely hidden (instead of being removed). The other reason is that by always synchronising the states that matter (editor tabs, files, etc.), the link sharing behaviour is easier to reason with.

Noted on this. Perhaps it would be good to add an extra disclaimer modal when generating/accessing share links so that users can be aware of this? Listed below are some scenarios where this can potentially be an issue:

- User A creates a file `a.js` when writing code on Source 2+, and later switches back to Source 1. User A then writes some code on Source 1 (`program.js`), generates a share link for the code they have written, and sends it to User B. User B will then be able to access the contents of `a.js` when they switch to Source 2+ and enable folder mode.
- User A creates a file `a.js` when writing code on Source 2+, and later accesses a share link on Source 1 where `program.js` is the only file. Once User A switches back to Source 2+ and enables folder mode, they will find that `a.js` no longer exists, as the entire folder structure has been overwritten when accessing the share link.

martin-henz:
> > Yes. The reason is that imports & exports are handled by js-slang, meaning that in the frontend, we do not actually know which subset of files are evaluated as part of the program. It is possible that the user is importing from another local file while in Folder mode, but then disabled Folder mode. In such cases, the program would still work as the other files are merely hidden (instead of being removed). The other reason is that by always synchronising the states that matter (editor tabs, files, etc.), the link sharing behaviour is easier to reason with.
> 
> Noted on this. Perhaps it would be good to add an extra disclaimer when generating/accessing share links so that users can be made aware of this? Listed below are some scenarios where this can potentially be an issue:
> 
> * User A creates a file `a.js` when writing code on Source 2+, and later switches back to Source 1. User A then writes some code on Source 1 (`program.js`), generates a share link for the code they have written, and sends it to User B. User B will then be able to access the contents of `a.js` when they switch to Source 2+ and enable folder mode.
> * User A creates a file `a.js` when writing code on Source 2+, and later accesses a share link on Source 1 where `program.js` is the only file. Once User A switches back to Source 2+ and enables folder mode, they will find that `a.js` no longer exists, as the entire folder structure has been overwritten when accessing the share link.
> 
> We could perhaps leave this as a future enhancement in a separate PR, given that the existing behaviour of share links is to overwrite everything anyway.

I think the second item isn't surprising. When accessing a share link, the original program (including all files) will disappear.

The first item is a bit arcane. If we can catch such cases, a warning might be good. However, we should avoid general pop-up messages to highlight the possibility of this issue to occur.

zhaojj2209 (approved):
LGTM!

PR #2399, opened 2023-03-31
Handle cases where no editor tabs are open

ianyong:
### Description

It is possible for users to close all editor tabs. After discussion with Prof @martin-henz, we want to add the following behaviours:
* The run button is now disabled when Folder mode is enabled with no open editor tabs. This is because Folder mode relies on the currently open editor tab to determine what the entrypoint of the program should be. Without an open editor tab, there is no program entrypoint and the program cannot be evaluated.
* When the user disables Folder mode without any editor tabs open, a default file (`/program.js`) will be open as an editor tab. If the default file does not exist in the file system, it is created with the default editor value.

Part of #2176.

### Type of change

### How to test

Check that the behaviours described above work.


zhaojj2209 (approved):
LGTM

PR #2413, opened 2023-04-02
Use require provider to load tabs

leeyi45:
In accordance with [this PR](https://github.com/source-academy/js-slang/pull/1388), allow the frontend to use the new module loading system.

There was a persistent issue breakpoints being triggered even when there were none after an errored evaluation. Not sure what the issue was, but changing the code to its current form seems to work.

Reopening from a new branch as the original require-provider branch was very broken


src/commons/sagas/WorkspaceSaga.ts
```diff
@@ -1128,7 +1152,7 @@ export function* evalCode(
   }
 
   // For EVAL_EDITOR and EVAL_REPL, we send notification to workspace that a program has been evaluated
-  if (actionType === EVAL_EDITOR || actionType === EVAL_REPL) {
+  if (actionType === EVAL_EDITOR || actionType === EVAL_REPL || actionType === DEBUG_RESUME) {
```
leeyi45, line 1155: I'm not sure if this is the intended behaviour, but without this when in interpreter and ec-evaluator modes tabs do not spawn

RichDom2185:
Can I delete the old `require-provider` branch?

PR #2425, opened 2023-04-06
Give focus to existing editor tab when opening already opened file

ianyong:
### Description

Currently, trying to open a file that already has an existing editor tab open results in nothing happening, which can be confusing to the user. By making this change, the file that the user is interacting with will always end up at the active editor tab regardless of whether it was already open or not.

### Type of change

### How to test

Try to open a file that is already opened as an editor tab, but is not the active editor tab. Focus should be given to the editor tab corresponding to the file being opened.


martin-henz (approved):
Looks good!

PR #2427, opened 2023-04-06
Source Academy Contest Voting Update - Implementation of Voting Tierlist

kjw142857:
### Description
1. Dependencies required: Backend needs to be running and conform to latest commit so that putting multiple entries in the same tier can be accepted.
2. Summary: This change converts the previous scoring of entries using numeric-based ranking into an interactive tierlist with a graphical user interface.
3. Motivation: The old contest voting system felt like a chore to handle because of the large number of scores needed to be inputted. It also used a ranking-based system and did not accept duplicate score, leading to an entry's score being heavily determined by the other 9 entries it was grouped with for voting, instead of its own merit. In addition, the GUI for voting looked dull and uninspired. Hence, we have changed the voting system to a fairer tier-based voting system, where users can assign each entry to any tier they want without any restrictions. The reduced number of tiers (instead of scores) also simplifies the voting process and makes the GUI more attractive to users. Furthermore, due to the removal of checks on duplicate entries and score boundaries, the numeric-based weightage of the tiers can be altered by anyone seeking to improve the system in the future easily, without compromising security as voters will not be able to enter invalid scores at all. We will be giving users a choice of 5 tiers, S, A, B, C and D to choose from, with the possibility of more tiers such as F/SS in the future.

### Type of change

### How to test
Step 1. Run local version of Source Academy. A course containing at least 2 students and 1 staff is required. At least 11 students is recommended but not compulsory. Both a working frontend and backend are required for this, including a backend version after our other pull request (https://github.com/source-academy/backend/pull/938).
Step 2. Download and extract the Contest Voting test files attached. (Note: You can also use your own contest test files).
Step 3. Access ground control as a staff user and upload the contest-C3.xml file to create and publish a new contest ("Beautiful Runes").
Step 4. Generate contest entry submissions by logging in as the student users and accessing the contest. Minimum of 2 submissions needed.
Step 5. Access ground control again as the staff user and upload the contest-C3-voting.xml file to create and publish a contest voting assignment for the contest.
Step 6. Access as a student user. If n submissions were made in step 4, you should be able to view min(10, n - 1) submissions in the tierlist. Clicking on an entry should enable you to view and run its code on the left hand panel. Dragging and dropping entries should produce the message "Please assign each entry to a tier.", until all entries have been dropped in a tier. Upon all entries being assigned, the message "Saved!" should appear.
Step 7. Return to Academy and finalise the submission. 
Step 8. Next, access the backend database. Run "SELECT * FROM submission_votes ORDER BY updated_at DESC;" You should be able to see your votes with the tier mapped to the corresponding score: S: 10, A: 7, B: 4, C: 2, D: 1.
Step 9. You can make more voting submissions with other student users if desired. Wait for the contest-rolling leaderboard to be updated in the backend. (This happens every 2 hours; however this process can be accelerated by changing the scheduler configuration in /backend/config/config.exs)
Step 10. Log in as a staff user and access the voting assignment. You should be able to see the leaderboard populated with entries with nonzero scores.
Step 11. Access the backend database again and run "SELECT submission_id, relative_score FROM answers WHERE relative_score<>0 ORDER BY updated_at DESC;" You should be able to see the scores corresponding to the leaderboard scores.

[Contest Voting Test Files.zip](https://github.com/source-academy/frontend/files/11169916/Contest.Voting.Test.Files.zip)

### Checklist
- [x] I have tested this code


src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -49,30 +30,172 @@ const SideContentContestVoting: React.FunctionComponent<SideContentContestVoting
...
+      const container = (e.currentTarget as HTMLElement).closest('.item-container');
+      if (container && container.contains(placeholder)) {
+        container.insertBefore(currentDraggedItem as Node, placeholder as Node);
+        container.removeChild(placeholder as Node);
+      }
+    },
+    [currentDraggedItem, placeholder]
+  );
+
+  const tierBoard = ['S', 'A', 'B', 'C', 'D'].map(tier => (
```
RichDom2185, line 147: I think it would be good if we move this array to the top-level as a read-only constant.

```tsx
const TIERS = ['S', 'A', 'B', 'C', 'D'] as const; // Outside the component

const SideContentContestVoting // ...
  // ...
  const tierBoard = TIERS.map(tier => (
  // ...
```

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -49,30 +30,172 @@ const SideContentContestVoting: React.FunctionComponent<SideContentContestVoting
...
+    <div className={'tier'} key={`tier-${tier.toLowerCase()}`} id={`tier-${tier.toLowerCase()}`}>
```
RichDom2185, line 148:
```suggestion
    <div className="tier" key={`tier-${tier.toLowerCase()}`} id={`tier-${tier.toLowerCase()}`}>
```

src/commons/sideContent/SideContentContestVotingContainer.tsx
```diff
@@ -19,30 +19,29 @@ type StateProps = {
 /**
  * Container to separate behaviour concerns from rendering concerns
  * Stores component-level voting ranking state
+ * Removed minScore and maxScore due to invalid scores no longer being possible
```
RichDom2185, line 22: No need for this comment.

src/commons/sideContent/SideContentContestVotingContainer.tsx
```diff
@@ -19,30 +19,29 @@ type StateProps = {
...
 const SideContentContestVotingContainer: React.FunctionComponent<
   SideContentContestVotingContainerProps
 > = props => {
   const { canSave, contestEntries, handleSave, handleContestEntryClick } = props;
   const [isValid, setIsValid] = useState<boolean>(true);
   const [votingSubmission, setVotingSubmission] = useState<ContestEntry[]>([]);
-  const minScore = 11 - contestEntries.length;
-  const maxScore = 10;
+  // const minScore;
+  // const maxScore;
```
RichDom2185, line 31: Let's not keep commented, old code. We won't ever need it again.

src/commons/sideContent/SideContentContestVotingContainer.tsx
```diff
@@ -19,30 +19,29 @@ type StateProps = {
...
   useEffect(() => {
     setVotingSubmission(contestEntries);
   }, [contestEntries]);
 
   /**
-   * Validates input value and clamps the value within the min-max range [1, number of entries].
+   * Validates input value.
    * @param votingSubmission voting scores by user for each contest entry.
    * @returns boolean value for whether the scores are within the min-max range.
+   * Removed minScore and maxScore constraints.
```
RichDom2185, line 41: Ditto. This comment is not needed, since all the code related to `minScore` and `maxScore` should be removed, not commented.

src/commons/sideContent/SideContentContestVotingContainer.tsx
```diff
@@ -19,30 +19,29 @@ type StateProps = {
...
    */
   const isSubmissionValid = (votingSubmission: ContestEntry[]) => {
-    return votingSubmission.reduce((isValid, vote) => {
-      const score = vote.score!;
-      return isValid && score >= minScore && score <= maxScore;
-    }, true);
+    return votingSubmission.reduce((isValid, vote) => isValid, true);
```
RichDom2185, line 44: This line of code seems like it would always return `true`. Is that the intended behaviour? If yes, can you explain the rationale? What's this part trying to do?

src/commons/sideContent/__tests__/SideContentContestVoting.tsx
RichDom2185, line 1: I noticed you're removing all the existing tests as the component is effectively rebuilt from scratch, that's okay, but:

* I'd suggest creating new ones to fit the new component, if possible and if you have the time
* If you don't have the time to create new unit tests, just delete the whole file

RichDom2185 (changes requested):
Hi! Great work on the feature! Noticed a couple things as highlighted in the further comments below. Note that I did not include all the console.log statements that are to be deleted, make sure to delete all of them.

#### Personal opinion on code style

I think while the functionality is there, much of it seems to be "JavaScript-style" instead of "React-style" code, if that makes sense. For example, we don't usually use `let` a lot, neither do we use `document.createElement` inside React, as we usually abstract a lot of this logic into individual React components.

@shenyih0ng @chownces thoughts on this? Should we try to convert the code to "React-style" before merging?

---

Overall great work! I'm requesting some changes to be done (not including the code style opinion above), but they should be relatively minor and do reply in the individual threads if you require more clarifications on the requested changes. Thanks!

shenyih0ng:
> Hi! Great work on the feature! Noticed a couple things as highlighted in the further comments below. Note that I did not include all the console.log statements that are to be deleted, make sure to delete all of them.
> #### Personal opinion on code style
> 
> I think while the functionality is there, much of it seems to be "JavaScript-style" instead of "React-style" code, if that makes sense. For example, we don't usually use `let` a lot, neither do we use `document.createElement` inside React, as we usually abstract a lot of this logic into individual React components.
> 
> @shenyih0ng @chownces thoughts on this? Should we try to convert the code to "React-style" before merging?
> 
> Overall great work! I'm requesting some changes to be done (not including the code style opinion above), but they should be relatively minor and do reply in the individual threads if you require more clarifications on the requested changes. Thanks!

agree with @RichDom2185 on this, let's try to "reactify" this so that it is consistent with the rest of the code base. if possible, let's try to add some basic unit tests as well to ensure we don't face any regressions in the future

kjw142857:
Hi, some update on the suggested changes:
- Console.log statements and unnecessary comments have been removed!
- Style change to React from HTML/JavaScript: WIP! Our aim is to finish it in the week after our finals are over.
- Tests for our tierlist: Also WIP! Tbh, the number of possible scenarios has been greatly reduced as there is now no possible way for students to make an invalid submission based on non-numerical/duplicate scores, and so it might be more challenging to think of possible testcases. One possibility would be to add submissions with duplicate tiers as "successful" testcases and/or add a test that all entries have been dragged into the tierlist before a submission is marked as valid - let me know what you guys think too thanks!

src/commons/sideContent/SideContentContestVotingContainer.tsx
RichDom2185: Resolved in 7a5b1bb5afe0fe7524c3dea5ffc9df35a98c205a, closing this thread.

src/commons/sideContent/SideContentContestVotingContainer.tsx
RichDom2185: Resolved in 7a5b1bb5afe0fe7524c3dea5ffc9df35a98c205a, closing this thread.

src/commons/sideContent/SideContentContestVoting.tsx
RichDom2185: Resolved in 991843b751fb007cdd74aa7fd0e425bc8e1103da, closing this thread.

src/commons/sideContent/SideContentContestVoting.tsx
RichDom2185: Resolved in 991843b751fb007cdd74aa7fd0e425bc8e1103da, closing this thread.

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -49,30 +30,172 @@ const SideContentContestVoting: React.FunctionComponent<SideContentContestVoting
...
+  const handleDragStart: React.DragEventHandler = e => {
+    console.log('Drag Start');
+    setCurrentDraggedItem(e.currentTarget as HTMLElement);
+    e.dataTransfer.effectAllowed = 'move';
+  };
+
+  const handleDragEnd = useCallback(
+    (contestEntry: ContestEntry): React.DragEventHandler =>
+      e => {
+        console.log('Drag End');
```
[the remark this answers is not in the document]
RichDom2185: Resolved in 991843b751fb007cdd74aa7fd0e425bc8e1103da, closing this thread.

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -49,30 +30,172 @@ const SideContentContestVoting: React.FunctionComponent<SideContentContestVoting
...
-    maxScore,
-    minScore
+    handleVotingSubmissionChange
   } = props;
   const [showContestEntries, setShowContestEntries] = useState<boolean>(true);
+  const [currentDraggedItem, setCurrentDraggedItem] = useState<HTMLElement | null>(null);
+  const [placeholder, setPlaceholder] = useState<HTMLElement | null>(null);
+
+  const handleDragStart: React.DragEventHandler = e => {
+    console.log('Drag Start');
```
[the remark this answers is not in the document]
RichDom2185: Resolved in 991843b751fb007cdd74aa7fd0e425bc8e1103da, closing this thread.

RichDom2185:
> and/or add a test that all entries have been dragged into the tierlist before a submission is marked as valid

This sounds good :) Looking forward to the tests after your exams are over! All the best for your papers!

RichDom2185:
Hi, can I check on the status of this PR? It's okay if you have not had the time to work on it, but is there an ETA to finish this by?

kjw142857:
Hi, we would be planning to finish the style changes and test cases by this Sunday!

kjw142857:
Hi, just an update on our overall progress - we are currently in the midst of finalising some the style changes to React! In particular, we have found it somewhat challenging to replace document.createElement with a React-based method of creating HTML elements -- perhaps some seniors with experience could guide us on this (many thanks)!

If the above is realised, we should be able to update our pull request to its final version (i.e. ready for merging) by the middle of this week. We are really looking forward to the next batch being able to try out the new contest voting system and hope that it will be an exciting new experience for students and educators alike
photo1681049260
!

RichDom2185:
> we have found it somewhat challenging to replace document.createElement with a React-based method of creating HTML elements

In React, perhaps due to the syntax, we often think of `<Element/>` as an element, but you can actually think of it as a "function call" that creates an element instead.

E.g. for `createPlaceholder`, you can just create a React Functional Component:

```tsx
const Placeholder: React.FC = () => <div className='placeholder'/>
```

Of course, in terms of when to render the placeholder, the logic needs to be reworked a bit. Currently, it seems to be a bit backwards. You are using `e.currentTarget.classList.contains('item-container'`, basically means the created elements (which is supposed to be the result) is what is "dictating" the logic/state of React.

Why not use `useState`, and then have the thing you're dragging use `className={someStateVariable ? 'item-container' : ''}` to dynamically add the `item-container` class depending on the state? You can then use this state variable for conditional rendering of the `Placeholder` component.

This way, then React's logic is what is dictating the state of the rendered components (e.g. what classes the elements contain, whether they are shown or not, etc.), which is what should be the case.

kjw142857:
Tierlist 2.0 major update: we have addressed the suggested changes and also made some improvements to our tierlist!

Code quality improvements:
- Removed the need for the placeholder, and hence unnecessary non-React style statements such as 'let' and 'createElement' from our code.
- Rewritten the test cases! The SideContentContestVoting test file will now test if the tiers render properly upon rendering the ContestVotingContainer, and also will simulate dragging and dropping entries to test if the submission saves only  when all entries are assigned to a tier.
- tierBoard is now memoised along with ContestEntryCards.
- Saved the CSS styles in an array called TIERS in SideContentContestVoting.

Bug fixes/UI Improvements:
- Fixed a bug where all entries went back into the entry bank upon exiting and re-entering the assessment.
- Fixed a bug where dragging an entry into a tier but over another entry would cause it not to enter that tier.
- Dragging an entry over a container now causes it to highlight with a white border.

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,183 @@ type StateProps = {
...
+        id={`tier-${tier.name.toLowerCase()}`}
+      >
+        <h2 style={{ backgroundColor: tier.color }}>{tier.name}</h2>
+        <div
+          className="item-container"
+          onDragOver={handleDragOver}
+          onDragLeave={handleDragLeave}
+          onDragEnter={handleDragEnter}
+          onDrop={handleDrop}
+        ></div>
```
RichDom2185, line 117: Since there's no children, let's clean this up a little by making the div self-closing.

```suggestion
        />
```

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,183 @@ type StateProps = {
...
+          );
+          const bankElement = document.querySelector(`#bank .item-container #item-${index + 1}`);
+          if (tierElement && bankElement) {
+            tierElement.appendChild(bankElement);
+          }
+        }
+      }
+    });
+  }, [contestEntryCards, sortedContestEntries]);
+
```
RichDom2185, line 196: I'm not entirely sure what this `useEffect` is doing. Do you mind explaining?

RichDom2185 (commented):
Hi, thanks for the hard work to refactor everything! My comments below:

src/commons/sideContent/SideContentContestVoting.tsx
kjw142857: Resolved!

src/commons/sideContent/SideContentContestVoting.tsx
kjw142857: Hi, this useEffect has 2 main purposes:
1. Enabling the contestVoting to "save" the tiers of the contest entries. Previously, the cards would automatically be assigned to the bank upon re-opening the assignment, even if the user already had a saved submission. Thus, we came up with a method to automatically map entries back to their previously saved tier upon reopening. We wanted to implement it in contestVotingContainer (as it is the pre-render score check) but this was not possible due to the fact that a card (the bankElement) needed to be rendered _first_ before it could be automatically assigned to the tier it was previously saved in by the user. Hence, we implemented it to take effect upon a change to contestEntryCards (i.e. the initial build of contestEntryCards would trigger it), which would be after the creation of the cards but before the render of SideContentContestVoting, i.e. at just the right time so that on first opening, the viewer would see the correctly saved submission that they had inputted earlier.
2. This useEffect also opened the door to migrating legacy submissions (using the old score-based system) to the new tier system, and I am happy to say that we have successfully implemented it in the latest upvoted file! Now, all past submissions will show a tier based on the score the user assigned them in the past, using our new function mapTiers which maps 10 -> S; 8, 9 -> A etc, and the useEffect function implemented. This means that if the user reopens a past submission from say AY 22/23 Sem 1, they will be able to see their past vote with the new Tierlist design, and all the entries saved in a tier instead of a numerical score. This is also useful for keeping data from contests using the old system.

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,205 @@ type StateProps = {
...
+    }
+    return (tierScore: number) => false;
+  };
+
+  // Renders contest entries in the correct tier using saved score upon initial render
+  useEffect(() => {
+    sortedContestEntries.forEach((entry, index) => {
+      if (entry.score !== undefined && entry.score !== null) {
+        const entryScore = entry.score as number;
+        const tierIndex = TIERS.findIndex(tier => mapTier(entryScore)(tier.score));
```
RichDom2185, line 205: I understand where you are coming from for backwards compatibility, but personally I don't think this is very necessary. @shenyih0ng what do you think (for context, refer to point 2 here: https://github.com/source-academy/frontend/pull/2427#discussion_r1203494196)

I argue performing a database migration would be the right thing to do here (even though that is a non-reversible change). Having said that, I think there's merit to keeping things backwards compatible, although at the expense of additional clutter to the codebase.

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,205 @@ type StateProps = {
...
+        if (tierIndex !== -1) {
+          const tierElement = document.querySelector(
+            `#tier-${TIERS[tierIndex].name.toLowerCase()} .item-container`
```
RichDom2185, line 208: The index here is unnecessary.

```suggestion
        const tier = TIERS.find(tier => mapTier(entryScore)(tier.score));
        if (tier) {
          const tierElement = document.querySelector(
            `#tier-${tier.name.toLowerCase()} .item-container`
```

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,205 @@ type StateProps = {
...
+    if (entryScore >= 10) {
+      return (tierScore: number) => tierScore === 10;
+    }
+    return (tierScore: number) => false;
+  };
+
+  // Renders contest entries in the correct tier using saved score upon initial render
+  useEffect(() => {
+    sortedContestEntries.forEach((entry, index) => {
+      if (entry.score !== undefined && entry.score !== null) {
```
RichDom2185, line 203: There is no need to add the check that it is not equal to undefined here, because undefined is a (reasonably) valid value. This is a sign of improper abstraction. See my below comment.

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,205 @@ type StateProps = {
...
+        const entryScore = entry.score as number;
```
RichDom2185, line 204: This type cast is incorrect because `entry.score` can be undefined. It is better (and more correct) instead for `mapTier` to take in `number | undefined`.

src/commons/sideContent/SideContentContestVoting.tsx
RichDom2185, line 205: I think if you want to implement backwards compatibility in the frontend, this would not be the best way to code it out. Personally I don't think the higher order function is necessary. I understand you want to make it explicit, but it can be done simply by:

```suggestion
        const tierIndex = TIERS.findIndex(tier => entry.score ? tier.score <= entry.score : false);
```

This would only work under the assumption that the `TIERS` array is arranged in descending tier order (which is what we want anyway).

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,205 @@ type StateProps = {
...
+      return (tierScore: number) => tierScore === 4;
+    }
+    if (entryScore >= 7 && entryScore < 10) {
+      return (tierScore: number) => tierScore === 7;
+    }
+    if (entryScore >= 10) {
+      return (tierScore: number) => tierScore === 10;
+    }
+    return (tierScore: number) => false;
+  };
```
RichDom2185, line 198: There shouldn't be hardcoded magic numbers here. Instead, reference the `TIERS` array directly (which is also why I don't think this HOF is necessary to begin with).

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,205 @@ type StateProps = {
...
+  useEffect(() => {
+    sortedContestEntries.forEach((entry, index) => {
+      if (entry.score !== undefined && entry.score !== null) {
+        const entryScore = entry.score as number;
+        const tierIndex = TIERS.findIndex(tier => mapTier(entryScore)(tier.score));
+        if (tierIndex !== -1) {
+          const tierElement = document.querySelector(
+            `#tier-${TIERS[tierIndex].name.toLowerCase()} .item-container`
+          );
+          const bankElement = document.querySelector(`#bank .item-container #item-${index + 1}`);
```
RichDom2185, line 210: This variable naming is what confused me. On second look after your explanation, I realised this is actually referring to the entry element that is inside the bank, not the bank (container) itself. Perhaps we can change this to a different name? Especially since `tierElement` above actually refers to the tier container.

RichDom2185 (changes requested):

RichDom2185:
@shenyih0ng I still think there is room for improvement and "Reactify"-ing the code even more. I am not a huge fan of how `document.querySelector` is used here; I believe in most use cases, the `useRef` hook will be able to replace this. But having said that, without playing around with the code, I don't see a straightforward way to use `useRef` here, so I think I might let it pass despite the code smell.

What do you think regarding this?

kjw142857:
Hi, just an update on the proposed changes:

- Some good news - we have successfully implemented useRef in place of the query selector using the arrays contestEntryRefs and tierContainerRefs, which means that useEffect is now fully Reactified (i.e. no more `document.querySelector`)!
- Suggested changes on simplifying the score check, removing the typecasting of entry.score and renaming tierElement/bankElement have been implemented.
- For the backward compatibility, database migration may be a good idea! However, a potential issue is that this might cause unexpected consequences to the old leaderboard. For instance (taking the most extreme case), if someone scored 9 for all votes, their score would be reset to an average of 7, while someone with a few 10 votes and the rest 7 would get a higher score. This would likely result in an incorrect leaderboard being rendered if `update_rolling_contest_leaderboards` were ever to be called in the backend again. While the chance that this will happen is low, it is possible that an instructor may accidentally re-publish the legacy assignment, which would cause `update_rolling_contest_leaderboards` to be called and the database to be updated to the inaccurate leaderboard data, causing the old leaderboard data to be lost. (Ofc, if the old leaderboard data is completely obsolete this would be perfectly fine.)

src/commons/sideContent/SideContentContestVoting.tsx
```diff
@@ -17,79 +17,190 @@ type StateProps = {
...
+      tierBoard
     ]
   );
 
+  // Upon initial render, loads contest entries in the correct tier using saved score
+  useEffect(() => {
+    sortedContestEntries.forEach((entry, index) => {
+      if (entry.score !== null) {
+        const savedTierIndex = TIERS.findIndex(tier =>
+          entry.score ? tier.score <= entry.score : false
```
RichDom2185, line 191:
```suggestion
          // We are using <= instead of === backwards
          // compatibiilty with legacy voting submisions.
          entry.score ? tier.score <= entry.score : false
```

RichDom2185 (commented):
Thanks for the hard work! Just a minor comment, let us add a comment to explain why we are using the `<=` here instead of `===`.

RichDom2185 (approved):
LGTM, thanks for the hard and consistent work on this PR!

PR #2429, opened 2023-04-07
support search in SICP JS 

yiwen101:
I add search functionality to the current NavigationBarSicp component.

need to work together with my modification for SICP backend which generate searchDataFile
run yarn json to generate searchjson Data before running yarn tryjson in the SICP backend, and run yarn start at the frontend


martin-henz (approved):
Nice from-scratch implementation of trie.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -1,8 +1,12 @@
-import { Alignment, Drawer, Navbar, NavbarGroup, Position } from '@blueprintjs/core';
+import { Alignment, Drawer, InputGroup, Navbar, NavbarGroup, Position } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
+import { memoize } from 'lodash';
 import * as React from 'react';
+//import React, {useEffect} from 'react';
+//import Highlighter from 'react-highlight-words';
```
RichDom2185, line 6: Commented code should be removed.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+        node = node[char];
+      } else {
+        return [];
+      }
+    }
+    if (node['value']) {
+      return node['value'];
+    }
+    return [];
+  }
```
RichDom2185, line 143: There's a lot of duplicated code here. There should just be one function, with the trie being queried passed in as an argument instead of relying on closure.

RichDom2185: I've fixed this in #2471. Do you mind taking a look?

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+
+  function autoComplete(str: string, limit: number, jsonData: any) {
+    if (str.length === 0) {
+      return [];
+    }
+    str = str.toLowerCase();
+    const trie = jsonData;
+
+    function next(node: any, i: number) {
```
RichDom2185, line 152: If we are using `any` like this, it's ideally only temporary, but more importantly, as it presents a lot of potential type safety issues, it should be aliased.

```ts
type TrieNode = any;
// Then, wherever `any` refers to a node in the trie...
function next(node: TrieNode, i: number) {
  // ...
}
```

This will not only greatly add readability and prevent bugs, but also developer experience (from completions, etc).

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+      return {
+        title: textbook[array[0]][array[1]][array[2]],
+        url: SearchUrl + array[0] + array[1]
+      };
+    }
+    const SearchUrl = '..';
+    let words = str.toLowerCase().split(' ');
+    words = words.filter(word => word !== '');
+    if (searchQuery.length === 0 || words.length === 0) {
+      setQueryResult([]);
```
RichDom2185, line 233: Why is it set as an empty array here, but set as `[{ title: 'no result found', url: '' }]` in the initial state (when there are also no results)?

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+      return;
+    }
+    let globalAns = queryTextTrie(words[0]).map((array: any) => lineMap(array));
+
+    globalAns = globalAns.filter((obj: any) => obj.title.toLowerCase().includes(str.toLowerCase()));
+    return globalAns;
```
RichDom2185, line 239: Let us prefer to use `const`. The mutation here is unnecessary:

```suggestion
    const globalAns = queryTextTrie(words[0]).map((array: any) => lineMap(array));
    return globalAns.filter((obj: any) => obj.title.toLowerCase().includes(str.toLowerCase()));
```

Also note the use of the unaliased `any` type (not addressed in the suggestion above).

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
RichDom2185: I've fixed this in #2471.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
RichDom2185: I've fixed this in #2471.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+
+    function recur(node: any, path: string) {
+      if (node == null || limit < 0) {
+        return;
+      }
+      if (node.value) {
+        ans.push(path);
+        limit--;
+      }
+      for (let i = 0; i < 27; i++) {
```
RichDom2185, line 210: What is 27? Is this a magic number?

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+      return [];
+    }
+    str = str.toLowerCase();
+    const trie = jsonData;
+
+    function next(node: any, i: number) {
+      if (i === 26) return node[' '];
+      else if (i >= 0 && i <= 25) {
+        return node[String.fromCharCode(i + 97)];
+      }
```
RichDom2185, line 156: Where are all these magic numbers coming from?

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+      if (array == null || array[0] == null || array[1] == null || array[2] == null) {
+        return { title: '', url: '' };
+      }
+      return {
+        title: textbook[array[0]][array[1]][array[2]],
+        url: SearchUrl + array[0] + array[1]
+      };
+    }
+    const SearchUrl = '..';
+    let words = str.toLowerCase().split(' ');
```
RichDom2185, line 230: What does this function do? The naming is unclear and it is not helped by the following:
* everything being typed as `any`
* non-descriptive variables names (e.g. `array` – what array?, `str` parameter – what string is this??)
* the code having zero comments

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+    function lineMap(array: any[]) {
+      if (array == null || array[0] == null || array[1] == null || array[2] == null) {
+        return { title: '', url: '' };
+      }
+      return {
+        title: textbook[array[0]][array[1]][array[2]],
+        url: SearchUrl + array[0] + array[1]
+      };
+    }
+    const SearchUrl = '..';
```
RichDom2185, line 229: What does this `'..'` mean?

RichDom2185 (commented):
Quite a lot of severe type-safety and readability issues. I hope my comments aren't too mean. I can help with the fixes; can you respond to the comments for clarification?

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
yiwen101: Thank you for your amazing efforts and helps!

yiwen101: I feel very sorry for the trouble caused by my incompetency; Thank you for your fixes in 2471; will definitely pay more attention to code quality in terms of type safety and readability.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
RichDom2185: Similarly, what does the `voidSearch` function do? The readability is slightly hampered by the functions being called out-of-order they're defined in.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
RichDom2185: No worries, I'll fix it. Thanks for the clarification!

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
yiwen101: I intend to have the array as the returned array of "queryTrie", with information of the queried result stored in different index of it;

lineMap takes the result of the queryTrie, and transformed it to a format that is suitable for generating content that correspond to a line in the search result displayed to the user in web

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
RichDom2185: In that case, what does `queryTrie` do?

yiwen101: for index search, since all the index are in the IndexTrie, we can just use queryTire index;

But for user search, since the textTrie only contains the occurrences of individual words rather than sentences, for example, "instruction", but not "follow instructions", I need to use voidSearch instead

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
yiwen101: Sorry, I could not make sense of it either

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+      'x',
+      'y',
+      'z',
+      ' '
+    ];
+
+    function toEnd(query: any) {
+      let node = trie;
+      for (let i = 0; i < query.length; i++) {
+        const char = query[i];
```
RichDom2185, line 192: This can be replaced with a for-of loop.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+      .toLowerCase()
+      .split(' ')
+      .filter(word => word !== '');
+    if (words.length < 2) {
+      return autoComplete(str, limit, jsonData);
+    }
+    let pre = words[0];
+    for (let i = 1; i < words.length - 1; i++) {
+      pre += ' ' + words[i];
+    }
```
RichDom2185, line 253: String concatenation using `+=` is extremely slow. The `.join()` method can be used instead, since this is an array.

RichDom2185: Fixed in #2471.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
RichDom2185: Fixed in #2471.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -63,15 +98,412 @@ const SicpNavigationBar: React.FC = () => {
...
+    if (preResults.length < 1) {
+      return [];
+    }
+    let lastwords = autoComplete(lastWord, 3000, jsonData);
+    console.log(lastwords);
+    lastwords = lastwords.filter(
+      (word: string) =>
+        preResults.filter((obj: any) => obj.title.toLowerCase().includes(pre + ' ' + word)).length >
+        0
+    );
```
RichDom2185, line 271: What is this code chunk trying to do? `.includes()` gives either `true` or `false`. What's the `.length > 0` for?

RichDom2185: These are the types of issues that will be caused by having `any` for everything; it's even worse than writing in JavaScript as writing `any` disables all checks by your editor too.

RichDom2185: Did you mean to just stop after `.includes()`? Can you explain how this entire function works, step by step, in plain English?

RichDom2185 (commented):
Some more comments.

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
yiwen101: sorry for bringing this up so late; Maybe you could skip deep "cleaning up" of my code, since many of them are meant to be abolished; I have extended it into CP3108B, and will keep on working on it next semester to properly clean it up

yiwen101: As for this function, it goes like: it firstly separate the content that in the input line into two parts, parts that are completed and parts that are not. for example, "As for this fu" will be separated into ""As for this" and "fu". 


Then, "voidSearch" search for all the occurrences of the phrase "As for this" in the textbook coupled with the complete sentence they appear in the textbook. let us say there are only two results "As for this function" "as for this example".

Then, I use the autocomplete to generate guesses for the last word, for example, "fun", "function" "fuel"......

then I check whether they appear in the sentences and filter out absent ones. For example, only "function" appear in the  "As for this function" "as for this example".

so for my sentenceComplete, everyTime it can only produce the guess for the next one word (shrug)

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
RichDom2185: Nevermind, this seems to be a path traversal.

PR #2430, opened 2023-04-08
Update README.md 

yiwen101:

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ The regression tests are generated using `jest` and stored as snapshots in `src/
 $ yarn test --updateSnapshot
 ```
 and then typing `a` to update all snapshots.
+（if you run this on a macBook with ARM chip, you may run into errors; try to do this on a computer with x86 chip instead)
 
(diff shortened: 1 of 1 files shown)


martin-henz (approved):
Thanks for sharing.

README.md
```diff
@@ -153,6 +153,7 @@ The regression tests are generated using `jest` and stored as snapshots in `src/
 $ yarn test --updateSnapshot
 ```
 and then typing `a` to update all snapshots.
+（if you run this on a macBook with ARM chip, you may run into errors; try to do this on a computer with x86 chip instead)
```
RichDom2185, line 156: May I know what errors, exactly? I'm on an M1 Macbook and there have been no errors whenever I try to update snapshots.

yiwen101: there is not "node-canvas" version available for my Mac. On visiting the website, I also did not find available version.

RichDom2185: What's your node version? As well as what Mac are you using?

PR #2460, opened 2023-04-28
Clean up language selector

### LINKED ISSUE #2459: Sublanguage selector goes blank at SA @ X deployment
Testing locally, we see that when a course does not enable a language, the language selector at the top right is still rendered. This leads to cases where selecting a language with no sublanguages enabled for the course results in the following empty dropdwon:

image

RichDom2185:
### Description

Cleans up a bunch of code style and quality issues following #2444. Also closes #2459. Do look at commit history for more details.

Turns out the chapter list renderer has been used wrongly this whole time. Items were fetched from exported variables instead of the state. This made it hard to filter. Has since been fixed in 2a3801e7694db23ea2f6f787254458d588071598.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "@blueprintjs/popover2": "^1.4.3",
     "@blueprintjs/select": "^4.5.1",
     "@octokit/rest": "^19.0.5",
-    "@reduxjs/toolkit": "^1.9.3",
     "@sentry/browser": "^7.8.0",
     "@sourceacademy/sharedb-ace": "^2.0.2",
     "@sourceacademy/sling-client": "^0.1.0",
--- a/src/commons/application/ApplicationTypes.ts
+++ b/src/commons/application/ApplicationTypes.ts
@@ -237,14 +237,6 @@ export const sourceLanguages: SALanguage[] = sourceSublanguages.map(sublang => {
   };
 });
 
-export const defaultLanguages = sourceLanguages.filter(
-  sublang => sublang.variant === Variant.DEFAULT
-);
-
-export const variantLanguages = sourceLanguages.filter(
-  sublang => sublang.variant !== Variant.DEFAULT
-);
-
 export const isSourceLanguage = (chapter: Chapter) =>
   [Chapter.SOURCE_1, Chapter.SOURCE_2, Chapter.SOURCE_3, Chapter.SOURCE_4].includes(chapter);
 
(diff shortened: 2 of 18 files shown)


RichDom2185:
Before this PR, the existing behaviour was that the following languages:

- Full JS
- Full TS
- HTML
- All Scheme sublanguages
- All Python sublangages

Will only show up in the public deployment (and not Source Academy @ X). I've changed the behaviour such that Scheme and Python now also shows up in SA@X deployments.

I believe @zhaojj2209 mentioned something about potential security vulnerabilities/XSS/etc. if we run Full JS/TS/HTML. @zhaojj2209 Can I check if there were any other reasons these languages were set to the public deployment only? I've made sure to maintain that same behaviour for SA@X deployments, as seen in:

https://github.com/source-academy/frontend/blob/81baaea268a2dbc6048c3ee5624cb811f83593e6/src/commons/controlBar/ControlBarChapterSelect.tsx#L85

I think enabling Scheme and Python are fine (since they are not running "full" versions of the language). This also "fixes" #2459. What are your thoughts on this @martin-henz ?

RichDom2185:
@s-kybound can I check on the difference between "Scheme §4" and "Full Scheme"? Would the abovementioned security issue be likely?

@Fidget-Spinner @dylkaw likewise, are there plans for "Full Python"?

Fidget-Spinner:
No plans for Full Python. A full Python to JS transpiler would be a monumental project.

s-kybound:
> @s-kybound can I check on the difference between "Scheme §4" and "Full Scheme"? Would the abovementioned security issue be likely?

Hi,  "Scheme §4" and "Full Scheme" currently have no differences.  "Scheme §4" only enables enough of Scheme to complete SICP Chapter 4, while "Full Scheme" will enable features like macros and first-class continuations. I don't think there will be security issues that could arise from this.

s-kybound (approved):
Looks good to merge

zhaojj2209:
> I believe @zhaojj2209 mentioned something about potential security vulnerabilities/XSS/etc. if we run Full JS/TS/HTML. @zhaojj2209 Can I check if there were any other reasons these languages were set to the public deployment only?

Full JS/TS should not ever be deployed to the NUS version because they make use of [`eval()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!), which is a huge security risk. The HTML chapter uses a sandboxed iframe that allows scripts but disables everything else (e.g. cross-origin, popups, etc) so it is relatively safer, but let's still keep it playground-only to be on the safe side.

PR #2461, opened 2023-04-29
Refactor playground

RichDom2185:
### Description

To prepare for language refactoring. Part of #2463. I know converting the if-else statements into switch-case blocks somewhat makes it less readable, but this is a temporary change as once the logic for rendering is abstracted out, I promise it would be much cleaner.

### Type of change

### Checklist

- [ ] I have tested this code


src/pages/playground/Playground.tsx
```diff
@@ -732,85 +689,62 @@ const Playground: React.FC<PlaygroundProps> = ({ workspaceLocation = 'playground
...
+        // (TEMP) Remove tabs for fullJS until support is integrated
+        return [...tabs, dataVisualizerTab];
+      case !usingRemoteExecution:
+        // Don't show the following when using remote execution
+
+        // Enable Data Visualizer for Source Chapter 2 and above
+        const shouldShowDataVisualizer = currentLang >= Chapter.SOURCE_2;
+        if (shouldShowDataVisualizer) {
+          tabs.push(dataVisualizerTab);
+        }
```
ianyong, line 730: > I know converting the if-else statements into switch-case blocks somewhat makes it less readable, but this is a temporary change as once the logic for rendering is abstracted out, I promise it would be much cleaner.

I won't block on this since it's a temporary change, but wouldn't writing this using chained if statements be easier to read without having to perform huge restructuring the code here? Using a switch statement is a bit weird since conventionally, the expression to be matched is the one whose values changes, and not the case clauses.

src/commons/sideContent/SideContentTypes.ts
```diff
@@ -37,6 +37,10 @@ export enum SideContentType {
   htmlDisplay = 'html_display'
 }
 
+export const isVisualizerTab = (tab: SideContentType) => {
+  return tab === SideContentType.substVisualizer || tab === SideContentType.envVisualizer;
+};
```
ianyong, line 42: Should the check in the `onChangeTabs` callback really be extracted out here? By the name of this function, I would expect tabs such as the data visualizer to return true when passed as argument into this function but this is not the case. You could rename this function to specifically check for the stepper and environment visualizer, but at that point you might as well inline this function in the one place it is used.

There's also the matter of future visualizers being added. If we ignore the current usage in the `onChangeTabs` callback and assume this function should do what its name suggests, it's very easy for this function to be overlooked (and not be updated) when new visualizers are added. Perhaps whether a tab is a visualizer or not can be an attribute on the `SideContentType` if we want this particular check.

src/pages/playground/PlaygroundTabs.tsx
```diff
@@ -0,0 +1,89 @@
+import { IconNames } from '@blueprintjs/icons';
```
ianyong, line 1: It would be good if how the other workspaces handle side content tabs can be standardised with the playground. It seems that the functions within this file should be easily generalisable?

No changes required in this PR since this would be out of scope but just noting this down.

ianyong (changes requested):
Only blocking issue is my comment on `isVisualizerTab`.

src/pages/playground/Playground.tsx
RichDom2185: Actually yea, on second thought, I might revert this in the future PR. Thanks!

src/pages/playground/PlaygroundTabs.tsx
RichDom2185: Agreed, perhaps we can further abstract and generalise this logic to work with other workspaces. The main reason I separated the playground tabs into a separate file was because `Playground.tsx` was getting too huge with over 1100 LoC in that file alone. I'll look into this.

src/commons/sideContent/SideContentTypes.ts
RichDom2185: That's true. Perhaps we can just create this function locally within `Playground.tsx`, instead of exporting it. Thanks!

RichDom2185: Resolved in 8e2cea90de8fd233bdc7c49e6b28d10386c765a1.

src/pages/playground/Playground.tsx
RichDom2185: Changed in 5fe986b74d4cebdadf9a930481e9c643546d558d.

src/commons/sideContent/SideContentTypes.ts
ianyong: Hmm, even if the function is placed within `Playground.tsx`, wouldn't the fact that it returns false for the data visualizer still be an issue? Or at least, this is a surprising result for me.

RichDom2185: Hmm, that's fair. I guess this is a relatively minor this. I'll revert it back for now since it's the only thing that's blocking while we think about it.

ianyong (approved):

Issue #2463, opened 2023-04-29
Refactor language configuration


RichDom2185:
Currently, language-specific behaviours (e.g. what tabs to show for which chapters of Source) are coded into the respective components/pages (e.g. `Playground.tsx`).

As we add more and more languages and variants, the mess of if-else, conditionals, etc. becomes more and more unreadable. Ideally, the behaviour of these should be dictated by the language itself.

Let us refactor the codebase to shift this logic into a language configuration type, with this configuration dictating the behaviour of the components.

Roadmap:

- [ ] Create and persist action to update Playground language configuration
- [ ] Migrate logic in Playground to use language configuration
- [ ] Remove redundant `playground.lang` state value and related actions (simply having a language name will be redundant once we have a language configuration)
- [ ] Refactor `chapterSelect`, related actions, state, props, etc. to use language configuration (simply having `chapter` and `variant` will be redundant once we have a language configuration)

PR #2464, opened 2023-04-29
Create language configuration for conditional rendering of tabs

RichDom2185:
### Description

Part of #2463. Depends on #2461. Persists the current language configuration (the entire `SALanguage` object), beyond just the chapter and variant.

Also adds "flags" to `SALanguage` to dictate whether to use the subst visualizer or the environment visualizer instead.

Using these flags will improve the scalability when we make the visualizers support more languages.

### Type of change

### Checklist

- [x] I have tested this code


RichDom2185:
@ianyong perhaps we can move the check for multi-file support (`!== Chapter.SOURCE_1`) into this language configuration object as well.

ianyong:
> @ianyong perhaps we can replace the check for multi-file support (`!== Chapter.SOURCE_1`) to use a flag like `supportsMultiFile` in this language configuration object as well. Thoughts on this?

Sure, that would be great! Thanks.

src/commons/application/ApplicationTypes.ts
```diff
@@ -207,26 +209,29 @@ export const styliseSublanguage = (chapter: Chapter, variant: Variant = Variant.
   }
 };
 
-export const sourceSublanguages: Language[] = [
-  { chapter: Chapter.SOURCE_1, variant: Variant.DEFAULT },
+export const sourceSublanguages = [
```
ianyong, line 212: Would be good to type this anyway, if only to prevent typos.

src/commons/application/ApplicationTypes.ts
```diff
@@ -207,26 +209,29 @@ export const styliseSublanguage = (chapter: Chapter, variant: Variant = Variant.
...
+  { chapter: Chapter.SOURCE_1, variant: Variant.DEFAULT, usesSubstVisualizer: true },
   { chapter: Chapter.SOURCE_1, variant: Variant.TYPED },
   { chapter: Chapter.SOURCE_1, variant: Variant.WASM },
   { chapter: Chapter.SOURCE_1, variant: Variant.LAZY },
-  { chapter: Chapter.SOURCE_1, variant: Variant.NATIVE },
-  { chapter: Chapter.SOURCE_2, variant: Variant.DEFAULT },
+  { chapter: Chapter.SOURCE_1, variant: Variant.NATIVE, usesSubstVisualizer: true },
+
```
ianyong, line 218: I'm surprised the linter doesn't complain!

src/commons/application/ApplicationTypes.ts
```diff
@@ -273,9 +287,16 @@ export const defaultAchievement: AchievementState = {
   assessmentOverviews: []
 };
 
+export const defaultLanguageConfig: SALanguage = ALL_LANGUAGES.find(
+  sublang =>
+    sublang.chapter === Constants.defaultSourceChapter &&
+    sublang.variant === Constants.defaultSourceVariant
+)!;
```
ianyong, line 294: I really don't like the non-null assertion operator. It might be the case that the default Source chapter & variant will always have a valid `SALanguage` now, but we cannot guarantee this to be true in the future. A preferable alternative would be to add a guard clause that throws an error so that if and when this invariant is broken in the future, we know exactly what the problem is.

ianyong (changes requested):

src/commons/application/ApplicationTypes.ts
ianyong: This would be caught when the declaration of `ALL_LANGUAGES` is processed, but it still would be good to catch any such issues earlier.

RichDom2185: We can't type it as `Language[]` because the flags/other config stuff like `usesSubstVisualizer` are not defined in the `Language` interface. We can't type as `SALanguage[]` either since they are missing keys like `mainLanguage`.

Any errors will actually be caught earlier, not in `ALL_LANGUAGES`, but in `sourceLanguages`/`pyLanguages`/`schemeLanguages` as they are all typed as `SALanguage[]`.

Since errors would be caught anyway, moreover, these `xxxSubLanguages` are not exported (only the languages are), I think it's fine to leave them untyped.

What do you think?

src/commons/application/ApplicationTypes.ts
RichDom2185: I believe if I try to add another key, it'll auto convert to having each key in a separate line. This happens to be just nicely within the line length limit :)

src/commons/application/ApplicationTypes.ts
RichDom2185: Noted and agreed, will make the changes. Thanks!

RichDom2185: Resolved in 2013b7d1bee0909e5a6c7b9d7a64ebbabb4c9dec.

src/commons/application/ApplicationTypes.ts
ianyong: I was thinking that it could have its own type which is a subset of `SALanguage`, but you're right that the errors will be caught anyway so this is more of a minor nit.

src/commons/application/ApplicationTypes.ts
```diff
@@ -230,16 +240,40 @@ const sourceSubLanguages: Language[] = [
...
+  // Enable Data Visualizer for Source Chapter 2 and above
+  supportedFeatures.dataVisualizer = chapter >= Chapter.SOURCE_2;
+
+  // Enable Subst Visualizer only for default Source 1 & 2
+  supportedFeatures.substVisualizer =
+    chapter <= Chapter.SOURCE_2 && (variant === Variant.DEFAULT || variant === Variant.NATIVE);
+
+  // Enable Env Visualizer for Source Chapter 3 and above
+  supportedFeatures.envVisualizer =
+    chapter >= Chapter.SOURCE_3 && variant !== Variant.CONCURRENT && variant !== Variant.NON_DET;
```
RichDom2185, line 255: I decided to backtrack on manually specifying the language configurations and placed them in a test file instead. I think going with this "rule-based" approach seems neater

src/commons/application/ApplicationTypes.ts
```diff
@@ -157,7 +165,7 @@ export const htmlLanguage: SALanguage = {
   mainLanguage: SupportedLanguage.JAVASCRIPT
 };
 
-const schemeSubLanguages = [
+const schemeSubLanguages: Array<Pick<SALanguage, 'chapter' | 'variant' | 'displayName'>> = [
```
ianyong, line 168: This makes the typing unnecessarily convoluted. :/

Could `displayName` be part of `Language` instead? It doesn't make sense that the Source sublanguages are handled differently from Scheme and Python in this regard.

Alternatively, the display names could be derived for Scheme and Python in the same manner as for Source. The way Scheme and Python currently do it is absolutely terrible because the `styliseSublanguage` function is the source of truth for the Source derived names, but not for Scheme and Python:
https://github.com/source-academy/frontend/blob/ce58e6d5d3a2db9719957065156a70ec4a43a604/src/commons/application/ApplicationTypes.ts#L184-L208
Instead, we get the above hacks that make use of the non-null assertion operator. (We really should ban it in the linter so that people stop adding them...)

To be honest, I'm not sure why when it comes to displaying these language strings, we sometimes call `styliseSublanguage` and other times just access the `displayName` attribute. If they are truly one and the same, we should make use of the latter only, with the former possibly remaining as a convenience function for calculating all of the Source sublanguage display names.

src/commons/application/ApplicationTypes.ts
```diff
@@ -230,16 +241,56 @@ const sourceSubLanguages: Language[] = [
 ];
 
 export const sourceLanguages: SALanguage[] = sourceSubLanguages.map(sublang => {
+  const { chapter, variant } = sublang;
+  const supportedFeatures: LanguageFeatures = sublang['supports'] ?? {};
```
ianyong, line 245: Is there a need to retrieve the `supports` object if it will always be empty and all of its fields will always be overridden?

src/pages/__tests__/createStore.test.ts
```diff
@@ -43,7 +44,8 @@ const mockChangedStoredState: SavedState = {
   playgroundSourceChapter: Constants.defaultSourceChapter,
   playgroundSourceVariant: Variant.DEFAULT,
   playgroundExternalLibrary: 'NONE' as ExternalLibraryName,
-  playgroundLang: SupportedLanguage.JAVASCRIPT
+  playgroundLang: SupportedLanguage.JAVASCRIPT,
+  playgroundLanguage: defaultLanguageConfig
```
ianyong, line 48: Can these fields be better named?

src/pages/login/__tests__/Login.tsx
```diff
@@ -9,7 +9,9 @@ jest.mock('../../../commons/utils/Constants', () => {
   return {
     __esModule: true,
     default: {
-      authProviders: new Map([['luminus', { name: 'LumiNUS' }]])
+      authProviders: new Map([['luminus', { name: 'LumiNUS' }]]),
+      defaultSourceChapter: 4,
+      defaultSourceVariant: 'default'
```
ianyong, line 14: Possible to use the enums here instead?

src/pages/playground/Playground.tsx
```diff
@@ -251,14 +253,14 @@ export async function handleHash(
...
-    const variant: Variant =
-      sourceLanguages.find(
-        language => language.chapter === chapter && language.variant === qs.variant
-      )?.variant ?? Variant.DEFAULT;
-
     if (chapter) {
-      props.handleChapterSelect(chapter, variant);
+      // TODO: To migrate the state logic away from playgroundSourceChapter
+      //       and playgroundSourceVariant into the language config instead
+      const languageConfig = getLanguageConfig(chapter, qs.variant as Variant);
```
ianyong, line 259: If this is temporary, I won't block on this. Otherwise, I'm not comfortable with this type cast since `qs.variant` can be some garbage value.

src/pages/playground/Playground.tsx
```diff
@@ -695,14 +710,31 @@ const Playground: React.FC<PlaygroundProps> = ({ workspaceLocation = 'playground
...
+    );
+    // Hardcoded for Playground only for now, while we await workspace refactoring
+    // to decouple the SicpWorkspace from the Playground.
+    dispatch(playgroundConfigLanguage(languageConfigToSet));
+  }, [dispatch, props.playgroundSourceChapter, props.playgroundSourceVariant]);
+
+  const languageConfig: SALanguage = useTypedSelector(state => state.playground.languageConfig);
+  const shouldShowDataVisualizer = languageConfig.supports?.dataVisualizer ?? false;
+  const shouldShowEnvVisualizer = languageConfig.supports?.envVisualizer ?? false;
+  const shouldShowSubstVisualizer = languageConfig.supports?.substVisualizer ?? false;
```
ianyong, line 728: Would it be better to simply enforce the presence of the `supports` object instead? That way, we handle all of this complexity in a single location rather than having to check everywhere.

ianyong (changes requested):

src/commons/application/ApplicationTypes.ts
RichDom2185: > Could `displayName` be part of `Language` instead?

I agree this should be the case. Unfortunately `Language` is defined in js-slang, not the frontend. Perhaps we can create our own version of `Language` temporarily? That way we can use `SALanguage` $<:$ `Language` (frontend) $<:$ `Language` (js-slang).

Regarding `styliseSublanguage`, I wholeheartedly agree. I wasn't sure if I should do the refactoring for that in this PR as well. Essentially, I believe that what is happening is that originally, when we only had Source, they tried to abstract it this way such that one can get the display name simply by passing a chapter and a variant, which might have seemed like a good idea then, but now it's just messy since we have languages that are not Source, so in an effort to add support for these languages for now, they did this patchwork.

Here's my proposal:

1. Move https://github.com/source-academy/frontend/blob/93396dc689e743e5f517e813a93e7ffe520c8b6d/src/commons/application/ApplicationTypes.ts#L212-L213 in `styliseSublanguage` to https://github.com/source-academy/frontend/blob/93396dc689e743e5f517e813a93e7ffe520c8b6d/src/commons/application/ApplicationTypes.ts#L263
2. (Temporarily) rewrite `styliseSublanguage` to simply return `getLanguageConfig(chapter, variant).displayName`: https://github.com/source-academy/frontend/blob/93396dc689e743e5f517e813a93e7ffe520c8b6d/src/commons/application/ApplicationTypes.ts#L280-L292 for compatibility
3. Once logic fully migrated away from simply passing around the chapter and variant (see last task of #2463), we replace every instance of `styliseSublanguage` to the `displayName` attribute directly

What do you think?

src/commons/application/ApplicationTypes.ts
RichDom2185: My bad, this was originally done as a guard against someone potentially adding a `supports` object in `sourceSubLanguages`, before I explicitly typed the array. This is not needed anymore and can just be initialized as an empty object.

src/pages/__tests__/createStore.test.ts
RichDom2185: `playgroundLang` is meant to be deprecated and removed once `playgroundLanguage` (which contains the entire config, not just the `mainLanguage`, which is what `playgroundLang` is) set up &rarr; see task 3 of #2463.

`playgroundLang` was added recently in #2444, in a rush, as we wanted to quickly add the language dropdown selector so that people can start experimenting with the non-Source languages. Its value is equal to `playgroundLanguage.mainLanguage`, so this will be made redundant.

I'm keeping it in this PR as the components that depend on `playgroundLang` are not ready yet for the migration.

src/pages/login/__tests__/Login.tsx
RichDom2185: I'm not too familiar with Jest, so I don't know if it's truly impossible, but I tried using the enums at first, and it gave me an error saying that we can't reference external variables (including enums) in Jest mocks. I didn't think it was absolutely necessary to use the enums as this was a test file.

src/pages/playground/Playground.tsx
RichDom2185: 1. It is temporary
2. We are using `getLanguageConfig` which will catch the garbage value and throw an error (as it would result in not being able to find a language config that matches the constraints), so even now it's still safe: https://github.com/source-academy/frontend/blob/93396dc689e743e5f517e813a93e7ffe520c8b6d/src/commons/application/ApplicationTypes.ts#L280-L292

src/pages/playground/Playground.tsx
RichDom2185: Hmm that's fair. I guess keeping `supports` optional for the sake of backward compatibility will only increase the tech debt in the future. I will enforce the presence of this key. Thanks!

src/pages/playground/Playground.tsx
ianyong: Hmm, the type cast is a code smell. Should the type of `variant` be made wider in `getLanguageConfig` then if it accepts and handles garbage values correctly? Can ignore if the comment on line 280 means that this function is just temporary.

src/pages/login/__tests__/Login.tsx
ianyong: Ah I see. :(

src/commons/application/ApplicationTypes.ts
ianyong: > I agree this should be the case. Unfortunately `Language` is defined in js-slang, not the frontend. Perhaps we can create our own version of `Language` temporarily? That way we can use `SALanguage` $<:$ `Language` (frontend) $<:$ `Language` (js-slang).

That makes sense.

> Regarding `styliseSublanguage`, I wholeheartedly agree. I wasn't sure if I should do the refactoring for that in this PR as well. Essentially, I believe that what is happening is that originally, when we only had Source, they tried to abstract it this way such that one can get the display name simply by passing a chapter and a variant, which might have seemed like a good idea then, but now it's just messy since we have languages that are not Source, so in an effort to add support for these languages for now, they did this patchwork.

Yea, I agree that this is slightly out of scope of this PR already. Unfortunately, it also unnecessarily complicates the typing here.

> Here's my proposal:
>
> 1. Move https://github.com/source-academy/frontend/blob/93396dc689e743e5f517e813a93e7ffe520c8b6d/src/commons/application/ApplicationTypes.ts#L212-L213 in `styliseSublanguage` to https://github.com/source-academy/frontend/blob/93396dc689e743e5f517e813a93e7ffe520c8b6d/src/commons/application/ApplicationTypes.ts#L263
> 2. (Temporarily) rewrite `styliseSublanguage` to simply return `getLanguageConfig(chapter, variant).displayName`: https://github.com/source-academy/frontend/blob/93396dc689e743e5f517e813a93e7ffe520c8b6d/src/commons/application/ApplicationTypes.ts#L280-L292 for compatibility
> 3. Once logic fully migrated away from simply passing around the chapter and variant (see last task of #2463), we replace every instance of `styliseSublanguage` to the `displayName` attribute directly
>
> What do you think?

Sounds great! Also sounds very painful... Thanks for carrying out this refactoring.

src/pages/playground/Playground.tsx
RichDom2185: Resolved in d0acd9810c14607036f64f8513613508a79547f6

src/commons/application/ApplicationTypes.ts
RichDom2185: Resolved in bbe80f4ce064a2d4afb5170655c1f0fe332859bf.

RichDom2185:
@ianyong based on the discussion above in terms of what will be in scope for this PR, is the current state of it okay to be merged?

src/pages/playground/Playground.tsx
```diff
@@ -695,16 +710,27 @@ const Playground: React.FC<PlaygroundProps> = ({ workspaceLocation = 'playground
...
+    );
+    // Hardcoded for Playground only for now, while we await workspace refactoring
+    // to decouple the SicpWorkspace from the Playground.
+    dispatch(playgroundConfigLanguage(languageConfigToSet));
+  }, [dispatch, props.playgroundSourceChapter, props.playgroundSourceVariant]);
+
+  const languageConfig: SALanguage = useTypedSelector(state => state.playground.languageConfig);
+  const shouldShowDataVisualizer = languageConfig.supports.dataVisualizer ?? false;
+  const shouldShowEnvVisualizer = languageConfig.supports.envVisualizer ?? false;
+  const shouldShowSubstVisualizer = languageConfig.supports.substVisualizer ?? false;
```
ianyong, line 728: The `??` operators should not longer be necessary, right?

ianyong (approved):
Other than that, LGTM!

src/pages/playground/Playground.tsx
RichDom2185: It is still necessary. While `supports` is now enforced, the individual keys of `LanguageFeatures` are still optional. While they are defined for Source languages, at the moment, there is still no such "rule-based" logic for Python and Scheme, so they are still undefined there.

After taking another look at it now, perhaps we should make these keys required as well to make the code cleaner, and just explicitly set their values as false for the non-Source languages. I will do that in a subsequent PR.

Thanks for the comments! Sorry I forgot to request your review again 😅.

ianyong: Ah ok, sounds good!

src/commons/application/ApplicationTypes.ts
RichDom2185: Steps 1 and 2 have been done in 7850dbce629649614093631ca90ccd5027a21746 as part of #2470.

PR #2466, opened 2023-05-01
Fix incorrect syntax highlighting for non-Source languages

RichDom2185:
Enables the correct syntax highlighting rules for non-Source languages.

Note that this is a hotfix. Ideally, the highlighting rules should be defined in js-slang together with the language itself, just like how it is currently implemented for Source languages.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/utils/AceHelper.ts
+++ b/src/commons/utils/AceHelper.ts
@@ -32,8 +32,25 @@ export const selectMode = (chapter: Chapter, variant: Variant, library: string) 
 };
 
 export const getModeString = (chapter: Chapter, variant: Variant, library: string) => {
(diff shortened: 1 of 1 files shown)


RichDom2185:
@zhaojj2209 I was wondering, do we have the syntax highlighting rules for full TS/JS defined in js-slang?

zhaojj2209:
> @zhaojj2209 I was wondering, do we have the syntax highlighting rules for full TS/JS defined in js-slang?

Full TS no, Full JS yes but the rules are in the frontend repo (`src/features/fullJS/fullJSHighlight.js`).

zhaojj2209 (approved):

Issue #2458, opened 2023-04-23
Env Viz: inconsistent state after first breakpoint is hit


martin-henz:
https://user-images.githubusercontent.com/30220768/233819396-b6412df2-4a28-4524-92ae-0da27a11c60e.mov

The slider indicates that the first step is displayed, but the state shown is the last state.

PR #2473, opened 2023-05-09
Fixed the inconsistency in the environment visualiser with breakpoint…

vansh284:
### Description

This PR fixes https://github.com/source-academy/frontend/issues/2458.

Fixing this issue required that the number of steps in the environment visualizer slider was always explicitly obtained from the redux state. Previously, it was only obtained when the environment visualizer side content tab was open, which is not the case when breakpoints are used.

There was a similar issue in the _total_ number of steps for the slider not being updated when using breakpoints. Previously, this was only updated when the environment visualizer side content tab was open, but it has now been changed to update whenever code is executed using the explicit control evaluator (this effectively includes both cases).

Documentation has also been added.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/sagas/WorkspaceSaga.ts
+++ b/src/commons/sagas/WorkspaceSaga.ts
@@ -993,6 +993,20 @@ export function* evalCode(
             .usingSubst
       )
     : false;
+  const stepLimit: number = yield select(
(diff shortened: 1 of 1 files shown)


RichDom2185 (approved):
LGTM

PR #2476, opened 2023-05-10
Errors from the previous context are reset when evaluating the REPL

vansh284:
### Description

This PR fixes https://github.com/source-academy/js-slang/issues/1383.

The reason for this bug was that context.errors was not being cleared every time the REPL was evaluated resulting in the same error propagating repeatedly even for valid programs. 

### Type of change

- [x] I have tested this code


RichDom2185 (approved):
LGTM

Issue #1376, opened 2020-07-20
Coding standard: Import statements using mixed relative & absolute path


jetkan-yk:
Currently we have an iron rule of using relative import paths, which creates some unnecessarily long paths:
`import { AchievementItem } from '../../../../../../features/achievement/AchievementTypes';`
could have been:
`import { AchievementItem } from 'src/features/achievement/AchievementTypes';`
Besides, all import paths need to be updated everytime a file is moved to a different directory. This happens quite frequently when implementing a new feature.

I'm proposing a new standard for import statements:
#### Imports that have >= 3 levels of **common root path** should use **relative** path. Otherwise use **absolute** path.

`src/pages/achievement/dashboard/Dashboard.tsx`
Import paths with prefix `src/page/achievement/...` use **relative**, otherwise use **absolute** e.g.

Import: `src/pages/achievement/dashboard/subcomponents/AchievementTask.tsx` -> use **relative** `./subcomponents/AchievementTask.tsx`
Import: `src/pages/achievement/Achievement.tsx` -> use **relative** `../Achievement.tsx`
Import: `src/pages/academy/game/Game.tsx` -> use **absolute**
Import: `src/common/utils/Constants.ts` -> use **absolute**

PR #2477, opened 2023-05-12
Standardize imports

RichDom2185:
### Description

**Motivation:** Copied over from their own [documentation](https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths)

> Moving a file to different folder, could result in changing all imports statement in that file. This will not happen is the import paths are absolute.

Also, the current way of importing leads to stuff like

```tsx
import { a } from 'src/.../file'

import { b } from '../file'
```

Even though `a` and `b` are defined in the same file, which does not really look nice at best and potentially misleading at worst.

**Changes:**

* Migrated `.eslintrc` from JSON to a JS file to allow for greater readability when writing rules
* Added no-relative-import-paths plugin
* Enforced absolute imports for files that are not test files (reason being test files generally import their SUT from their siblings/cousins)

### Type of change

### Checklist

- [x] I have tested this code


RichDom2185:
Gonna leave this as draft for a while because the changes across the files might cause merge conflicts with other PRs; let us merge as many of them in as possible first.

.eslintrc.js
```diff
@@ -0,0 +1,26 @@
...
+    '@typescript-eslint/interface-name-prefix': 'off',
+    '@typescript-eslint/camelcase': 'off',
+    '@typescript-eslint/no-inferrable-types': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/no-non-null-assertion': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/ban-ts-comment': 'warn',
+    '@typescript-eslint/ban-types': 'off',
+    'no-relative-import-paths/no-relative-import-paths': ['error', { allowSameFolder: true }],
```
ianyong, line 15: > **Motivation:** Copied over from their own [documentation](https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths)
>
> > Moving a file to different folder, could result in changing all imports statement in that file. This will not happen is the import paths are absolute.
>
> Also, the current way of importing leads to stuff like
>
> ```tsx
> import { a } from 'src/.../file'
>
> import { b } from '../file'
> ```
>
> Even though `a` and `b` are defined in the same file, which does not really look nice at best and potentially misleading at worst.

Hmm, given the motivation for this change, is there a reason why the `allowSameFolder` option is set to true? More specifically, why are relative import paths for imported files from the same folder allowed? Moving a file containing such relative import paths will result in these import statements changing.

ianyong (approved):
LGTM! Only checked the changes made by hand in detail; briefly scrolled through the linter/formatter changes and they seem to be correct.

Just one minor clarification.

.eslintrc.js
ianyong: No strong preferences! Just wanted to know if there was a rationale (which you do have) because it would be good to state it somewhere for future reference.

chownces:
Refer to #1376

RichDom2185:
> Refer to #1376

Thanks! Is there a way to lint this, though? I tried searching for eslint plugins but couldn't find anything that offers this level of customisation

Issue #2469, opened 2023-05-08
Expired refresh token does not sign the user out


RichDom2185:
As discussed with @chownces @YaleChen299 @shenyih0ng  last winter:

For SA@X deployments, the user is not automatically signed out after the token expires. Instead, the user gets a lot of 401 popup notifications and they would have to manually log out.

chownces:
Refer to #2100 for more context. As previously discussed with the profs, we should account for the case where the user has unsaved work on refresh token expiry. We should allow the user to copy their work first instead of forcing a logout immediately, which would redirect them to the login page.

vansh284:
@chownces Could you provide some elaboration? 

From my understanding https://github.com/source-academy/frontend/issues/2100 seemed to be concerned about cases where the user's refresh token was valid but he/she was auto-logged out by other background requests which returned a 401/403. 

I may be missing some context regarding your discussions, what would be some cases where the user has unsaved work after refresh token expiry i.e. 1 week of not logging in?

Also, wouldn't the expired refresh token make the users unable to save their work since the backend authentication fails? (not 100% sure about this because I'm not too familiar with our backend).

RichDom2185:
> what would be some cases where the user has unsaved work after refresh token expiry i.e. 1 week of not logging in?

E.g. A user opens a mission, works on it without saving/submitting, closes their laptop for the night (leaving their browser tab open), only to come back to it tomorrow and realise that they can't submit because their access token happens to have expired overnight.

> wouldn't the expired refresh token make the users unable to save their work since the backend authentication fails?

Correct. What #2100 is trying to say by "saving their work" is the user and copy whatever code they need elsewhere (or simply sign in from another browser tab) and submit that way, retaining all their work. If we log the user out immediately (and redirect to login page) the user would lose all the work they have done since the last save point.

PR #2483, opened 2023-05-15
Invalid Refresh Token Signs the User Out

vansh284:
### Description

This PR fixes https://github.com/source-academy/frontend/issues/2469. 

For context, previously all backend errors would cause the user to auto-logout but this caused users to be unnecessarily logged out. This was fixed in https://github.com/sourceacademy/frontend/commit/691baa22b7adee268ada8499ea1a373cd3e1085d. However, this resulted in users remaining logged in even after their refresh tokens expired.

This commit checks whether the refresh token is invalid by checking whether the response text matches the backend response for an invalid refresh token (https://github.com/source-academy/backend/blob/master/lib/cadet_web/controllers/auth_controller.ex) and only logs out the user in this case,

### Type of change

### How to test
Set up a local source-academy frontend and backend. Change the `generate_tokens` function in the backend at `backend/lib/cadet_web/controllers/auth_controller.ex` to reduce the refresh and access token expiry to a few seconds/minutes for testing. Requests made after the refresh token expiry time will force the user to log out.

### Checklist

- [x] have tested this code


chownces:
Possibly superseded by #2531

RichDom2185:
> Possibly superseded by #2531

The direction of #2531 looks good. Will proceed to close this as superseded by it.

PR #2485, opened 2023-05-17
Use separate EditSession for each editor tab

### LINKED ISSUE #2401: Maintain separate Ace editor state for every editor tab
Currently, the Ace editor maintains a global state across all instances of the Ace editor component. This means that cursor position, undo/redo, etc is shared across all editor tabs.

We will need to devise a mechanism to make use of separate [`EditSession`](https://ajaxorg.github.io/ace-api-docs/classes/Ace.EditSession.html)s for each editor tab.

ianyong:
### Description

Store a separate `EditSession` object for each editor tab. This allows each editor tab to maintain its own cursor position, highlighting, edit history, scroll position, etc.

vmconnect_TUniDmDC4m

Resolves #2401.

### Type of change

### How to test

Enable Folder mode, then check that each editor tab has a separate history of operations.

### DIFF
--- a/src/commons/editor/Editor.tsx
+++ b/src/commons/editor/Editor.tsx
@@ -342,6 +342,8 @@ const EditorBase = React.memo((props: EditorProps & LocalStateProps) => {
   if (editor !== undefined) {
     if (filePath !== props.filePath) {
       editor.setSession(props.session);
+      // Give focus to the editor tab after switching.
+      editor.focus();
       setFilePath(props.filePath);
     }
   }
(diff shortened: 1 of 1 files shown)


src/commons/editor/Editor.tsx
```diff
@@ -330,23 +331,43 @@ const handlers = {
...
+  const editor = reactAceRef.current?.editor;
+  // Set edit session history when switching to another editor tab.
+  if (editor !== undefined) {
+    if (filePath !== props.filePath) {
+      // Unfortunately, the current editor sets the mode globally.
+      // The side effects make it very hard to ensure that the correct
+      // mode is set for every EditSession. As such, we add this one
+      // line to always propagate the mode whenever we set a new session.
+      // See AceHelper#selectMode for more information.
+      props.session.setMode(editor.getSession().getMode());
```
ianyong, line 349: Can't say I'm too happy about this, but in theory, as long as the current mode is always correct, this should work. The proper way would be to not set the mode globally but this hack is too ingrained in the codebase...

zhaojj2209 (approved):

Issue #2434, opened 2023-04-09
SICP JS: add section information to search results


martin-henz:
<img width="558" alt="Screen Shot 2023-04-09 at 10 23 54 PM" src="https://user-images.githubusercontent.com/30220768/230778408-9adc06a2-a9de-4be8-8ff1-622b2f0f5c6a.png">

Instead of bullet points, we should indicate the section, such as

Foreword: Primitive means...

1: However it is the ability...

1.1.2: Constant declaration...

yiwen101:
added in #2486

PR #2486, opened 2023-05-17
Fix some issues with SICP search

yiwen101:
### Description

fix the bug that index search can not function as expected (return no result for every query)

add section information to search result (#2434)


RichDom2185 (approved):
LGTM! Thanks for the slight refactoring and explanatory comments in the code as well, great work!

Issue #2479, opened 2023-05-12
Investigate how we can improve CI times


RichDom2185:
Currently, the "Run CI" workflow takes up approximately 9-12 minutes. This is arguably quite slow. Can we do something to speed it up?

PR #2488, opened 2023-05-18
Parallelize Coveralls check in CI

RichDom2185:
Also adds a separate `yarn test` command step.

Further improves #2479.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        commands: [tsc, 'format:ci', eslint, build]
+        commands: [tsc, 'format:ci', eslint, build, test, test-coverage]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        commands: [tsc, 'format:ci', eslint, build]
+        commands: [tsc, 'format:ci', eslint, build, test, test-coveralls]
```
ianyong, line 19: I don't see any sharing of workflow artifacts between jobs, yet the `test-coveralls` job does run the test cases if we take a look at the run logs. Does this mean that the `test` job is redundant?

ianyong (changes requested):

.github/workflows/ci.yml
RichDom2185: It is somewhat redundant, based on what `yarn run test-coveralls` does (calls `./scripts/test-coveralls.sh` which calls `craco test`.

But I'm not too familiar with how `craco test` works, especially given the different flags that we pass in. Although they both run test cases, I'm not sure if the flags would result in different behaviour (e.g. some tests perhaps skipped?). Do you happen to be familiar with how it works?

Regardless, I don't think the redundancy affects much since they are all happening concurrently – does not add to the CI times.

ianyong: The total number of tests run seem to be the same for both jobs for the latest commit on this branch.

> Regardless, I don't think the redundancy affects much since they are all happening concurrently – does not add to the CI times.

Would be good to remove any redundancy but you're right about this. I'll approve this PR.

ianyong (approved):

Issue #2422, opened 2023-04-04
Sessions: Join arrow button not working


martin-henz:
<img width="1003" alt="Screen Shot 2023-04-04 at 11 45 21 AM" src="https://user-images.githubusercontent.com/30220768/229682210-b43ad436-5c19-4ac0-9950-9d442774bae1.png">

This button does not seem to work. The expected behavior (I guess) is that the entered session key becomes activated. This is not the case. Instead the user needs to hit "return" inside the text entry field for the session key.

(Chrome on mac)

shenyih0ng:
related to #2417, the switch to `AnchorButton` affects buttons with `type = submit` props

<img width="702" alt="image" src="https://user-images.githubusercontent.com/47914880/233608954-e064a1d5-5d0e-427f-87a7-cf5afe5e560c.png">

Issue #2417, opened 2023-04-03
Tooltips on disabled buttons do not go away on Chromium browsers


ianyong:
This is not an issue on Firefox.

![image](https://user-images.githubusercontent.com/5585517/229507022-3e5aa166-ae4c-4bbf-b9f6-bfdf7be75063.png)

martin-henz:
Some discussion of this: https://github.com/ant-design/ant-design/issues/9581
not sure if this helps.

ianyong:
> Some discussion of this: [ant-design/ant-design#9581](https://github.com/ant-design/ant-design/issues/9581) not sure if this helps.

Thank you! I managed to fix it by using an `AnchorButton` instead of a `Button` for the control buttons. Detailed explanation in the upcoming PR.

PR #2489, opened 2023-05-22
Fix sessions button

vansh284:
### Description

This PR aims to resolve https://github.com/source-academy/frontend/issues/2422. As mentioned by @shenyih0ng this issue is related to the switch to `AnchorButton`. Blueprintjs recommends the use of `AnchorButton` for tooltips over disabled buttons and so is required to resolve https://github.com/source-academy/frontend/issues/2417. However `AnchorButton` does not support submit buttons. 

The proposed solution is to use `AnchorButton` to render the `ControlButton` component only for disabled buttons. Buttons that are not disabled continue to use the regular Blueprintjs `Button`. Since `AnchorButton` and `Button` look the same, this shouldn't pose any problems. For reference https://blueprintjs.com/docs/#core/components/button.

Another possible and 'hacky' fix would be wrapping the `AnchorButton` with a HTML 'submit' button (refer to my first commit) but I prefer the current fix,

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/ControlButton.tsx
+++ b/src/commons/ControlButton.tsx
@@ -37,7 +37,7 @@ const ControlButton: React.FC<ControlButtonProps> = ({
   const buttonOptions: ButtonOptions = { ...defaultOptions, ...options };
   const iconElement = icon && <Icon icon={icon} color={buttonOptions.iconColor} />;
 
-  return (
+  const anchorButton: JSX.Element = (
     <AnchorButton
       disabled={isDisabled}
       fill={buttonOptions.fullWidth}
@@ -52,6 +52,17 @@ const ControlButton: React.FC<ControlButtonProps> = ({
       {label}
     </AnchorButton>
   );
+
+  if (buttonOptions.type) {
+    // Hacky fix to allow button types to work with anchor buttons
+    return (
+      <button type={buttonOptions.type} style={{ padding: '0px', border: 'none' }}>
+        {anchorButton}
+      </button>
+    );
+  }
+
+  return anchorButton;
 };
 
 export default ControlButton;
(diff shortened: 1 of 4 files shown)


vansh284:
> Is there any reason we want to keep using `Button` instead of `AnchorButton` when `isDisabled` is false?


`Button` allows passing in the HTML button type attribute so we are able to use our `ControlButton` component to submit form data. This does not work on `AnchorButton`. This is what caused the bug in the join sessions button in the first place.
image

When `isDisabled` is true, no functionality is expected from the button, so we can switch to `AnchorButton` to allow the tooltips to function properly.

src/commons/ControlButton.tsx
```diff
@@ -37,8 +37,26 @@ const ControlButton: React.FC<ControlButtonProps> = ({
...
+        type={buttonOptions.type}
+        onClick={onClick}
+        icon={!buttonOptions.iconOnRight && iconElement}
+        rightIcon={buttonOptions.iconOnRight && iconElement}
+      >
+        {label}
+      </AnchorButton>
+    );
+  }
+
```
RichDom2185, line 57:
```suggestion
  // Refer to #2417 and #2422 for why we conditionally
  // set the button component. See also:
  // https://blueprintjs.com/docs/#core/components/button
  const ButtonComponent = isDisabled ? AnchorButton : Button;
```

src/commons/ControlButton.tsx
```diff
@@ -37,8 +37,26 @@ const ControlButton: React.FC<ControlButtonProps> = ({
...
   return (
-    <AnchorButton
+    <Button
```
RichDom2185, line 59:
```suggestion
    <ButtonComponent
```

src/commons/ControlButton.tsx
```diff
@@ -50,7 +68,7 @@ const ControlButton: React.FC<ControlButtonProps> = ({
       rightIcon={buttonOptions.iconOnRight && iconElement}
     >
       {label}
-    </AnchorButton>
+    </Button>
```
RichDom2185, line 71:
```suggestion
    </ButtonComponent>
```

RichDom2185 (changes requested):
Let's try to reduce duplicated code here.

RichDom2185 (approved):
LGTM, thanks!

PR #2490, opened 2023-05-24
Game: Fix prompt buttons going offscreen

Eola-Z:
### Description

Fixed an issue where if there are more than 5 button choices in a dialog prompt, the buttons cut into the dialog box and eventually go offscreen.

### Type of change

### Checklist

- [x] I have tested this code


src/features/game/effects/Prompt.ts
```diff
@@ -49,6 +49,10 @@ export async function promptWithChoices(
   choices: string[]
 ): Promise<number> {
   const promptContainer = new Phaser.GameObjects.Container(scene, 0, 0);
+  
+  
+  const promptPartitions = Math.ceil(choices.length / 5);
```
lhw-1, line 54: I think these blank lines can be removed?

lhw-1 (commented):
Aside from the nit, LGTM - thanks for the fix!

Firstly, remember to update the documentation - specifically, [here](https://github.com/source-academy/general/blob/master/educator/game/README.md#prompts).

Also, this is an example of what it currently looks like:

Screenshot from 2023-05-24 16-06-07

Since we now have columns for prompts, I feel it would be better to have the prompts display from left to right instead of right to left - what do you think?

lhw-1:
Can I double check if the prompt ordering that was brought up above has been addressed? What's the status / decision?

Eola-Z:
Yes, the prompt ordering should go from left to right now.
Capture

lhw-1 (approved):
LGTM! Thanks for the PR @Eola-Z @eyelessrhyme7 :)

Eola-Z:
Thank you so much for the feedback, unless there is anything else of note, I'll go ahead and merge it?

PR #2492, opened 2023-05-24
Refactor state handlers memoization

RichDom2185:
### Description

~~Also uses `propsRef.current.XXX` instead of `props.XXX` as much as possible, as the former persists across re-renders, making dependency arrays cleaner.~~ (backtracked)

~~Ideally, if we migrate everything to use `useTypedSelector` instead of `connect`, we will have single source of truth for the workspace state types (as defined in `WorkspaceTypes.ts`)~~ (done in #2495 – thanks @chownces!)

Also follows up and resolves comments in #2495.

### Type of change

### Checklist

- [x] I have tested this code


RichDom2185:
I was looking at https://react.dev/reference/react/useRef (under the "Pitfall" section) and realise that this actually seems to be the wrong thing to do. I wonder why that change was done here (https://github.com/source-academy/frontend/pull/1449 – commit 1ed69fec88697c9079f141938a32fed6a839e7d5). Having said that, I tried it does seem to reduce the re-renders. Perhaps someone more familiar with `useRef` can advise on this.

RichDom2185:
Blocked by #2495. That PR should take priority.

chownces (approved):

Issue #2499, opened 2023-06-08
Improve performance


RichDom2185:
Our lighthouse score is really bad:

<img width="635" alt="image" src="https://github.com/source-academy/frontend/assets/34370238/888429d7-4c23-4e21-9db7-f9563bf96728">

Not to mention, our main bundle size as-is is already 15.7MB, which can take forever to load; we will hit the current 16MB limit in the near future (in fact, we already did simply by upgrading dependencies – see the CI failure on #2498).

We should try to do code-splitting or investigate some other way to improve the performance.

RichDom2185:
> After #2507, here is the breakdown of our bundles (via `yarn analyze`).
> 
> <img alt="image" width="1706" src="https://user-images.githubusercontent.com/53928333/245817179-e149d6ac-7ea2-4c93-adf4-8c51a6abc79f.png">

What's the use of `typescript.js`? At first glance, it seems like this is more of a dev dependency than a prod one.

Issue #2506, opened 2023-06-14
Remove 'disabled periods' feature


chownces:
We can disable a frontend deployment of Source Academy via the `.env` configuration for users with the `student` role.

Previously, this was used to disable the NUS and public deployments during examinations, under the assumption that students cannot have the `staff` or `admin` role.

However, anyone can be `staff` or `admin` by creating a new course (even in the disabled configuration), and circumvent the disabled configuration.

Would we still be using the disable feature in the future? If so, more safeguards need to be in place to ensure students cannot access the academy during disabled periods. Else we can remove this feature entirely if unused @martin-henz

martin-henz:
Need to check with the teaching team. Will get back on this.

martin-henz:
The authoritative answer from Prof Kok Lim: 

We probably had "taken down" Source Academy for the Reading Assessments in 2018 and/or 2019, but have not done that subsequently. I don't think we need this feature anymore.

martin-henz:
So I think we can safely remove this feature.

martin-henz:
Yes, now that SA is multi-tenant and used by places outside of NUS, we shouldn't have a feature that globally disables the frontend for all users, I think.

I'm proposing to permanently remove this feature. Details of the feature are in the README here: https://github.com/source-academy/frontend#disable-periods

PR #2507, opened 2023-06-14
Code splitting

chownces:
### Description

- Code splitting and lazy loading to reduce max bundle size. #2499
- Added `yarn analyze` to view breakdown of bundles
- More can be done to reduce the bundle size arising from js-slang dependencies, which are unused most of the time
  - ts-morph (for typed source variant) - 4.07MB
  - alt-ergo (for infinite loop detection) - 1.61MB
  - results in our main.js bundle being 10.76MB

Followup:
- We can consider lazy loading parts of the `Academy` itself. But this would involve further refactoring to hoist the academy routes to the top level router config in `routerConfig.tsx`

image

### Type of change

- [x] Performance improvements

### How to test

- Check that components load lazily (e.g. throttle the internet connection via browser dev panel)

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/pages/disabled/Disabled.tsx
+++ b/src/pages/disabled/Disabled.tsx
@@ -2,21 +2,31 @@ import { Classes, NonIdealState } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import classNames from 'classnames';
 import * as React from 'react';
+import { useLoaderData } from 'react-router';
 
 interface DisabledProps {
   reason?: string;
 }
 
-const Disabled: React.FC<DisabledProps> = ({ reason }) => (
-  <div className={classNames('NoPage', Classes.DARK)}>
-    <NonIdealState
-      icon={IconNames.ERROR}
-      title="Disabled"
-      description={`The Source Academy has been disabled${
-        reason ? ` for this reason: ${reason}` : ''
-      }.`}
-    />
-  </div>
-);
+const Disabled: React.FC<DisabledProps> = () => {
+  const disabledReason = useLoaderData() as string | undefined;
+
+  return (
+    <div className={classNames('NoPage', Classes.DARK)}>
+      <NonIdealState
+        icon={IconNames.ERROR}
+        title="Disabled"
+        description={`The Source Academy has been disabled${
+          disabledReason ? ` for this reason: ${disabledReason}` : ''
+        }.`}
+      />
+    </div>
+  );
+};
+
+// react-router lazy loading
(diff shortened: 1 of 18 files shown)


src/pages/disabled/Disabled.tsx
```diff
@@ -2,21 +2,31 @@ import { Classes, NonIdealState } from '@blueprintjs/core';
...
-    <NonIdealState
-      icon={IconNames.ERROR}
-      title="Disabled"
-      description={`The Source Academy has been disabled${
-        reason ? ` for this reason: ${reason}` : ''
-      }.`}
-    />
-  </div>
-);
+const Disabled: React.FC<DisabledProps> = () => {
```
RichDom2185, line 11: It seems that we can simply remove `DisabledProps` here. Is there a way we can still check that the component is rendered properly, with the correct reason?

src/routes/routerConfig.tsx
```diff
@@ -26,60 +15,60 @@ import Welcome from '../pages/welcome/Welcome';
...
   },
   {
     path: 'sicpjs/:section?',
-    element: <Sicp />
+    lazy: () => import('../pages/sicp/Sicp')
   },
   {
     path: 'githubassessments/*',
-    element: <GitHubClassroom />,
+    lazy: () => import('../pages/githubAssessments/GitHubClassroom'),
```
RichDom2185, line 71: Would it be better if we define all of the imports at the top-level, for better readibility? E.g.

```tsx
const Contributors = () => import('../pages/contributors/Contributors');
const GitHubCallback = () => import('../pages/githubCallback/GitHubCallback');
const Sicp = () => import('../pages/sicp/Sicp');
const GitHubClassroom = () => import('../pages/githubAssessments/GitHubClassroom');
// and so on...
```

and then, later on:

```tsx
{
  // ...
  lazy: Contributors // or GitHubCallback or Sicp or...
}
```

RichDom2185 (commented):
Thanks a lot for working on this! Some comments below, let me know your thoughts! 😄

src/pages/disabled/Disabled.tsx
chownces: yup i updated the testcases to reflect that. also #2506 might be relevant

src/routes/routerConfig.tsx
chownces: yup good idea, was deciding whether to do this haha, but did not since these imports are mostly used once

will update this tmr!

RichDom2185 (approved):
LGTM, thanks for working on this!

PR #2509, opened 2023-06-16
Refactor navbar

chownces:
### Description

- Fix CSS issues from react-router upgrade
- Improve readability of navbar component

### Type of change

### Checklist

- [x] I have tested this code


src/commons/navigationBar/subcomponents/AcademyNavigationBar.tsx
```diff
@@ -152,4 +51,86 @@ export const icons: IconName[] = [
...
+      disabled: !isAdmin,
+      hiddenInBreakpoints: ['xs', 'sm', 'md', 'lg']
+    },
+    {
+      to: `/courses/${courseId}/adminpanel`,
+      icon: IconNames.SETTINGS,
+      text: 'Admin Panel',
+      disabled: !isAdmin,
+      hiddenInBreakpoints: ['xs', 'sm', 'md', 'lg']
+    }
```
RichDom2185, line 132: Are the ordering changes (originally they were XP, Grading, Story, Admin Panel) intentionally part of the UX improvements?

src/commons/navigationBar/subcomponents/AcademyNavigationBar.tsx
```diff
@@ -152,4 +51,86 @@ export const icons: IconName[] = [
...
+      text: 'Story Simulator',
+      disabled: !isStaffOrAdmin,
+      hiddenInBreakpoints: ['xs', 'sm', 'md']
+    },
+    {
+      to: `/courses/${courseId}/xpcalculation`,
+      icon: IconNames.CALCULATOR,
+      text: 'XP Calculation',
+      disabled: !isAdmin,
+      hiddenInBreakpoints: ['xs', 'sm', 'md', 'lg']
```
RichDom2185, line 124: Likewise, originally, XP Calculation was only hidden in the `xs` and `sm` breakpoints. Is this change intentional as part of the improvements to readability of NavBar? I'm not sure if this refers to code readability of UX improvements. It definitely does improve the code readability, but in terms of the UX, was this an intentional change? Just making sure haha

src/commons/navigationBar/subcomponents/AcademyNavigationBar.tsx
```diff
@@ -152,4 +51,86 @@ export const icons: IconName[] = [
...
+export const assessmentTypesToNavlinkInfo = ({
+  assessmentTypes,
+  courseId,
+  isEnrolledInACourse
+}: {
+  assessmentTypes?: string[];
+  courseId?: number;
+  isEnrolledInACourse: boolean;
+}): NavbarEntryInfo[] =>
+  (assessmentTypes || []).map((assessmentType, idx) => ({
```
RichDom2185, line 63:
```suggestion
  assessmentTypes = [],
  courseId,
  isEnrolledInACourse
}: {
  assessmentTypes?: string[];
  courseId?: number;
  isEnrolledInACourse: boolean;
}): NavbarEntryInfo[] =>
  assessmentTypes.map((assessmentType, idx) => ({
```

src/commons/navigationBar/subcomponents/AcademyNavigationBar.tsx
```diff
@@ -152,4 +51,86 @@ export const icons: IconName[] = [
...
+    to: `/courses/${courseId}/${assessmentTypeLink(assessmentType)}`,
```
RichDom2185, line 64: What happens if `courseId` is `undefined`?

src/commons/navigationBar/NavigationBar.tsx
```diff
@@ -65,277 +55,222 @@ const NavigationBar: React.FC = () => {
...
+        )}
+        <NavLink
+          className={classNames('NavigationBar__link', Classes.BUTTON, Classes.MINIMAL)}
+          to={
+            Constants.playgroundOnly ? '/' : courseId == null ? '/welcome' : `/courses/${courseId}`
+          }
+        >
+          <Icon icon={IconNames.SYMBOL_DIAMOND} />
+          <NavbarHeading style={{ paddingBottom: '0px' }}>
+            {courseShortName || 'Source Academy'}
```
RichDom2185, line 122: Can we define a default fallback value in `Constants.ts` instead of hardcoding it here? It would make it easier for not only customisation, but also i18n efforts.

src/commons/navigationBar/NavigationBar.tsx
```diff
@@ -65,277 +55,222 @@ const NavigationBar: React.FC = () => {
...
-          {desktopLogoButton}
+          <NavLink
+            className={classNames('NavigationBar__link', Classes.BUTTON, Classes.MINIMAL, {
+              [Classes.ACTIVE]: highlightDesktopLogo(location)
+            })}
+            to={courseId == null ? '/welcome' : `/courses/${courseId}`}
+          >
+            <Icon icon={IconNames.SYMBOL_DIAMOND} />
+            <NavbarHeading style={{ paddingBottom: '0px' }}>
+              {courseShortName || 'Source Academy'}
```
RichDom2185, line 261: Ditto. Noticed that "Source Academy @ NUS" is also no more. Perhaps the default fallback value can be done conditionally?

RichDom2185 (commented):
Looks good! The codebase is so much more readable now 😄 I only have some minor comments as follows.

Thanks for this long-overdue refactor!

src/commons/navigationBar/subcomponents/AcademyNavigationBar.tsx
chownces: yes. lets keep the admin related routes to the right

src/commons/navigationBar/NavigationBar.tsx
chownces: This is done on purpose to take out the hardcoded @ NUS. Replaced with a more general purposed 'Source Academy'

src/commons/navigationBar/subcomponents/AcademyNavigationBar.tsx
chownces: This component is only rendered when `role` is defined, in which case `courseId` should also be defined. To make it more robust, I added an additional early return to not render this component when `courseId` or `role` is undefined

src/commons/navigationBar/NavigationBar.tsx
RichDom2185: Noted, that makes sense. I'm just thinking of potential integrations with i18n, hence my suggestion on putting this into `Constants.tsx` or something like that. But on second though, let us not prematurely do all these refactoring. Thanks!

RichDom2185 (approved):
LGTM, thanks!

PR #2518, opened 2023-06-18
Standardise MobileWorkspace and Workspace props

RichDom2185:
### Description

Part of #2515. Fixes #2516.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Code quality improvements

### Checklist

- [ ] I have tested this code
- [ ] I have updated the documentation

PR #2523, opened 2023-06-20
Improve workspace code quality

RichDom2185:
### Description

* Refactor Sourcecast handlers with a similar pattern to #2511
* Use abstracted `workspaceLocation` value in workspaces.

### Type of change

### Checklist

- [x] I have tested this code


src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
```diff
@@ -516,7 +516,7 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
         afterDynamicTabs: []
       },
       onChange: onChangeTabs,
-      workspaceLocation: 'assessment'
+      workspaceLocation: workspaceLocation
```
chownces, line 519:
```suggestion
      workspaceLocation
```

chownces (approved):

src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
RichDom2185: Hmm I don't think this is really needed (also separating it this way helps when ctrl-clicking). Perhaps we can take another look at this after the Workspace and MobileWorkspace props have been refactored (#2518).

PR #2560, opened 2023-07-26
Implement Stories Search Functionality

gracelimshiern:
### Type of change

### Checklist

- [ ] I have tested this code


src/pages/stories/Stories.tsx
```diff
@@ -82,11 +94,11 @@ const Stories: React.FC = () => {
             </TableRow>
           </TableHead>
           <TableBody>
-            {data
+            {filteredData
               // .filter(story => story.authorId.toLowerCase().includes(query)
```
RichDom2185, line 98: ...just uncomment it here.

RichDom2185 (changes requested):
As discussed outside of GitHub, this is the wrong approach, `filteredData` is derived state from `data`, so no need `useState`...

src/pages/stories/Stories.tsx
```diff
@@ -30,18 +29,11 @@ type StoryListView = {
 };
 
 const Stories: React.FC = () => {
-  // const [user, setUser] = useState<string>('');
   const [data, setData] = useState<StoryListView[]>([]);
-  // const [query, setQuery] = useState('');
+  const [query, setQuery] = useState<string>('');
```
RichDom2185, line 33: We don't have to manually specify the type, we can use type inference;

```suggestion
  const [query, setQuery] = useState('');
```

src/pages/stories/Stories.tsx
```diff
@@ -51,7 +43,9 @@ const Stories: React.FC = () => {
 
   useEffect(() => {
     getStories().then(res => {
-      res?.json().then(setData);
+      res?.json().then(data => {
+        setData(data);
+      });
```
RichDom2185, line 48: No need to change this, the extra function is not necessary.

RichDom2185 (changes requested):

RichDom2185 (approved):
LGTM

PR #2575, opened 2023-07-27
Allow overriding request host in request helper

RichDom2185:
Done to support multiple backends.

### Description

In order to pipe API requests to the stories backend (and potentially more backends in the future) through the request helper, we will need a way to override the URL.

The alternative would be a major refactor which I don't think is worth the effort at the moment.

### Type of change

### Checklist

- [x] I have tested this code


YaleChen299 (approved):

RichDom2185:
As discussed offline, we will merge first as this is blocking; my proposed use case for this change involves another wrapper around this function so it would require very minimal refactoring should anything be raised later.

PR #2589, opened 2023-08-01
Add UI changes for questions that are unanswered

emptygx:
### Description

Previously, questions that were not modified by students could cause inefficiency for graders when sifting through the template code to look for any modifications or code to grade, especially for assignments with long code. This solves this by making it clear for the grader if the question was not answered through the following improvements:

- Checks if any changes were made to the answer template
- Ignores and whitespace changes
- Displays a prompt when grading such unanswered questions
- Adds a header to the top of editor to emphasise blank template

### Type of change

### How to test

No special configurations, test with an empty submission and check if the grading workspace flags out questions with no modifications to the template code.

### Checklist

- [x] I have tested this code


RichDom2185 (commented):
Direction looks good, just one comment: can we reduce the code duplication?

* Assign the regex into a variable
* Assign the prefix to a variable

emptygx:
> Direction looks good, just one comment: can we reduce the code duplication?
> 
> * Assign the regex into a variable
> * Assign the prefix to a variable

Done with this! @RichDom2185

src/pages/academy/grading/subcomponents/GradingWorkspace.tsx
```diff
@@ -238,6 +257,16 @@ const GradingWorkspace: React.FC<GradingWorkspaceProps> = props => {
       if (!editorValue) {
         editorValue = questionData.solutionTemplate!;
       }
+      if (
+        editorValue.trim() === questionData.solutionTemplate?.trim() ||
+        editorValue.replace(whitespaceRegex, '') ===
```
RichDom2185, line 262: Might be better to use `replaceAll` instead of `replace` as it is safer (it will throw a `TypeError` if the regex passed in is not created with the `/g` flag).

RichDom2185 (commented):
I just realised that this approach treats the entire program as a single string. While generally JS/Source is not whitespace-sensitive, the same can't be said for other programming languages (notably Python).

Since we aim to be language-agnostic, this should not be the de-facto check for all languages.

Moreover, another bug is with strings. This implementation would consider a change from `""` to `"      "` (or any other string change involving only whitespaces) as insignificant. This is not strictly correct.

I'll withhold this from being merged for now while we discuss this. Thanks for sparking this discussion!

Q: Is it possible to get the answer status (attempted/not attempted/etc.) from the backend – or perhaps it's already in the state somewhere? Can we possibly use that (just ideas, not asking you to implement)?

emptygx:
> e aim to be language-a



> I just realised that this approach treats the entire program as a single string. While generally JS/Source is not whitespace-sensitive, the same can't be said for other programming languages (notably Python).
> 
> Since we aim to be language-agnostic, this should not be the de-facto check for all languages.
> 
> Moreover, another bug is with strings. This implementation would consider a change from `""` to `" "` (or any other string change involving only whitespaces) as insignificant. This is not strictly correct.
> 
> I'll withhold this from being merged for now while we discuss this. Thanks for sparking this discussion!
> 

I see, in that case, would it be okay if we remove the extra check on whitespace and for now keep it simple to just check if the solution matches the template exactly? This was what Prof and I discussed.

> Q: Is it possible to get the answer status (attempted/not attempted/etc.) from the backend – or perhaps it's already in the state somewhere? Can we possibly use that (just ideas, not asking you to implement)?

From my understanding, there is no answer status because an unattempted question simply would not exist in the database, i.e. if the question was not saved and auto-submission occurs after the deadline. However, questions which were simply saved without any edits are currently not detected by the backend and this solution checks it at the frontend to provide cosmetic alerts.

RichDom2185:
> I see, in that case, would it be okay if we remove the extra check on whitespace and for now keep it simple to just check if the solution matches the template exactly? This was what Prof and I discussed.

Ok sure, let's proceed with that then. Thanks!

RichDom2185 (approved):
LGTM, thanks!

PR #2593, opened 2023-08-03
Fix syntax highlighting

### LINKED ISSUE #2586: Syntax highlighting is broken
~~Suspect something to do with client-side routing.~~

~~Request URL should be correct (I think):~~

image

~~But the request seems to redirect to our main app instead, instead of fetching the asset:~~

image

~~Looking at https://github.com/source-academy/sourceacademy.org, it seems like the requested asset file does not actually exist in the `static` folder, which would make sense why it falls back to the main app (for client-side routing).~~

**EDIT:** Refer to comments below for cause.

RichDom2185:
### Description

Manually reverts the changes in `package.json` and `yarn.lock` for `ace-builds` that were introduced in #2545 (the cause of the regression below).

There is no need for us to be in the bleeding edge so we will just rollback for now.

Closes #2586.

### Type of change


martin-henz (approved):
Looks good!

PR #2605, opened 2023-08-06
Better Tooltips for Instructions on the Agenda

vansh284:
### Description

Feedback was received that it was not immediately obvious how the 'RESET' and 'MARKER' instructions interacted with each other and how they served the explicit control evaluator. The tooltips for these instructions were to be improved.

I realized that all the instructions on the agenda could benefit from having more descriptive tooltips so that students can remind themselves of the functions of the different instructions when necessary.

### Type of change


RichDom2185 (approved):
LGTM

PR #2614, opened 2023-08-08
Restore some stories tabs

RichDom2185:
### Description

* Restore data visualiser, HTML display, and stepper tabs
* Fix incorrect stories reducer (it was missing the hotfix from #2305)

There are still some bugs remaining, but nothing too major:

* The stepper runs the code chunk as if it were a fresh program, i.e. without any prepends. Thus we need to ensure that everything is defined in the same code block.
  * e.g. we can't have function definitions in Block 1, and their function calls in Block 2, and run the stepper from Block 2.
* _(only affects authors, not viewers) Updating chapters/variants is not being propagated correctly into the component lifecycle. If you change a code block's environment from Source 4 to Source 2 (or something that does not support the stepper to something that does), you will need to save the file and refresh the page in the browser for the stepper to work.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/features/stories/StoriesReducer.ts
+++ b/src/features/stories/StoriesReducer.ts
@@ -106,9 +106,10 @@ export const StoriesReducer: Reducer<StoriesState> = (
         }
       };
     case EVAL_STORY_SUCCESS:
+      const execType = state.envs[env].context.executionMethod;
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Looks good!

PR #2625, opened 2023-08-16
reset mcq autograding result to empty list

YaleChen299:
### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### How to test
 I tested locally with and without this line.

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #2517, opened 2023-06-18
Use `useTypedSelector` and `useDispatch` in AssessmentWorkspace

RichDom2185:
### Description

Part of #2514.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### Checklist

- [ ] I have tested this code
- [ ] I have updated the documentation

PR #2627, opened 2023-08-16
Fix assessment workspace runtime error

RichDom2185:
### Description

There was a regression in #2517 when I refactored the naked variables into a single config object – I missed out on setting fallback values (as they were no longer initialised).

This resulted in a runtime error when opening assessments where autograding is enabled and contains a multiple-choice question.

In the process of restoring the old behavior, I also noticed that the type guard function can produce incorrect results when checking the type of non-optional values. Thus, this PR also fixes the `assertType` function so that undefined values, unless the underlying type being checked against also allows for them, will result in a compile-time error.

This should prevent reoccurrence in the future.

Supersedes #2625.

### Type of change

### Checklist

- [x] I have tested this code


YaleChen299 (approved):
Thank you for the fix!

PR #2628, opened 2023-08-16
wrap action with function in useEffect

YaleChen299:
### Description

Fix the issue that visiting any other tab after visiting achievement causing runtime error.

This is caused by the useEffect directly using the action, which might be async(?).
The fix is to wrap it.

### Type of change

### How to test

I have tested locally.

### Checklist

- [x] I have tested this code


RichDom2185 (approved):
Not because it's async, but simply because it's the wrong type: the `getUsers` function is equivalent to
```ts
() => dispatch(getUsers())
```
But `dispatch` does not return void (it returns the created action) so in the context of `useEffect` it will think that this returned action object is a clean up function and try to call it when the component unmounts (hence only when navigating away from the achievements tab).

LGTM, thanks!

Issue #2632, opened 2023-08-19
Disable stories edit page route when user does not have rights


RichDom2185:
While the permissions are already checked in the backend, we should also ideally disable the frontend route, especially since the frontend already as a notion of the users' roles in the context of stories.

PR #2642, opened 2023-08-25
Disable stories routes under public deployment

### LINKED ISSUE #2633: Disable stories routes in public deployment
The public deployment does not have support for stories yet at the moment.

sayomaki:
### Description

This PR attempts to resolve #2633 to disable the routes in the public deployment as stories are not yet available. The fix moves all stories-related routes to under `fullAcademy` router, instead of being under common routes where they would still be accessible in the public deployment.

Alternatively, adding a environment variable can also be considered instead, to toggle whether stories routes will be enabled (similar to how `githubAssessments` work right now). If this is the right way going forward, let me know and I will refactor the code accordingly.

Also as a side note, the stories component might also need to updated to have its own router instead, as right now they are statically defined in the main `routerConfig` file. This will also ease the implementation of #2632 where more fine-grained permissions control can be done within its own router. However, this is outside the scope of this PR and should be in its own standalone update. 

### Type of change

### How to test

Check if the routes are accessible when `REACT_APP_USE_BACKEND` is true and when is false (fullAcademy mode enabled and disabled).

The code has been tested locally, and confirmed that stories routes will no longer be present.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/routes/routerConfig.tsx
+++ b/src/routes/routerConfig.tsx
@@ -91,22 +91,6 @@ const commonChildrenRoutes: RouteObject[] = [
     path: 'githubassessments/*',
     lazy: GitHubClassroom,
     loader: conditionalLoader(!Constants.enableGitHubAssessments, '/')
-  },
-  {
-    path: 'stories/new',
-    lazy: EditStory
-  },
-  {
-    path: 'stories/view/:id',
-    lazy: ViewStory
-  },
-  {
-    path: 'stories/edit/:id',
-    lazy: EditStory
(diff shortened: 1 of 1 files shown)


chownces (approved):
Thanks for looking into this! This implementation is okay for now, but we will eventually have a separate public stories backend in the future.

On a side note, we have a Source Academy Dev Team chat to coordinate development work. Would you be interested to join in? Feel free to drop me an email at chowenrong@u.nus.edu!

PR #2437, opened 2023-04-11
Redesign Avenger dashboard

zhyuhan:
## Description

This is a major redesign of the dashboard at `/grading` to improve performance and speedup the workflow of Avengers. This is down by migrating the existing submissions table from [AG Grid](https://www.ag-grid.com) to the headless [TanStack Table](https://tanstack.com/table/v8) and components from [Tremor](https://www.tremor.so).

The demo video recorded for the final Show-and-Tell can be found [here](https://www.loom.com/share/7b96695071e246e5b19fb21a333cb5b2).

## Changes

The changes in this PR include:
- Rewrite submissions table using TanStack Table and Tremor
- Improve column and global filtering
  - Persist filters in Redux state (closes #2298, closes #737)
- Add grading statistics (number of graded/ungraded submission and ungraded assessments)
- Refactor `pages/academy/grading/Grading.tsx` to a functional component
- Implement custom `exportCSV` function (previously part of AG Grid API)
  - A few columns that don't exist in the data but exported to CSV with empty values (`initialGrade`, `gradeAdjustment`, `currentGrade`, maxGrade`) are removed, is this okay @martin-henz?

## Future improvements

There are still a few existing issues that can be improved:
- The submission/grading statuses cannot be used to tell whether a submission has been unsubmitted by the Avenger (see #2189)
  - Currently when a submission is unsubmitted, the submission status stays as `submitted` and the grading status stays as `graded`, while the XP adjustments get reverted. There should be a better way to tell whether a submission has been unsubmitted or not.
- Allow filtering by assessments through clicking items in the "Ungraded assessments" list
  - The new grading stats allow an Avenger to see which assessments still have ungraded submission. It will be good if they can click on one of them to add a filter.
- Migrate Tremor to v2 (see [changelog](https://www.tremor.so/changelog))
  - Tremor v1 uses internal [Tailwind](https://tailwindcss.com) stylesheets for styling. In v2, the recommended styling solution is to install our own Tailwind. We will have to resolve a lot of style conflicts if we were to upgrade to v2.
- This redesign is also a first step to migrating away from `ag-grid-community` and `ag-grid-react`, which have a combined bundle size of 33MB.

PR #2643, opened 2023-08-25
Refactor grading dashboard and reduce API overhead

RichDom2185:
### Description

Mainly refactored code, but also added new custom components to support minimal dropdowns, and also restored the old behaviour of allowing staff to toggle between viewing only submissions from their own group (now the default) or all groups. This was previously removed in #2437.

With over 100 groups, this would result in a >100x performance improvement.

New UI:

image

### Type of change

### Checklist

- [x] I have tested this code


chownces (approved):
lgtm!

PR #2647, opened 2023-09-02
Fix breakpoints state going out of sync

### LINKED ISSUE #2646: Spurious invisible breakpoints are causing errors
photo_2023-09-02_15-03-42

The only ways to recover at the moment are:
* try to click on the gutter at the offending line, or if that fails
* visit Source Academy in incognito mode

Here is a recording that shows that the breakpoint remains active when the line is deleted.

https://github.com/source-academy/frontend/assets/30220768/dc678db5-d7ec-4f9a-85a5-16168b9d4610

(The breakpoint disappears when the line is reentered.)

ianyong:
### Description

In #2269 & #2346, the behaviour of breakpoints was updated to be consistent with most code editors & IDEs available. However, due to the way the `Editor` component and the workspaces are implemented, the `Editor` maintains its own source of truth that is separate from the Redux store. The bug in #2646 occurs when the breakpoints state in the Redux store and the `Editor` component become unsynchronised.

To fix this issue, we write the current `Editor` breakpoints state to the Redux store whenever the `Editor` value updates. Ideally, the `Editor` should rely on the Redux store as the single source of truth, but this would require major refactoring and I'm not confident of not breaking other things along the way due to the lack of comprehensive tests + the rather hacky implementation of the `Editor`.

Fixes #2646.

### Type of change

### How to test

Replicate

### Checklist

- [x] I have (manually) tested this code

Ideally, I would write regression tests here, but testing this particular logic requires some test framework that can simulate user input such as Cypress (and setting up such a test framework is beyond the scope of this PR).

### DIFF
--- a/src/commons/editor/Editor.tsx
+++ b/src/commons/editor/Editor.tsx
@@ -411,10 +411,7 @@ const EditorBase = React.memo((props: EditorProps & LocalStateProps) => {
     // hopelessly incomplete
     editor.on(
       'gutterclick' as any,
-      makeHandleGutterClick(
-        (...args) => handleEditorUpdateBreakpointsRef.current(...args),
-        props.editorTabIndex
-      ) as any
+      makeHandleGutterClick(handleEditorUpdateBreakpointsRef.current, props.editorTabIndex)
     );
 
     // Change all info annotations to error annotations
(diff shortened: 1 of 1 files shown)


src/commons/editor/Editor.tsx
```diff
@@ -485,8 +482,16 @@ const EditorBase = React.memo((props: EditorProps & LocalStateProps) => {
       if (!reactAceRef.current) {
         return;
       }
-      handleEditorValueChange(props.editorTabIndex, newCode);
+
       shiftBreakpointsWithCode(reactAceRef.current.editor, delta);
+
+      // Write editor state for the active editor tab to the store.
+      handleEditorValueChange(props.editorTabIndex, newCode);
+      handleEditorUpdateBreakpointsRef.current(
```
ianyong, line 490: No idea why this callback function is wrapped in a ref but I'm afraid that removing it will cause some other regression somewhere.

martin-henz (approved):
Looks good. Fixes the example in https://github.com/source-academy/frontend/issues/2646

Issue #570, opened 2019-05-30
Run button in Grading resets the editorValue


geshuming:
To reproduce: write non-erratic code and run the code. It may be due to the lack of UPDATE_EDITOR_VALUE action in grading.

Issue #350, opened 2018-09-04
Students should be able to see grades for individual tasks.


blackening:
Currently students can see the total grade and total xp, as well as comments for each task.

Students should be able to view their grade for each task.

jiachen247:
closed with #426

PR #2652, opened 2023-09-03
Fix unable to remove highlighting from code

### LINKED ISSUE #2645: Ace editor cursor change not being detected after multiple edit sessions introduced
To reproduce:
#### Bug 1
1. Highlight the scope of a name declaration via `Cmd-Shift-H` or `Ctrl-Shift-H`
2. Click somewhere else in the editor
3. The scope highlight does not disappear

#### Bug 2 (regression)
1. Place the cursor on a name in the program
2. The dotted boxes around the names belonging to the same declaration are not drawn

#### On further investigation...
- The `onCursorChange` method in `aceEditorProps` is not being called despite the cursor change
- This might be due to the editor session being changed
  - Either our implementation is leading to stale props & editor hooks
  - Or the library has a bug (unlikely)
- However, the `onChange` method is still being called properly when the editor value changes

Offending commit: https://github.com/source-academy/frontend/pull/2485/
```
// Comment out this line and the onCursorChange method works as expected
editor.setSession(props.session);
```

### Expected behaviour (`Cmd-Shift-H`, then clicking somewhere in the editor)
https://github.com/source-academy/frontend/assets/53928333/597bbd54-68ce-4828-80e8-f2ca5ca26995

### Bug 1
https://github.com/source-academy/frontend/assets/53928333/245dbc5c-8945-4ed1-a48d-01a79153dc27

### Bug 2
(Expected - dotted boxes around names belonging to the same declaration)
image

(Buggy behaviour - dotted boxes do not show up)
image

sayomaki:
### Description

This PR attempts to fix #2645 where the highlighting is stuck after the implementation to use multiple editor sessions, due to the `onCursorChange` event not firing properly. After much code testing and extensive troubleshooting, the issue has been narrowed down to a likely upstream library problem.

As seen from the source code in ReactAce ([`ace.tsx` on line 263](https://github.com/securingsincity/react-ace/blob/5e2290f716fa965ba70889cc94b4221ac93334ac/src/ace.tsx#L263)), the event registration for `changeCursor` (which is linked to `onChangeCursor`) is only performed once during the `componentDidMount` lifecycle stage. This means that, only the default session initially created together with the editor will have the events bound, and any other sessions (created by us) will not be registered with the events. Our implementation of sessions completely ignore the default built-in session, and is replaced with our own default session (see [`Editor.tsx`#570](https://github.com/source-academy/frontend/blob/7cea1313d591eed07f9265243a431db7a285d2e0/src/commons/editor/Editor.tsx#L350)). This causes the problem even when not in folder mode.

I have already opened an issue over at the upstream library (securingsincity/react-ace#1839), and will monitor the progress on the issue (although there seems to be little activity there as of writing). In the meantime, a hacky-ish workaround will be proposed to resolve this issue for now.

Therefore, the current proposed fix will try to bind the `onCursorChange` handler back onto the session every time the session changes (see [`Editor.tsx`#350](https://github.com/source-academy/frontend/blob/7cea1313d591eed07f9265243a431db7a285d2e0/src/commons/editor/Editor.tsx#L350)). However, a event listener check (that relies on some internal non-documented property) is also required, to ensure that the event is only registered once (for a total of 2 event listeners, one from the default ace editor and the handler from react-ace), preventing any memory/listener leakages.

The changes have been tested locally on the latest codebase, with the latest Chrome browser. Folder mode switching also appears to be working fine (which required the use of multiple `EditorSession` in the first place).

https://github.com/source-academy/frontend/assets/26355099/413d230a-6871-4524-b14d-efcf4d5fc2c7

### Type of change

### How to test

Checking whether moving the cursor after highlighting a code block with `Ctrl + Shift + H` (or `Cmd + Shift + H` on macOS) will un-highlight the block of code.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/editor/Editor.tsx
+++ b/src/commons/editor/Editor.tsx
@@ -348,6 +348,22 @@ const EditorBase = React.memo((props: EditorProps & LocalStateProps) => {
       // See AceHelper#selectMode for more information.
       props.session.setMode(editor.getSession().getMode());
       editor.setSession(props.session);
+      /* eslint-disable */
+
+      // Add changeCursor event listener onto the current session.
+      // In ReactAce, this event listener is only bound on component
+      // mounting/creation, and hence changing sessions will need rebinding.
+      // See react-ace/src/ace.tsx#263,#460 for more details. We also need to
+      // ensure that event listener is only bound once to prevent memory leaks.
+      // We also need to check non-documented property _eventRegistry to
+      // see if the changeCursor listener event has been added yet.
+
+      // @ts-ignore
+      if (editor.getSession().selection._eventRegistry.changeCursor.length < 2) {
+        editor.getSession().selection.on('changeCursor', reactAceRef.current!.onCursorChange);
+      }
+
+      /* eslint-enable */
       // Give focus to the editor tab only after switching from another tab.
       // This is necessary to prevent 'unstable_flushDiscreteUpdates' warnings.
       if (filePath !== undefined) {


martin-henz (approved):
Nice temporary solution. Hopefully react-ace gets fixed soon based you your issue there.

PR #2660, opened 2023-09-03
Polyfill `path/posix` library when building

ianyong:
### Description

Node.js APIs are only available in Node.js environments (and not browser environments). In order to make use of Node.js APIs in the browser, Webpack includes polyfills of these APIs (libraries that contain the same or similar API but are implemented in pure JavaScript so that they can be run in the browser).

Webpack appears to polyfill the `path` module by default, but not the `path/posix` module. This PR adds the `path-browserify` library as a polyfill for `path/posix`. Note that the lock file does not change because Webpack already depends on `path-browserify`.

Fixes the build issue seen in #2655.

Relevant js-slang PR: https://github.com/source-academy/js-slang/pull/1470

### Type of change

### How to test

1. Bump js-slang to v1.0.34.
1. Run `yarn run build` and verify that there are no errors.

### Checklist

- [x] I have tested this code

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -46,6 +46,7 @@ const cracoConfig = (module.exports = {
         'http': require.resolve('stream-http'),
         'https': require.resolve('https-browserify'),
         'os': require.resolve('os-browserify/browser'),
+        'path/posix': require.resolve('path-browserify'),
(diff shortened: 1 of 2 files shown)


RichDom2185 (approved):
LGTM, thanks for the quick fix!

PR #2670, opened 2023-09-09
Use `SimpleDropdown` to reduce code duplication

RichDom2185:
### Description

Reduces code duplication.

### Type of change

### Checklist

- [x] I have tested this code


src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
```diff
@@ -35,29 +35,16 @@ const NavigationBarLangSelectButton = () => {
     setIsOpen(false);
   };
 
+  console.log('lang', lang); // FIXME: Not updating
+
```
RichDom2185, line 39: For some reason this change introduces a regression: The currently selected option is cleared on refresh (and defaults back to JS, even if the playground is already set to the correct language).

RichDom2185 (commented):
Just need to find out the root cause of this and fix, then we can merge.

sayomaki:
@RichDom2185 I see this PR is stalled for quite a while already, so I have gone ahead and did some simple analysis & investigation for the problem. 

The issue was related to how React tracks state and props, especially in the `useState` hook, as the initial value of `useState` is only considered during component **mount**, but otherwise discard in component **re-render**. Thus, this brings us to the question, does `SimpleDropdown` really needs to maintain its own state over the selected value?

If we look again at what state it is handling (`selectedValue`), we can see that it is not really needed, as any changes in `selectedValue` (that is triggered by the clicking of the inner `MenuItem` component) will be propagated upwards through the onClick event, which will eventually cause an update in the `lang` value (or any value that is tied to the current `defaultValue` prop on `SimpleDropdown`). [Here is something worth reading that goes into more detail for `useState` and props.](https://tkdodo.eu/blog/putting-props-to-use-state) Therefore, I suggest doing away with its own internal state for the `SimpleDropdown` wrapper, and to rename `defaultValue` to `selectedValue` for better representation of its functionality.

This will not affect how the dropdown works, as this is handled and managed by the inner `Popover2` component instead, which holds and controls its own state (to render or not render the dropdown) which is unrelated to the scope of the `SimpleDropdown` wrapper. 

As a side issue, the `MenuItem` component rendered within `SimpleDropdown` also does not have a key property, which did exist previously and is good practice to have one when dealing with multiple same component children.

I will be pushing and updating these changes accordingly, and this completes the refactor to `SimpleDropdown`, bringing this PR out of the (stalled) draft state. I have also updated the relevant files that currently uses `SimpleDropdown` (which is Grading.tsx), could you kindly help me check and verify that there are no regressions introduced caused by the change in the implementation of `SimpleDropdown`? 

As for the tests I'm not sure if my changes are the best way to go with, so do let me know your comments if you have any.

RichDom2185 (commented):
> ... could you kindly help me check and verify that there are no regressions introduced caused by the change in the implementation of `SimpleDropdown`?
> 
> As for the tests I'm not sure if my changes are the best way to go with, so do let me know your comments if you have any.

Looks good! I've also tested `Grading.tsx` locally and there seem to not be any regressions.

Thanks for helping out with this PR! Unfortunately, I can't approve it (because I authored the PR), but I can enable an auto-merge. Once you approve the PR, the auto-merge will automatically merge this PR into master.

Thanks once again!

sayomaki:
> Looks good! I've also tested `Grading.tsx` locally and there seem to not be any regressions.

Thanks! Alright, I'll wait for a 3rd person opinion on this (since I authored a significant portion of the code), and I think it should be auto-merged once approved with no outstanding issues.

PR #2675, opened 2023-09-11
Provide Module Side Content with more packages

leeyi45:
Source modules use several packages that are already available on the frontend. This means that these packages should not have to be bundled with Source modules. These packages are:
- `react` and `react/jsx-runtime` (both are necessary to maintain compatibility with both the old and new types of JSX transform)
- `react-dom`
- `react-ace`
- `@blueprintjs/core`,  `@blueprintjs/icons`,  `@blueprintjs/popover2`
- `js-slang`
If there are any other packages that fall under this category, I will modify this PR to further include them.

This PR makes these other packages available to the Source modules.

This PR also fixes an issue where even with folder mode disabled, files that had been previously created could still be imported

### DIFF
--- a/src/commons/editor/tabs/utils.ts
+++ b/src/commons/editor/tabs/utils.ts
@@ -1,4 +1,4 @@
-import * as path from 'path';
+import { posix as pathlib } from 'path';
 
 /**
  * Returns the shortest file paths that is uniquely identifiable among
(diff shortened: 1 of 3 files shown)


RichDom2185 (approved):
LGTM, thanks! Perhaps we can document which packages are accessible in the modules repo (something like "refer to file XXX for the list of available packages")?

leeyi45:
> LGTM, thanks! Perhaps we can document which packages are accessible in the modules repo (something like "refer to file XXX for the list of available packages")?

Good idea I'll update the modules wiki

PR #2681, opened 2023-09-18
Remove "excluded" grading items from ungraded pool

RichDom2185:
### Description

Closes #2662. Part of #2665. Refer to the linked issue for details.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #2686, opened 2023-09-22
Fix `GradingSubmissionsTable` and `GradingSummary` behavior

RichDom2185:
### Description

Follows up from #2681 with improvements to the code quality as well as fixing incorrect/inconsistent behaviour.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/features/grading/GradingUtils.ts
+++ b/src/features/grading/GradingUtils.ts
@@ -3,10 +3,10 @@ import { GradingStatuses } from 'src/commons/assessment/AssessmentTypes';
 import { GradingOverview } from './GradingTypes';
 
 export const isSubmissionUngraded = (s: GradingOverview): boolean => {
-  const isNotSubmitted = s.submissionStatus !== 'submitted';
+  const isSubmitted = s.submissionStatus === 'submitted';
   const isNotGraded =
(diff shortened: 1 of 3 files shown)


sayomaki (commented):
One small comment, if the excluded submissions have already been filtered by `Grading.tsx`, is it still required to be checked in `isSubmissionsUngraded`? Apart from that, everything looks great to me!

RichDom2185:
> One small comment, if the excluded submissions have already been filtered by `Grading.tsx`, is it still required to be checked in `isSubmissionsUngraded`? Apart from that, everything looks great to me!

`isSubmissionsUngraded` is also used in the `GradingSubmissionsTable`, which does not have this filter, so it's more of a utility.

In my view, I think that it's fine as the they represent different levels concerns; from within the abstraction boundary of each component, both are doing the right thing. It is up to the top level `Grading` component to decide what gets actually passed.

Also, worth to note that if we don't remove excluded in `Grading`, then, as `GradingSummary` only knows to remove ungraded submissions by calling `isSubmissionUngraded` (which it doesn't know the implementation of), it will give the wrong results of total number of submissions.

sayomaki (approved):
Alright, sounds good to me!

PR #2688, opened 2023-09-27
Manage Side Content using Redux

leeyi45:
This PR is mainly aimed at resolving the issues related to the flashing of the icons and persistence of side content across evaluations. Related to these issues in the modules repo:
- https://github.com/source-academy/modules/issues/147
- https://github.com/source-academy/modules/issues/171

I've tried three different solutions at this point:
- Using `@reduxjs/toolkit` to simplify our redux implementations: Unfortunately Redux Toolkit is very opinionated and mandates the use of Immer, which auto freezes objects in the store. This creates a myriad of issues, for example, since `js-slang` mutates the context object, which can't be done if it's frozen. To use this would require a major refactoring of the entire Redux state to ensure mutable objects and immutable objects are handled separately.
- Refactoring the Workspaces: I wanted to make the side content a property of a workspace, rather than an entirely separate Redux slice. However, this would also require a major refactor of the workspace slices, which, again, would have ramifications far beyond the scope of what I am trying to do here.
- The solution I have settled on is creating a separate slice for side content to attach to the `OverallState`. Then through the use of hooks and `mapDispatchToProps`, a workspace can indicate that a tab should display an alert, or reset an alert after a tab has been visited.
```ts
dispatch(actions.alertSideContent(workspaceLocation, SideContentType.dataVisualizer))
const { selectedTab, setSelectedTab } = useSideContent(workspaceLocation)
```
Stories can be specified as workspace locations using the given format: ``stories.${env}``. This abstracts away the handling of side content from the workspace.

I did consider a React context based approach, however because of the way the loading of side content is reliant on the evaluation of code, it was necessary to hook into the Redux Sagas, so I figured the best way to handle side content is through the store.

image
I've also added a small dev settings window as shown above. I wanted to be able to freely switch between the local modules server and the deployed one.
This screen is accessible via a dev settings button in the playground, which should only appear when `process.env.NODE_ENV === 'development'`


src/pages/playground/PlaygroundTabs.tsx
```diff
@@ -64,7 +39,10 @@ export const makeRemoteExecutionTabFrom = (
   id: SideContentType.remoteExecution
 });
 
-export const makeSubstVisualizerTabFrom = (output: InterpreterOutput[]): SideContentTab => {
+export const makeSubstVisualizerTabFrom = (
```
RichDom2185, line 42: Should we move all of these tabs under `PlaygroundTabs` into the respective `SideContentXXX` (and just delete the `PlaygroundTabs` file)? Since the `makeXXX` are being used in not just the playground (even if they are, the functions are technically workspace-agnostic).

RichDom2185 (commented):
Thanks for working on this! Definitely will help with efforts to bring in the CSE Machine to Stories :)

I haven't reviewed everything yet, but some preliminary comments so far as follows:

package.json
```diff
@@ -47,6 +47,7 @@
     "classnames": "^2.3.2",
     "flexboxgrid": "^6.3.1",
     "flexboxgrid-helpers": "^1.1.3",
+    "formik": "^2.4.5",
```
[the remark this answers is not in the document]
leeyi45: In my experience Formik provides excellent features for managing forms. If most of our forms are manually written right now I think there might be benefits to migrating

src/commons/sagas/AchievementSaga.ts
```diff
@@ -192,6 +192,7 @@ export default function* AchievementSaga(): SagaIterator {
       (state: OverallState) => state.session.enableAchievements
     );
     if (action.payload.find(e => e === EventType.ERROR)) {
+      // TODO update this to work with new side content system
```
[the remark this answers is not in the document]
leeyi45: It's not been done because I'm not sure what the intended behaviour is supposed to be, or how best to go about integrating it (like what should actually be responsible for the handling of this specific alert)

src/commons/sagas/CommonsSaga.ts
```diff
@@ -0,0 +1,17 @@
...
+import { SagaIterator } from 'redux-saga';
+import { call } from 'redux-saga/effects';
+
+import { changeModuleBackend } from '../application/actions/CommonsActions';
+import { CHANGE_MODULE_BACKEND } from '../application/types/CommonsTypes';
+import { safeTakeEvery } from './SafeEffects';
+
+export default function* CommonsSaga(): SagaIterator {
+  yield safeTakeEvery(
+    CHANGE_MODULE_BACKEND,
```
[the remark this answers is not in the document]
leeyi45: If we want the message to display when the page is opened, then no. The saga only triggers when the module backend actually changes

RichDom2185:
> * Using `@reduxjs/toolkit` to simplify our redux implementations: Unfortunately Redux Toolkit is very opinionated and mandates the use of Immer, which auto freezes objects in the store. This creates a myriad of issues, for example, since `js-slang` mutates the context object, which can't be done if it's frozen. To use this would require a major refactoring of the entire Redux state to ensure mutable objects and immutable objects are handled separately.

I'm ok with this honestly, better bite the bullet now, I suppose, before the tech debt gets worse. I've started some of the work with #2758 with more coming soon, trying to make the migration as painless as possible.

> The solution I have settled on is creating a separate slice for side content to attach to the OverallState. Then through the use of hooks and mapDispatchToProps, a workspace can indicate that a tab should display an alert, or reset an alert after a tab has been visited.

I think we can use both approaches in the interim while js-slang awaits a context refactor. Let's try to get this merged ASAP and the tests working so we don't keep getting merge conflicts.

src/commons/application/ApplicationReducer.ts
RichDom2185, line 1: Interesting that this seems to have reappeared, I'll just cherry pick the refactor commit into this branch, probably something happened in between all the merge conflict resolutions.

leeyi45: I'm actually of the opinion that this should be moved to `IOptions` in `js-slang`, but it will need to be left here in the frontend in soem form. When working with modules I do find myself having to switch between module servers occasionally, and it also means that I can't access modules while using my local dev server from another device (such as my phone to test mobile functionality)

leeyi45:
> > * Using `@reduxjs/toolkit` to simplify our redux implementations: Unfortunately Redux Toolkit is very opinionated and mandates the use of Immer, which auto freezes objects in the store. This creates a myriad of issues, for example, since `js-slang` mutates the context object, which can't be done if it's frozen. To use this would require a major refactoring of the entire Redux state to ensure mutable objects and immutable objects are handled separately.
> 
> I'm ok with this honestly, better bite the bullet now, I suppose, before the tech debt gets worse. I've started some of the work with #2758 with more coming soon, trying to make the migration as painless as possible.
> 
> > The solution I have settled on is creating a separate slice for side content to attach to the OverallState. Then through the use of hooks and mapDispatchToProps, a workspace can indicate that a tab should display an alert, or reset an alert after a tab has been visited.
> 
> I think we can use both approaches in the interim while js-slang awaits a context refactor. Let's try to get this merged ASAP and the tests working so we don't keep getting merge conflicts.

I can definitely help with the refactor if needed. I think my own RTK code is still residing as a stash somewhere.

src/pages/playground/PlaygroundTabs.tsx
leeyi45: I think it might be worth looking to as the rest of the workspaces get refactored, but for now I'm trying to minimize the number of things that I need to change.

leeyi45:
If all is good, we should probably merge this asap

RichDom2185:
> If all is good, we should probably merge this asap

Thanks for updating this PR and fixing the conflicts! I'll review and merge over the weekend.

RichDom2185 (approved):
LGTM, thanks a lot! I'll merge this in some ordering to minimise the possible conflicts that happen.

PR #2706, opened 2023-10-29
Update Search for SICP

yiwen101:
### Description

There are two major changes and one minor change:

Major change 1: The frontend is updated to work with the rewrites search data that: 
1.1 Improve the coverage and accuracy of search result. 
1.2 Fix known bugs of wrong url to search results
1.3 Support showing index over a range in index search.
1.4. Support url to all existing reachable reference (footnote, exercise, paragraph, begin/end...) for example. Used to be only for paragraph.

Major change 2: The UI&UX is improved, namely:
2.1. Now the search result will be displayed in a submenu style. The user could see the search result by simply hovering the cursor over the autocomplete entry
2.2 The text search autocomplete is enhanced from predicting the next words to predicting the next meaningful phrase. 
2.3 The result of the text search is also improved from displaying the whole paragraph to only the surrounding context.
2.4. The user could hide autocomplete panel by clicking else where, and unhide by clicking the input field again.
2.5 Support the display of latex in index search result.
2.6. Enhance the index search autocomplete to be insensitive be Capitalization of the first letter.
2.7. Add #begin and #end reference to SICP text, so that the page will not be unnecessarily refreshed.
2.8 No more no regular text ("/n", latex text for example) in the text search result as well.

A minor change is the improvement in code quality.

How ui looks like now:
Screenshot 2023-10-31 at 17 50 19
Screenshot 2023-10-31 at 17 50 35
e888a3696">
Screenshot 2023-10-31 at 17 50 28

Means of testing: could just clone from my repo, set up the frontend and go to the SICP page. No need to set up backend or modify the environment file.

### Type of change
-  Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality
-  Code quality improvements


src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -222,212 +111,541 @@ const SicpNavigationBar: React.FC = () => {
...
+      const searchData: SearchData = JSON.parse(xhr.responseText);
+      return searchData;
+    }
+  };
+
+  function search(keyStr: String, trie: TrieNode) {
+    const keys = [...keyStr];
+    let node = trie;
+    for (let i = 0; i < keys.length; i++) {
+      if (node === undefined || node.children === undefined) {
```
RichDom2185, line 96: Likewise over here, the way `TrieNode` is defined suggests that `node.children` will never be undefined, so can I double-check on what the proper structure of this object is supposed to be?

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -222,212 +111,541 @@ const SicpNavigationBar: React.FC = () => {
...
+          margin: '0',
+          width: '100%',
+          backgroundColor: focusedSearchResultIndex !== index ? 'white' : 'grey',
+          border: '1px solid black'
+        }}
+        key={index}
+        onClick={() => {
+          setIndexAutoCompleteCouldShow(false);
+          setSearchAutoCompleteCouldShow(false);
+          window.location.href = `http://localhost:8000/sicpjs/${result}`;
```
sayomaki, line 310: You might have missed out on this, it should be the current URL of the page (instead of `http://localhost:8000`, which means you don't actually need the protocol + host parts of the URL)

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -122,312 +62,536 @@ const SicpNavigationBar: React.FC = () => {
...
+        subStr = subStr + '...';
+      }
+      return subStr;
+    };
+
+    const isUserSearchSubMenuHidden = (searchInput: string) => {
+      return !(isSubmenuVisible === searchInput);
+    };
+
+    const getUserSearchResults = () => {
```
sayomaki, line 289: Is there the need for this additional function that basically just returns `results` (possible to use `results` directly instead?)

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -122,312 +62,536 @@ const SicpNavigationBar: React.FC = () => {
...
+            margin: '0',
+            width: '100%',
+            backgroundColor: focusedIndexSearchResultIndex !== index ? 'white' : 'grey',
+            border: '1px solid black'
+          }}
+          key={index}
+          onClick={() => {
+            setIndexAutoCompleteCouldShow(false);
+            setSearchAutoCompleteCouldShow(false);
+            window.location.href = `http://localhost:8000/sicpjs/${result.id}`;
```
sayomaki, line 380: Ditto here, should not be hardcoding the localhost URL location here

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
yiwen101: This is for the template method "menu" to works, which need to take in 4 functions

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
yiwen101: Sorry, forget to change this. Same as previous

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
yiwen101: It should not be undefined. I add this clause to be defensive

src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
```diff
@@ -222,212 +111,541 @@ const SicpNavigationBar: React.FC = () => {
...
-          url: SearchUrl + pure[i][0] + pure[i][1]
-        });
+      const searchData: SearchData = JSON.parse(xhr.responseText);
+      return searchData;
+    }
+  };
+
+  function search(keyStr: String, trie: TrieNode) {
+    const keys = [...keyStr];
+    let node = trie;
```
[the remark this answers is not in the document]
yiwen101: This should not be undefined. I add this line for defensive and easy debugging.

yiwen101:
> Hi, thanks for working on this!
> 
> I left some comments below; note that these are just just preliminary comments as I have not finished looking through all the changes. Was wondering, how can I test this locally? Do I need to set up the SICP backend as well? How do I do that?
> 
> Also, could you add screenshots of the new UI if possible?
> 
> Lastly, please take a look at some of the commits I've pushed (mostly cleaning) and let me know if you have any thoughts. Thanks!

Thank you for your prompt response! Your feedbacks are very valuable to me.
I will updated the screenshots and method of testing in PR description soon

sayomaki:
Thanks for the wonderful search interface update @yiwen101!

So far I have managed to set this up locally, and it seems to work as intended.

Screenshot 2023-10-31 at 5 33 34 PM
<br/><br/>

I do have a few quick questions (that may or may not be relevant):
* I see that there are a few helper functions for building the various search result menus (e.g. `buildUserSearchResultsMenuEntry`, `buildIndexSearchResultsMenuEntry`). Would it be possible to instead share them with a common function to reduce code duplication? (though it will be fine otherwise, if it would be too much work)
* Is the links added to `Sicp.tsx` (the `#begin` and `#end`) used as navigation buttons/links? I can't seem to find them anywhere

Besides this, for UI/UX-wise I have a few more observations:
* Previously, the keyword chosen would be highlighted in the search results, but it is no longer the case with your new search results, and this may cause slight trouble in spotting/identifying the word within the focused result
* It appears that the search results may still overflow below the page outside of the viewport, so it is possible to end up with results not visible without zooming out (likely has to do with the styling, could be lodged a separate issue to be resolved later on) 
* This change may negatively impact users on mobile screens to some extent, as the results may end up being not so user-friendly for these audiences (good to keep this in mind although not the main goal for this PR, we can explore how to improve the interface in the future)

yiwen101:
> Thanks for the wonderful search interface update @yiwen101!
> 
> So far I have managed to set this up locally, and it seems to work as intended.
> 
> Screenshot 2023-10-31 at 5 33 34 PM 
> 
> I do have a few quick questions (that may or may not be relevant):
> 
> * I see that there are a few helper functions for building the various search result menus (e.g. `buildUserSearchResultsMenuEntry`, `buildIndexSearchResultsMenuEntry`). Would it be possible to instead share them with a common function to reduce code duplication? (though it will be fine otherwise, if it would be too much work)
> * Is the links added to `Sicp.tsx` (the `#begin` and `#end`) used as navigation buttons/links? I can't seem to find them anywhere
> 
> Besides this, for UI/UX-wise I have a few more observations:
> 
> * Previously, the keyword chosen would be highlighted in the search results, but it is no longer the case with your new search results, and this may cause slight trouble in spotting/identifying the word within the focused result
> * It appears that the search results may still overflow below the page outside of the viewport, so it is possible to end up with results not visible without zooming out (likely has to do with the styling, could be lodged a separate issue to be resolved later on)
> * This change may negatively impact users on mobile screens to some extent, as the results may end up being not so user-friendly for these audiences (good to keep this in mind although not the main goal for this PR, we can explore how to improve the interface in the future)

Thank you for your prompt reply!
"I see that there are a few helper functions for building the various search result menus (e.g. `buildUserSearchResultsMenuEntry`, `buildIndexSearchResultsMenuEntry`). Would it be possible to instead share them with a common function to reduce code duplication? (though it will be fine otherwise, if it would be too much work)"
Emm, this is because I made changes to them very frequently and is more flexible to list them out separately and locate as local functions, easy to find and easy to edit.

" Is the links added to `Sicp.tsx` (the `#begin` and `#end`) used as navigation buttons/links? I can't seem to find them anywhere"
Yes, the change in made in the SICP rewrites search data, so that the user would not have to refresh the page to move to the start of the page.

"* Previously, the keyword chosen would be highlighted in the search results, but it is no longer the case with your new search results, and this may cause slight trouble in spotting/identifying the word within the focused result"
Because in the past, the search result display the whole paragraph, and autocomplete just predict the next word, so highlighting works well at help user to focus. But now the autocomplete is enhanced to predict the next meaningful sentence fragment, but search result is reduced to only the enclosing sentence, and I feel better to should the whole disclosing sentence, otherwise too much proportion of the sentence will be highlighted, and this might be distracting.

"> * It appears that the search results may still overflow below the page outside of the viewport, so it is possible to end up with results not visible without zooming out (likely has to do with the styling, could be lodged a separate issue to be resolved later on)
> * This change may negatively impact users on mobile screens to some extent, as the results may end up being not so user-friendly for these audiences (good to keep this in mind although not the main goal for this PR, we can explore how to improve the interface in the future)"
Thank you for pointing this out. Could I seek your further assistance on that?

sayomaki:
I've changed the navigation after clicking a search result to use React's built-in routing feature, so that there will not be unnecessary page refreshes caused by `window.location.href`. Other than that, it looks good to me.

@RichDom2185 do you mind taking a quick look and approve the changes? Thanks!

RichDom2185 (approved):
Note that there are the following issues (taken from my discussions with @sayomaki):

> I realise there's a global style clash with the black border
>
> and also clipping because of hardcoded submenu height of 4000% which overflows the screen when the focused menu item is near the bottom

But we've also come up with a much better UI refactor which will be split across multiple PRs, so we will merge this first and quickly rebase after.

RichDom2185:
Hi @yiwen101,

I hope you don't mind us making some changes to the UI. We feel that this is a really useful feature to have for revision for the upcoming exams so we wanted to get this merged quickly.

Please take a look at my linked follow-up PRs to see if there can still be further improvements.

Some comments below:

> Emm, this is because I made changes to them very frequently and is more flexible to list them out separately and locate as local functions, easy to find and easy to edit.

I understand what you are trying to do, and I agree with it. I think what @sayomaki was referring to, and I also noticed, was that a number of your helper functions relied on closures in order to do its job, which makes them less reusable. For example:

telegram-cloud-photo-size-5-6199491804232005827-y

`focusResult` is dependent on the closure of `searchInput` which comes from the parent scope, so every time `searchInput` changes you need to create a new `focusResult` function object. Likewise for `getIndex()`, even though some of this logic is actually reusable across text/user search and index search. If you rely on closures then you can't pull out the function as easily, and your functions can get really long as a result.

But if you use dependency injection, and pass `searchInput` as a parameter instead (here, it's called `query`):

telegram-cloud-photo-size-5-6199491804232005828-y

Now the function can be pulled out of the parent scope and memoized, and it's also more reusable, as you can simply pass in the dependency from the parent scope as an argument):

telegram-cloud-photo-size-5-6199491804232005829-y

And if you don't want to call the inner function immediately, you can always use currying, which still reduces code duplication.

RichDom2185:
> Now the function can be pulled out of the parent scope and memoized

Note that actually, if you take a look at my most recent PR, the following functions...

* `fetchSearchData`
* `search`
* `sentenceSearch`
* `autocomplete`
* `indexAutoComplete`
* `sentenceAutoComplete`
* `focusResult`
* `getIndex`

...can actually be entirely pulled out of the `SicpNavigationBar` component into a separate file, as everything is done via dependency injection.

I did not do that in the PR as I wanted to keep the diff understandable to facilitate the review process (also tagging @sayomaki in this comment for awareness during the PR review).

@yiwen101 , perhaps you could continue to move these functions into a separate "search utilities" file, or something similar. That way, `SicpNavigationBar` is really just focused on the UI rendering logic (Single Responsibility Principle in SWE), and the codebase also stays readable and organised.

yiwen101:
> I've changed the navigation after clicking a search result to use React's built-in routing feature, so that there will not be unnecessary page refreshes caused by `window.location.href`. Other than that, it looks good to me.
> 
> @RichDom2185 do you mind taking a quick look and approve the changes? Thanks!

Thank you! I do not know about the React's built-in routing feature. Preventing unnecessary page refreshes will significantly improve user experience.

PR #2707, opened 2023-10-31
Reduce code duplication, standardise styling in SICP search

RichDom2185:
### Description

Builds on #2706. Mainly an intermediate change (non-breaking), but simply done to:

* Reduce code duplication
* Standardise component interfaces

in order to lay the groundwork for eventual migration to a much better UI for searching using omnibox (as discussed with @sayomaki over Telegram).

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Code quality improvements

### How to test

Everything should still work with regards to SICP search.

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #2708, opened 2023-10-31
Migrate to Omnibox for SICP search

RichDom2185:
### Description

Builds on #2706 and #2707. Both needs to be merged first, and this will be rebased later.

* Much cleaner UI and UX
* Removes global style conflict
* Fixes clipping issue
* More clearly demarcates between section numbers and search results
* Fixes miscellaneous issues
* Improve code quality
* Improve messages for a clearer user experience

https://github.com/source-academy/frontend/assets/34370238/394c97fc-f98a-4489-b398-39a75453304f

(Above video was taken during development when only text search was migrated. This PR now migrates both text and index search in a similar way)

### Type of change

### How to test

Search results would still remain the same and functional.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
+++ b/src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
@@ -367,7 +367,10 @@ const SicpNavigationBar: React.FC = () => {
     <>
       <Omnibar
         isOpen={isOmnibarOpen}
-        inputProps={{ disabled: omnibarMode === 'submenu' }}
+        inputProps={{
+          disabled: omnibarMode === 'submenu',
+          placeholder: `${omnibarMode.charAt(0).toUpperCase()}${omnibarMode.slice(1)} Search...`
+        }}
         onClose={() => setIsOmnibarOpen(false)}
         items={searchResults}
         // Handled by individual items
--- a/src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
+++ b/src/commons/navigationBar/subcomponents/SicpNavigationBar.tsx
@@ -306,12 +306,14 @@ const SicpNavigationBar: React.FC = () => {
     setOmnibarMode('text');
     setIsOmnibarOpen(true);
     setQuery('');
+    setSearchResults([]);
   };
 
   const initIndexSearch = () => {
     setOmnibarMode('index');
     setIsOmnibarOpen(true);
     setQuery('');
+    setSearchResults([]);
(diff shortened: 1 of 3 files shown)


RichDom2185 (commented):
There are some type issues because we are not using a discriminated union to type the omnibox component (which is shared by two different kinds of results); they have been resolved by typecast for now. But I feel that these are very minor since the program flow logic should be enough to prevent mistakes when coding.

@sayomaki could you pitch in with your thoughts on this? (The types in question are `IndexSearchResult` and `string`, which are used for index search and text search respectively).

Alternatively we can come up with some shared interface but I don't think the effort is worth it for such a small thing. I think let us not do potentially premature optimisation on this.

Regardless, the feature still works, and that's most important at the moment given the tight timeline.

RichDom2185:
To the reviewers, may I suggest going through the changes using a diff editor, commit-by-commit. I've tried to reorder some chunks of code to make the diff between every commit easier to understand.

RichDom2185 (commented):
@sayomaki I've rebased the changes on top of #2707 accordingly, and this PR is now ready for review. Thanks!

sayomaki:
Some general comments from testing locally that may adversely affect the user experience:
- The omnibox does not automatically close itself upon clicking a search result (after the page navigates to the referenced location)
- The omnibox cannot be dismissed when clicking outside of the box (and the Esc key only works when the input is focused, which is not possible on the secondary menu where the input is disabled)

I have yet to finish taking a closer look at the code changes yet, but I will update when I get there and give my thoughts on how we should address the typings.

RichDom2185:
Thanks for the comments and catching these issues! I've fixed them, but I also found another regression: trying to initiate search a second time after an index search results in a runtime error. Will get to fixing it ASAP, thanks!

sayomaki:
After a closer review on the code, I found out that the memoization of the search data request is not working properly and creates unnecessary web traffic whenever the search input changes.

I think it seems possible to unify the typings by changing the `TrieNode` type and create an intersection type for the `value` property, and I think this will have minimal refactoring required for the remaining code.

I will go ahead and push my changes on top, and this PR looks good to be merged!

RichDom2185 (commented):
Nice, thanks for catching these! I can't approve this cause it's my own PR, but LGTM and I'll set it to auto-merge once you reapprove it :)

sayomaki (approved):
Alright, code works and looks good to be merged! Thanks for the work once again!

Issue #2665, opened 2023-09-04
AY2023/24 Sem 1 Avenger feedback on Grading Table


martin-henz:
This issue is meant to collect feedback on the revamped grading table. Please let us know what worked and what doesn't and make suggestions how to improve it.

lizchow:
### What worked for me:
- The UI looks much nicer than the previous table. The statuses are more visible. 
- The clickable filters are nice, makes it easier to filter missions without typing. 
   <img width="124" alt="image" src="https://github.com/source-academy/frontend/assets/51410656/02482fa9-2db5-4c8a-a459-2f0b5ce6b963">
- Dropdown button to filter by "my groups". Its nice to see only my students' assignments.
   <img width="167" alt="image" src="https://github.com/source-academy/frontend/assets/51410656/8003a6b0-0b60-45e5-b1ce-4cd22dab88b2">


### What didnt work for me:
- The grading count is weird. I'm not sure what the number is supposed to mean. 
   <img width="98" alt="image" src="https://github.com/source-academy/frontend/assets/51410656/2c813352-ee0f-49d8-a749-ee016561b57b">
- The grading summary is not giving an accurate summary. It's including assignments that have not been submitted. Ideally, avengers only want to know the number of assessments that have been submitted and not graded.    
   <img width="395" alt="image" src="https://github.com/source-academy/frontend/assets/51410656/7784dc36-3692-40d9-b39e-213903e9d3ca">
- Unsubmitting missions/quests now requires many more clicks.
  The new grading table has 2 views, `ungraded` and `all`, For a avenger to unsubmit a misison, the avenger need to click once on the `ungraded` dropdown and click one more time to select `all`.

   Previously, the grading table only had one view. The mission was immediately visible when I clicked on the grading tab and I could unsubmit with less friction.  
- Grading list is not updated after I finish grading an assignment. The grading status of the assignment remains as `None` even after grading. We need to refresh the page to update the statuses.
- Grading table does not fit fully in a 14inch laptop screen. The grading summary takes up a considerable amount of screen space, blocking the "grade assignment" button. If I scroll to the right to click on the grading buttons, then the assignment names are blocked.
   <img width="1012" alt="image" src="https://github.com/source-academy/frontend/assets/51410656/c773efc6-2c77-463d-9a6a-9515e3e53d23">

### Suggestions:

**DISCLAIMER:** I am not familiar with UI/UX design, so these are just my thoughts on how I would make my own life easier.

[...]

PR #2709, opened 2023-11-04
Persist current user's group in local storage

RichDom2185:
### Description

Somewhat part of #2665. The following bug was reported by @lizchow:

Accessing the grading dashboard will default to showing "my groups" for staff, but after refreshing the page, it defaults back to all groups.

The cause was that although the current user and their role is persisted in the browser, the same is not the case for the user's group. As our check involves showing all groups for admins *or* those without a group, the bug occurs upon the browser refresh.

Solution is simply to persist the group data together with the rest of current user information.

Sorry it took a while, thanks for reporting!

---

This PR also replaced deprecated interfaces with their replacements.

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/dialogs/ConfirmDialog.tsx
+++ b/src/commons/dialogs/ConfirmDialog.tsx
@@ -1,10 +1,10 @@
 import {
   Button,
+  ButtonProps,
   Classes,
   Dialog,
-  IButtonProps,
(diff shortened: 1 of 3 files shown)


sayomaki (approved):
Looks good, thanks!

PR #2712, opened 2023-11-13
Bump Node.js to version 20, update documentation

RichDom2185:
### Description

Node 16 is EOL, and while upgrading to Node 18 seems to break GitHub runners and other x86 machines, surprisingly, bumping to Node 20 fixes these issues, although now they break on Apple Silicon.

But this is expected, since the version of `gl` does not technically support Node 18. Hence, I've manually bumped the version using selective dependency resolution. This is safe because if we take a look at their changelog, we see that the only reason for the major version bump from v5 to v6 is that they remove support for Node 12, with no other breaking changes besides that.

I guess it's good we never bothered with Node 18 after all, since Node 18 just stopped being actively maintained.

TODO:

- [x] Update README instructions

### Type of change

### Checklist

- [x] I have tested this code


RichDom2185 (commented):
I've tested running `yarn start` and `yarn build` locally on 3 different platforms:

* Apple Silicon (M1 Pro MBP)
* Windows 10
* Windows Subsystem for Linux

So they should work with everything, but just to be sure:

* @leeyi45 do you mind helping me test this branch on your Windows machine as well?
* Likewise, @chownces on macOS?

.github/workflows/build-development.yml
```diff
@@ -9,76 +9,76 @@ jobs:
...
-      uses: peaceiris/actions-gh-pages@v3
-      with:
-        external_repository: source-academy/sourceacademy.org
-        deploy_key: ${{ secrets.DEPLOY_PRIVATE_KEY }}
-        publish_dir: ./build
-        publish_branch: master
-        force_orphan: true
+      - uses: actions/checkout@master
+      - name: Use Node.js
+        uses: actions/setup-node@v2-beta
```
sayomaki: Can we standardise and use `actions/setup-node@v3` instead?

sayomaki:
Tested and working with M1 MBP Apple Silicon (`yarn install`, `yarn build`, `yarn test` and `yarn start`), thanks for the good work!

System Specifications:
- macOS 12.7 Monterey
- node 20.9.0 arm64
- python 2.7.18 arm64
- Xcode 14.2 arm64

Presumably it should work fine on x86_64 based macOS as well, given that there are no issues with M1-based systems. @RichDom2185 do you mind sharing more system information for those you have tested on?

sayomaki:
Also, I notice that `gl@5.0.3` is still present within the lockfile and it does prompts a warning when running `yarn install`, and it seems that `gpu.js` has the dependency tied to this version of `gl`. Should we proceed to bump the subdepedency to `gl@6` instead?

.github/workflows/build-development.yml
RichDom2185: Oh, this commit was just for the bumping and my IDE auto-formatted the YAML. I've already updated it to v3 in a subsequent commit.

I forgot that v4 has already came out haha; I don't see any reason why we should stick to v3 so I'll bump it, although I do want to keep the `node-version: 20` in the YAML, just to lock it and also make it clear to developers what version we're using.

RichDom2185:
> Presumably it should work fine on x86_64 based macOS as well, given that there are no issues with M1-based systems. @RichDom2185 do you mind sharing more system information for those you have tested on?

* Apple Silicon (M1 Pro MBP)
  * macOS 14.1, 14.1.1 Sonoma
  * Node 20.9.0 arm64
  * Python 3.10.13 arm64
  * xcode-select version 2403
* Windows 10
  * Build 19044
  * Intel i7-7700HQ
  * Node 20.9.0
  * Python 3.11.6
* Windows Subsystem for Linux
  * Ubuntu 20.04 Focal, running on the above Win10 machine
  * Node 20.9.0
  * Python 3.8.10 (default for 20.04)
  * GCC 9.4.0

leeyi45:
As far as I can tell there's no issues running on Windows

.github/workflows/build-development.yml
```diff
@@ -9,76 +9,65 @@ jobs:
...
+          cache: yarn
+      - name: Setup Sentry CLI
+        run: |
+          curl -sL https://sentry.io/get-cli/ | INSTALL_DIR=. bash
+      - id: get-time
+        run: echo "time=$(date -Iseconds)" >> $GITHUB_OUTPUT
+      - name: yarn install and build
+        run: |
+          yarn install --frozen-lockfile
+          yarn run build
```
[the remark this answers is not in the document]
RichDom2185: Using

```yaml
with:
  cache: yarn
```

in `actions/setup-node` handles this caching for us. We no longer need to manually set up caching of certain files/folders anymore.

chownces (approved):

PR #2715, opened 2023-11-17
Fix regression to deployments without backend

sayomaki:
### Description

As of #2710, it seems that local testing deployments and deployments without backend are affected as it will show the maintenance page (as the backend is not present, and thus the API health check would fail).

The code in `RequestsSaga.tsx` remain unchanged and no checks are required, as it is assumed that a backend exists when calling any functions within.

### Type of change

### How to test

Testing locally with `USE_BACKEND` environment config set to FALSE should work normally and as expected.
Note that the error shown in the playground is related to #2695 instead.

### Checklist

- [x] I have tested this code


RichDom2185 (approved):
LGTM, thanks a lot!

PR #2719, opened 2023-12-08
Support SAML authentication provider

RichDom2185:
### Description

Adds support for SAML/SSO login flow (which is primarily handled in the backend as a SP).

Relevant backend PR: https://github.com/source-academy/backend/pull/990

### Type of change

### How to test

The login flow works expectedly.

### Checklist

- [x] I have tested this code


src/commons/utils/Constants.ts
```diff
@@ -77,6 +77,18 @@ for (let i = 1; ; ++i) {
...
+  const id = process.env[`REACT_APP_SAML_PROVIDER${i}`];
+  if (!id) {
+    break;
+  }
+
+  const name = process.env[`REACT_APP_SAML_PROVIDER${i}_NAME`] || 'Unnamed provider';
+  const endpoint = process.env[`REACT_APP_SAML_PROVIDER${i}_ENDPOINT`] || '';
+
+  authProviders.set(id, { name, endpoint, isDefault: false, type: AuthProviderType.SAML_SSO });
+}
```
chownces, line 90: maybe we can consider abstracting out this logic, since its repetitive

chownces (approved):

src/commons/utils/Constants.ts
RichDom2185: Hmmm, after considering it, the special case of `i === 1` being the default auth provider only for OAuth makes the abstraction unnecessarily slightly less trivial. Might be worth investigating if there is an actual need to have a "default" auth provider.

I think we can just leave it for now; we already have brotli compression enabled for our bundle in deployment so should not make much of a difference.

Issue #2722, opened 2024-01-24
Move out of Blueprint's CSS API


RichDom2185:
We use a lot of style overrides by directly referencing blueprint's classes in our own stylesheets. It is very error prone, and precisely because of this, Blueprint has also decided that it is time to [deprecate the CSS API](https://blueprintjs.com/docs/#core/components/menu.css-api).

We should move away from these hardcoded and inconsistent styles to reduce the codebase complexity.

PR #2724, opened 2024-01-25
Refactor/Cleanup Frontend (Part 2)

RichDom2185:
### Description

* Removes some unused/unnecessary styles
* Uses blueprint components where possible
* Migrates some styles to CSS modules

By using CSS modules (supported out of the box), not only do we prevent global style name conflicts/clashes, but it is also easier to see which styles are referenced by which parts of the code.

This PR simply aims to move as much as possible to use CSS modules. Though it includes some reorganisation of stylesheets, this is not the main focus. As CSS modules will have an explicit import, it would be much easier to subsequently reorganise/simplify the stylesheets into different files once everything is migrated.

Also did a minor change to SICP by removing unnecessary type annotations for the paramters.

### Type of change

### How to test

There should not be any UI regressions.

### Checklist

- [x] I have tested this code


src/commons/navigationBar/NavigationBar.tsx
```diff
@@ -304,7 +305,14 @@ const NavigationBar: React.FC = () => {
 
   return (
     <>
-      <Navbar className={classNames('NavigationBar', 'primary-navbar', Classes.DARK)}>
+      <Navbar
+        className={classNames(
+          'NavigationBar',
+          'primary-navbar',
```
sayomaki, line 311: It appears that you have defined `primary-navbar` twice here

src/commons/navigationBar/NavigationBar.tsx
```diff
@@ -385,12 +393,7 @@ export const createDesktopNavlink: CreateNavlinkFunction = navbarEntry => (
     title={navbarEntry.text}
   >
     <Icon icon={navbarEntry.icon} />
-    <div
-      className={classNames(
-        'navbar-button-text',
```
sayomaki, line 390: Is this `navbar-button-text` class no longer required?

sayomaki (commented):
Looks good to me, apart from a few minor changes needed.

src/commons/navigationBar/NavigationBar.tsx
RichDom2185: It's not twice, but it's due to the (partial) migration. The first `'primary-navbar'` refers to the global style, the `primary-navbar` class (which still exists). The second is our scoped style from our CSS module, where I introduced a new class called `primary-navbar`.

src/commons/navigationBar/NavigationBar.tsx
RichDom2185: No, previously `navbar-button-text` was needed to fix spacing issues caused by children of `NavigationBar__link`. Since we removed the latter class from the parent component, the spacing is automatically fixed (and in fact, broken if we keep this class).

src/commons/navigationBar/NavigationBar.tsx
sayomaki: Alright, got it!

sayomaki (approved):
Looks good to me, thanks for the changes!

src/commons/navigationBar/NavigationBar.tsx
RichDom2185: The global `primary-navbar` is defined in:

https://github.com/source-academy/frontend/blob/e050242b7ced98a5155189de2922a65d96045698/src/styles/_navigationBar.scss#L51-L60

The reason we can't bring it over to CSS modules is because of the nested classes `bp4-button` and `bp4-navbar-heading` that we apply custom styles too. If we port over this `primary-navbar` to CSS modules, the child classes also get transformed, and they don't target the Blueprint buttons and navbar heading components anymore.

This is another reason we shouldn't depend on the CSS API as mentioned in #2722.

For now, I think we just keep this, but ideally we'll move out from the entire CSS API + global styles in subsequent PRs.

PR #2756, opened 2024-02-02
CSE Machine: Fix display bug when scanning out declarations (frontend)

DiligentPenguinn:
### Description
This PR resolves issue #1503 . These are the changes implemented on the frontend side

### Changes
- [x] Display ':' or ':=' based on declaration type instead of value's writable property


martin-henz (approved):
Looks good.

PR #2762, opened 2024-02-05
Replace deprecated gapi.auth2 with GIS

sumomomomomo:
### Description

Replaces calls to the deprecated gapi.auth2 with calls to Google Identity Services. With gapi.auth2, only api keys generated before 2022 July are functional (https://developers.google.com/identity/sign-in/web/sign-in). This change will also allow for local testing of Google persistence with user generated api keys.

All other calls made to gapi.client were untouched - GIS is used only to facilitate getting the access token from Google. Existing Google picker UI remains the same.

Google login persistence is implemented by storing the access token in local storage.

Finally, login button is added to the Google Drive control bar for parity with Github control bar UI.

@types/google.accounts is required for this PR to add types to the GIS library

### Type of change

### How to test

Instructions from the wiki, "Obtaining the API keys from the Google API Console" is sufficient for setting up locally for testing.

### Checklist

- [ ] I have tested this code


sayomaki:
Hi @sumomomomomo, congratulations on creating your first PR in Source Academy and to start contributing your changes! Do note that even though you bypassed the code validation and checking when pushing to the repository sub-branch, you should still ensure that your code is well formatted and passes tests before opening the pull request. 

The points mentioned above can be done with the 2 commands as follows:
- Testing: `yarn test`. If updating snapshots are required, you may run `yarn test --updateSnapshot`
- Formatting: `yarn format`

I will take a look at your changes in a bit, and let you know some feedback for the changes that you have made. Keep up the good work and thanks for the contribution!

src/features/persistence/PersistenceUtils.tsx
RichDom2185, line 1: This file should have a `.ts` extension, not `.tsx`, as it does not have any JSX syntax.

package.json
```diff
@@ -38,6 +38,7 @@
     "@szhsin/react-menu": "^4.0.0",
     "@tanstack/react-table": "^8.9.3",
     "@tremor/react": "^1.8.2",
+    "@types/google.accounts": "^0.0.14",
```
RichDom2185, line 41: Two things here:

* Types should be a dev dependency, not a dependency
* I don't think this type is actually needed? You're not adding any new libraries?

src/commons/sagas/PersistenceSaga.tsx
```diff
@@ -307,35 +325,101 @@ const initialisationPromise: Promise<void> = new Promise(res => {
...
+    const scriptTag = document.createElement('script');
+    scriptTag.src = 'https://accounts.google.com/gsi/client';
+    scriptTag.async = true;
+    scriptTag.defer = true;
+    scriptTag.onload = () => resolve();
+    scriptTag.onerror = ev => {
+      reject(ev);
+    };
+    document.body.appendChild(scriptTag);
+  });
```
RichDom2185, line 342: We should follow React declarative-style coding conventions, instead of plain JS conventions. We can just use JSX to render the script tag directly.

src/commons/sagas/PersistenceSaga.tsx
```diff
@@ -307,35 +325,101 @@ const initialisationPromise: Promise<void> = new Promise(res => {
...
+        client_id: Constants.googleClientId!,
+        scope: SCOPES,
+        callback: () => void 0 // will be updated in getToken()
+      })
+    );
+  }).then(c => {
+    tokenClient = c;
+  });
+}
+
```
RichDom2185, line 368: If possible, we should use a library instead of writing our own handlers.

src/features/persistence/PersistenceUtils.tsx
```diff
@@ -0,0 +1,16 @@
+/**
+ * Calls Google useinfo API to get user's profile data, specifically email, using accessToken.
+ * If email field does not exist in the JSON response (invalid access token), will return undefined.
+ * Used with handleUserChanged to handle login.
+ * @param accessToken GIS access token
+ * @returns string if email field exists in JSON response, undefined if not
+ */
+export async function getUserProfileDataEmail(accessToken: string): Promise<string | undefined> {
```
RichDom2185, line 8: Ditto on using a library if possible.

src/pages/createStore.ts
```diff
@@ -50,7 +50,10 @@ function loadStore(loadedStore: SavedState | undefined) {
         octokit: loadedStore.session.githubAccessToken
           ? generateOctokitInstance(loadedStore.session.githubAccessToken)
           : undefined
-      }
+      },
+      googleUser: loadedStore.session.googleAccessToken
+        ? 'placeholder' // updates in PersistenceSaga
```
RichDom2185, line 55: This should not be a placeholder; either we set it to a value, or don't set it at all (i.e. use the defaults).

RichDom2185 (changes requested):
Hi, thanks for working on this; there seems to be a couple of weird changes that I'm not sure about. I don't think there's a need to add any dependency, at least at the moment.

Though I really feel you should try to see if Google provides an NPM library to make it easier to interact with the API; if possible, we should not need to worry about writing our own fetch calls, headers, etc. to the API endpoints.

Thanks again and let me know via Telegram if you encounter any issues!

PR #2775, opened 2024-02-12
Add Token Counter Displayer in Repl

jayjay19630:
### Description:
This pull request creates a new assessment config which enables a token counter. This token counter is a notification which lets the user know how many tokens their program has. This notification can be seen as an output in the Repl once the program is run.

### Tasks: 

- [x] Create new Repl output type called notification output.
- [x] Add new hasTokenCounter and customNotification properties to workspace.
- [x] Import Acorn dependency to count tokens in user's program.
- [x] Display token counter in Repl conditionally depending on assessment configurations.
- [x] Add new BooleanCell in admin panel to toggle token counter feature.
- [x] Fix testcases.

### Notes:
Note that modifying custom Repl notifications hasn't been implemented yet (possible future updates).

---------

Co-authored-by: Lau Xiang Yu


RichDom2185:
Corresponding backend PR: https://github.com/source-academy/backend/pull/1062

src/commons/workspace/WorkspaceReducer.ts
```diff
@@ -122,6 +126,15 @@ export const WorkspaceReducer: Reducer<WorkspaceManagerState> = (
   }
 
   switch (action.type) {
+    case ADD_TOKEN_COUNT:
```
RichDom2185, line 129: Would `SET_TOKEN_COUNT` make more sense instead of `ADD_TOKEN_COUNT` given the reducer's behaviour?

src/commons/workspace/WorkspaceReducer.ts
```diff
@@ -353,17 +382,35 @@ export const WorkspaceReducer: Reducer<WorkspaceManagerState> = (
...
-        } as ResultOutput);
+        } as ResultOutput;
+        const notificationOutputs: NotificationOutput[] = [];
+        if (state[workspaceLocation].hasTokenCounter) {
+          notificationOutputs.push({
+            consoleLog: `This program has ${tokens} tokens.`,
+            type: 'notification'
+          });
+        }
+        const customNotification = state[workspaceLocation].customNotification;
```
RichDom2185, line 404: Is this for future plans, or already used?

src/commons/sagas/WorkspaceSaga.ts
```diff
@@ -1256,6 +1257,14 @@ export function* evalCode(
 
   yield* dumpDisplayBuffer(workspaceLocation, isStoriesBlock, storyEnv);
 
+  // Change token count if its assessment and EVAL_EDITOR
+  if (actionType === EVAL_EDITOR && workspaceLocation === 'assessment') {
+    let tokenCounter = 0;
+    const tokens = [...tokenizer(entrypointCode, ACORN_PARSE_OPTIONS)];
+    tokenCounter = tokens.length;
```
RichDom2185, line 1264: Can we just

```ts
const tokenCount = tokens.length;
```

instead of declaring it at the top?

RichDom2185 (commented):
Thanks for working on this, looking forward to the new additions 👍 just some comments below:

src/commons/workspace/WorkspaceReducer.ts
jayjay19630: This is most likely for future plans. Not yet used.

src/commons/sagas/WorkspaceSaga.ts
jayjay19630: Yes, I'll make sure to edit that.

src/commons/workspace/WorkspaceReducer.ts
jayjay19630: That's true. I'll make sure to change all references to set token count instead

src/styles/_workspace.scss
```diff
@@ -766,6 +768,10 @@ $code-color-error: #ff4444;
     overflow-y: auto;
     margin: 0 0.5rem 0 0;
 
+    .#{$ns}-card:has(.notification-output) {
```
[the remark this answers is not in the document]
jayjay19630: Okay, understood

src/commons/workspace/WorkspaceReducer.ts
RichDom2185: Would it make more sense for `customNotification` to be `customNotifications` (a `string[]`) instead then? Since it's reasonable to assume you can have either 0, 1, or more than 1; there's no reason to only have either 0 or 1 notification present.

Anyway, it's not used now, so I'll just approve the PR, but perhaps you can create an issue: either remove it (cause it's unused), or make it support multiple notifications?

Issue #2701, opened 2023-10-24
CSE machine: animations


martin-henz:
See wiki: https://github.com/source-academy/frontend/wiki/Environment-Model-Visualiser#animations

Issue #2591, opened 2023-08-03
Env Viz: Animations


tsinyee:
Current env viz is quite static, would be great to add animations especially for the Agenda and stash
Refer to https://github.com/source-academy/frontend/wiki/Environment-Model-Visualiser#animations

martin-henz:
Lots of animations have been added already. Updated wiki entry under way.

PR #2776, opened 2024-02-12
CSE Machine: Add some control and stash animations

### LINKED ISSUE #2749: CSE Machine: Display array indexes on array objects
It would be nice if arrays created inside the CSE Machine would show their indexes as well, so a user can easily get an array item from their index instead of counting from the left.

A possible implementation would be to have each array cell to be labeled with a small number on the top left of the cell, which represents the index.

CZX123:
### Description

This PR adds animations to the CSE Machine, which only play when the user clicks the next button.

### Changes
- [X] Added display of array indexes. Fixes #2749
- [x] Added animations for specific CSE Machine actions:
  - Moving of literal from control to stash
  - Binary operation
  - Unary operation
  - Pop instruction
  - Addition of control items resulting from blocks, variable declarations and expression statements

Partially addresses #2701 #2591


src/features/envVisualizer/animationComponents/PopAnimation.tsx
```diff
@@ -0,0 +1,60 @@
...
+      {
+        ...stashItemPosition,
+        x: stashItemPosition.x + stashItemPosition.width / 2,
+        offsetX: stashItemPosition.width / 2,
+        y: stashItemPosition.y + stashItemPosition.height / 2,
+        offsetY: stashItemPosition.height / 2
+      },
+      { scaleX: 0.6, scaleY: 0.6 },
+      stashItem.text,
+      { delayMultiplier: 0.3 }
```
martin-henz, line 34: should we change this to 0.5, as discussed?

martin-henz (changes requested):

src/features/envVisualizer/animationComponents/PopAnimation.tsx
notnotmax: this is the pop animation, i believe we discussed increasing the duration of the assignment and lookup animations instead (which are not part of this PR).

RichDom2185 (commented):
Did a quick pass and overall direction looks good, loved the animations demo earlier today, I just have one minor nit below.

Also, as an extension: Perhaps we can multiply all these individual, animation-specific durations by a common shared constant exported from the utilities file. For now, it can be simply set to 1, but I can imagine in the future some people might want to slow down the animations by some factor, and we'd be able to relatively easily manipulate that if we already set up the multipliers beforehand.

martin-henz (approved):

PR #2787, opened 2024-02-17
Transfer pagination work from frontend to backend

josh1248:
Paired with backend PR: https://github.com/source-academy/backend/pull/1065.
### Description
This request shifts work of filtering from the frontend to the backend. Page state is converted into SQL offsets and limit for the backend, which stop querying once enough entries to fill the page is reached, hence saving time. State management by the grading submissions table state has been greatly reduced. Pagination work by tanstack has been removed. 

- Pagination work by tanstack has been removed, and now relies on the page and pageSize state supplied by the main page. The backend payload contains the total entries, which is converted to total pages. Offset in SQL querying is also converted within this page.
- Filtering for all categories have been changed to the backend. Adding or removing these filters / dropdown options resets the page index and requests a fresh set of entries to be displayed.
- Grading status dropdown option filter has been migrated to the backend. To simplify logic, the individual Grading status within the page is no longer filterable.

Further polish to explore to extend this feature change:
- Frontend-backend query conversion is currently hard-coded for GradingSubmissionTable to reverse the backend-frontend conversion by RequestsSaga. These conversion functions could be coupled.
- Smart caching of results from backend. For example, if the user visits page X, do LIMIT X * pageSize, store all results in frontend.
- Re-establish export grading CSV.

### Type of change
This is a breaking change:
- It expects a different backend response, {count: number, data: array of submissions} from backend PR number 1065, differing from current response structure, which is a simple array of submissions.
- It breaks the export grading CSV function, which can only export off the limited backend queries it gets.
- ~~It will diminish the functionality of the Grading Summary page as it can only read off the limited backend queries it gets.~~ (Edit: no longer an issue as the grading summary page will be deleted

Feature removals:
- Individual grading status can no longer be clicked on for filtering.
- Search bar feature has been restricted from global filter to assessment name to avoid OR or comparative querying.

Feature enhancement:
- additional dropdown option to allow the user to set the number of entries per page that they see.
- Introduces 2 additional page control buttons "<<" and ">>" that set the page to the minimum and maximum respectively.

image


public/externalLibs/sound/soundToneMatrix.js
josh1248, line 1: this must have been a merge that the fork did not update. this change is not intended.

RichDom2185: Actually, you committed it (according to the commit history). Most probably an automatic ESLint change. No worries, I'll revert it.

RichDom2185 (commented):
Hi, thanks for working on this! I've made some changes to the code for better typing, etc..

It seems like there are still failing tests. Could you fix them? Thanks!

josh1248:
I have updated the test, which incorrectly asserted an empty page state. i forgot to update this after including default values. all tests pass now except for the SicpNavigationBar.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -37,94 +39,216 @@ import GradingActions from './GradingActions';
...
+  // This is needed because a filter change is accompanied with a page reset.
+  const resetPage = useCallback(() => setPage(0), [setPage]);
+
+  const [showAllSubmissions, setShowAllSubmissions] = useState(false);
+  const showSubmissionOptions = [
+    { value: false, label: 'ungraded' },
+    { value: true, label: 'all' }
+  ];
+
+  // TODO: implement isAdmin functionality
```
RichDom2185, line 79: Why was this moved from `Grading` into `GradingSubmissionTable`? The logic for filtering by groups does not seem related to the table logic at all? Also, the TODO is still left here.

I'm in favour of bringing it back to the `Grading` component.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -107,24 +109,147 @@ const columns = [
...
+    return columnFilters
+      .concat(
+        showAllSubmissions
+          ? []
+          : [
+              {
+                id: 'gradingStatus',
+                value: GradingStatuses.none
+              }
+            ]
```
RichDom2185, line 189: Is this meant to replace

```ts
submissions.filter(
  s => showAllSubmissions || isSubmissionUngraded(s)
)
```

In the `Grading` component in the old version? If yes, note that `isSubmissionUngraded` is **not** check for grading status == none. Please refer to its implementation in `GradingUtils`.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -147,8 +273,41 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
     );
   }, [columnFilters, globalFilter, dispatch]);
 
+  // tells page to ask for new entries from main page when its state changes.
+  useEffect(() => {
+    updateEntries(limitGroup, pageParams, backendFilterParams);
+    // updateEntries seems to change between renders as a reference?
+    // eslint-disable-next-line
```
RichDom2185, line 280: DO NOT do this. There's a reason those warnings exist, when the `updateEntries` function is recreated (thus changes reference), you don't want to be calling the old reference, which may result in an error.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -147,8 +273,41 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
     );
   }, [columnFilters, globalFilter, dispatch]);
 
+  // tells page to ask for new entries from main page when its state changes.
+  useEffect(() => {
+    updateEntries(limitGroup, pageParams, backendFilterParams);
+    // updateEntries seems to change between renders as a reference?
```
RichDom2185, line 279: Yes, and as a result, it should be causing an infinite loop if you add it to the dependency array below.

And it should (cause the infinite loop) – this should point to something wrong in your code instead of disabling the warning.

In this case, your `updateEntries` comes from `updateGradingOverviewsCallback` in the `Grading` component, which is not memoized using `useCallback`. Therefore, when `updateEntries` is called and a new array for the grading overviews is made, it will trigger a rerender of `Grading` (as it should). But because `updateGradingOverviewsCallback` is not memoized, it also recreates this function, causing the reference change, causing the prop change, causing `useEffect` to fire again, repeating the whole cycle, resulting in infinite re-renders.

RichDom2185: Note that I didn't actually run this PR and just read through the code changes briefly, so please take this with a grain of salt and let me know if memoizing `updateGradingOverviewsCallback` doesn't solve the issue.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -147,8 +273,41 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
...
+          buttonProps={{ minimal: true, rightIcon: 'caret-down' }}
+        />
+        <Text>Submissions from</Text>
+        <SimpleDropdown
+          options={groupOptions}
+          selectedValue={limitGroup}
+          onClick={setStateAndReset(setLimitGroup)}
+          popoverProps={{ position: Position.BOTTOM }}
+          buttonProps={{ minimal: true, rightIcon: 'caret-down' }}
+        />
```
RichDom2185, line 301: Yea, I'm in favour of moving these back to `Grading` to keep the logic separate. After all, you're already using dependency injection to pass `updateEntries` as a prop – why can't the value of group: false/true be defined from `Grading`?

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -147,8 +273,41 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
...
+        <Text>Entries per page</Text>
+        <SimpleDropdown
+          options={pageSizeOptions}
+          selectedValue={pageSize}
+          onClick={setStateAndReset(setPageSize)}
+          popoverProps={{ position: Position.BOTTOM }}
+          buttonProps={{ minimal: true, rightIcon: 'caret-down' }}
+        />
```
RichDom2185, line 309: Would be good if you can add screenshots for changed/newly added UI elements like these.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -198,22 +357,36 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
...
+              onClick={() => resetPage()}
+              disabled={page <= 0}
+            />
             <Button
               size="xs"
               icon={() => <BpIcon icon={IconNames.ARROW_LEFT} />}
               variant="light"
-              onClick={() => table.previousPage()}
-              disabled={!table.getCanPreviousPage()}
+              onClick={() => setPage(page - 1)}
```
RichDom2185, line 371: You should ideally use an updater function for setting state to values that rely on the previous state. This minimises the possibility of bugs like multiple state updates getting uncaught:

```suggestion
              onClick={() => setPage(p => p - 1)}
```

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -198,22 +357,36 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
...
-              Page {table.getState().pagination.pageIndex + 1} of {table.getPageCount()}
+              Page {page + 1} of {Math.ceil(totalRows / pageSize)}
             </Bold>
             <Button
               size="xs"
               icon={() => <BpIcon icon={IconNames.ARROW_RIGHT} />}
               variant="light"
-              onClick={() => table.nextPage()}
-              disabled={!table.getCanNextPage()}
+              onClick={() => setPage(page + 1)}
```
RichDom2185, line 381: Ditto.

```suggestion
              onClick={() => setPage(p => p + 1)}
```

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -198,22 +357,36 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
...
+              disabled={page >= Math.ceil(totalRows / pageSize) - 1}
+            />
+            <Button
+              size="xs"
+              icon={() => <BpIcon icon={IconNames.DOUBLE_CHEVRON_RIGHT} />}
+              variant="light"
+              onClick={() => setPage(Math.ceil(totalRows / pageSize) - 1)}
+              disabled={page >= Math.ceil(totalRows / pageSize) - 1}
```
RichDom2185, line 389: There's quite a lot of duplication here. Can we assign the max page number to a `const` instead?

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -107,24 +109,147 @@ const columns = [
...
+                numGradedEqualToTotal: true
+              };
+            } else {
+              // case: excluded or grading. Not implemented yet.
+              return {};
+            }
+          default:
+            return column;
+        }
+      })
```
RichDom2185, line 229: This callback function seems like it can be moved outside the component, or perhaps even into `GradingUtils` to keep this table component small.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -198,22 +357,36 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({ submiss
 
         <Footer>
           <Flex justifyContent="justify-center" spaceX="space-x-3">
+            <Button
+              size="xs"
+              icon={() => <BpIcon icon={IconNames.DOUBLE_CHEVRON_LEFT} />}
+              variant="light"
+              onClick={() => resetPage()}
```
RichDom2185, line 364: I think semantically `setPage(0)` makes more sense here instead of `resetPage` even though they do the same thing.

RichDom2185 (changes requested):
Hi, thanks for working on this. My feedback below:

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
josh1248: Implemented! Makes sense that the function reference changes for each re-render.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
josh1248: ok, updated my PR description!

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
josh1248: implemented!

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
josh1248: implemented!

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
josh1248: implemented!

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
josh1248: the reason was to allow users to select their own entries per page, which is tied to the submissions table state.

for grading, it is to simplify the dispatch logic for backend grading parameters since it largely intersects with the filtering by grading status. similarly for grouping, it can largely intersect with filtering by group. noting your comment below, it may not be so easy after all. let me see if i can use your suggestion of prop passing to revert this.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
josh1248: i see... isSubmissionsUngraded() checks for statuses none or grading + submitted. this may need subquerying, so i think i will add mention this for discussion this week.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
RichDom2185: Entries per page is fine but in fact, it would be simpler to not move the logic to this table component.

> &hellip; it can largely intersect with filtering by group.

> &hellip; use your suggestion of prop passing

You've already done this; take a look at the type for `updateEntries`: 

```ts
{
  updateEntries: (
    group: boolean,
    pageParams: { offset: number; pageSize: number },
    filterParams: any
  ) => void;
}
```

You are calling it with a `group` parameter, but that's just `limitGroup` which comes from the dropdown. If move the dropdown back to `Grading`, and define `updateGradingOverviewsCallback` like this instead:

```ts
const updateGradingOverviewsCallback = useCallback(
  // No more `group` param
  (pageParams: { offset: number; pageSize: number }, filterParams: Object) => {
    // Replace the `group` argument with `limitGroup` directly
    dispatch(fetchGradingOverviews(limitGroup, pageParams, filterParams));
  },
  [dispatch]
);
```

It saves you the trouble from having to pass the `group` param back and forth, so the type of the `updateEntries` prop can be made simpler.

RichDom2185: Forgot to update the dependency array in my code suggestion above. Make sure to update to `[dispatch, limitGroup]`, that way, changing `limitGroup` recreates the function, which also triggers a re-render of the grading table thanks to `useEffect`, as it should.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
RichDom2185: I don't think it needs subquerying. `GradingStatus` is a purely frontend construct. See

https://github.com/source-academy/frontend/blob/de886953eb3fe317aa8e915970fd95fb481abac7/src/commons/sagas/RequestsSaga.ts#L1362-L1377

To search for "ungraded only", you only need to check that:

* `isManuallyGraded === false`
* `submissionStatus === 'submitted'`
* `numGraded < numQuestions`

Btw, @GabrielCWT, the last bullet point is what the commented TODO was doing before the previous team shifted to client-side filtering.

GabrielCWT: Yeah, an issue which we raised previously is that in order to find numGraded, we need to join answers with the submissions. But how can we guarentee that there is 10 submissions ungraded being sent over. What we have been doing so far is querying then limiting. If we apply the same logic to ungraded, we would join the entire answer table to submissions which is what caused long loading times.

GabrielCWT:
Hey @josh1248 maybe entries per page should be a drop down of 10 25 50 instead? If necessary a field for custom sizes.

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
GabrielCWT: Agreed, you could perhaps convert it into a helper function which takes in columnFilters and showAllSubmissions and remove the function from the component.

RichDom2185:
There are quite a bit of changes in this PR which has resulted in merge conflicts (for now), so best not to `git pull` until the comments are resolved – I'm quite confident that once the dropdown menus are moved back to `Grading`, the merge conflicts should mostly resolve themselves.

josh1248:
high prio polish: use debounce, an Enter key event listener, or a search button for searching by mission name. otherwise, a query is called everytime the string within the search bar changes.

GabrielCWT:
Hey @josh1248, one issue I have is that on typing in the search field, it sends a query everytime the input is changed. It would be better to have a submit button or onKeyPress 'Enter' to send a query to the backend

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
```diff
@@ -128,15 +128,22 @@ const GradingSubmissionTable: React.FC<GradingSubmissionTableProps> = ({
...
+  const [debouncedSearchValue, setDebouncedSearchValue] = useState('');
+
+  // timeout delay could be a placed within a config.
+  useEffect(() => {
+    const changeSearch = setTimeout(() => {
+      setDebouncedSearchValue(visibleSearchValue);
+    }, 300);
+
+    return () => clearTimeout(changeSearch);
+  }, [visibleSearchValue]);
```
GabrielCWT, line 141: This is a solution which I came up with using Blueprint's InputGroup, We could also implement a submit button alongside it to improve UI/UX experience 
<InputGroup
  style={{ borderRadius: '6px', width: 'clamp(200px, 20vw, 400px)' }}
  leftIcon={<BpIcon icon={IconNames.SEARCH} style={{ marginLeft: '0.75rem' }} />}
  placeholder="Search for assessment..."
  onKeyUp={e => {
    if (e.key === 'Enter') setSearchValue((e.target as HTMLInputElement).value);
  }}
/>

josh1248:
> Hey @josh1248, one issue I have is that on typing in the search field, it sends a query everytime the input is changed. It would be better to have a submit button or onKeyPress 'Enter' to send a query to the backend

Thanks for the reminder! I previously struggled with implementing this just a couple of days ago, and thought it was not trivial to implement. However, I just revisited the issue, and I have implemented a debounce that sends a query only after the text stops changing. The countless useEffect debugging experience over these few days have helped. =D

src/pages/academy/grading/subcomponents/GradingSubmissionsTable.tsx
RichDom2185, line 141: While this debounce implementation works, let's not do this manually. Lodash already provides a `debounce` helper: <https://lodash.com/docs/#debounce>

src/commons/sagas/RequestsSaga.ts
josh1248, line 1: Previous behaviour:
if incompatible submissions filtering and ungraded only dropdown options were chosen, the submissions filtering option would override the dropdown. For example, given Filter status == "Attempting" and ungraded only dropdown value, the ungraded dropdown value is ignored, even if the expectations are 0 entries (because ungraded and status Attempting are mutually exclusive)

Current behaviour: dropdown options overrule filtering by progress status instead.

BUG: if mutually incompatible status + dropdown is introduced, 0 entries are shown to the user, but we still fetch by "Attempted" only. Our dropdown options now do frontend filtering!!! This must have something to do with Filters being stored as part of the workspace session (which you can verify if you select filters, go to another page, then go back). This is part of a larger bug (link here later to some bug report and enhancement i will be filing). 

This bug does not occur for non-mutually exclusive intersections that play well together, like status==='Attempted' and viewing 'All': the backend still takes charge.


image

RichDom2185: Thanks for catching this! I think it's an edge case, and I agree the filters should have not been stored in the state anyway (this was an oversight that was missed in the original PR). Regardless, it works for a normal use case so I think we can just leave it as a separate issue – this PR is getting large anyway.

RichDom2185 (approved):
LGTM! Thanks for helping to review my changes @josh1248 !

Issue #2491, opened 2023-05-24
Language selector bypasses restrictions set on language chapters


ianyong:
This allows Folder mode to be accessed in Source 1.
![firefox_A52lqKGyth](https://github.com/source-academy/frontend/assets/5585517/7333b049-98e7-4873-a5a0-aebce1dcad7a)

Steps to replicate:
1. Switch to Source 2+ in the chapter selector.
1. Switch to Python/Scheme in the language selector.
1. Switch back to JavaScript in the language selector.

RichDom2185:
> Similar to #2465 except no page reloads are necessary to break the invariant that Source 1 cannot be used with Folder mode.

Hi, I don't think the two issues are similar. The root cause of that issue is more of it takes a while to persist the updated store after a state update. The root cause of this one is flawed business logic:

https://github.com/source-academy/frontend/blob/a231c3b883e85827bc09a3cc27d41f1b6a128786/src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx#L17-L24

This assumes that components that depend on the current language (e.g. Playground) would respect this and always only provide the appropriate features (e.g. Folder Mode) accordingly.

But in this case, there is no additional check in Playground to dispatch `toggleFolderMode` should `isFolderModeEnabled === true` and the language changes to something that does not support folder mode. This is because Playground is operating under the false assumption that `isFolderModeEnabled` will always be initially consistent with whether the current language supports it or not.

In fact, the issue has nothing to do with Source 1 specifically. Notice how in your screenshot, when you switched to Scheme or Python (which is also defined to not support multi-file), it exhibits the same behaviour as changing to Source 1.

I think the current solution for this is to do the same thing you did: disable the language dropdown when using folder mode (but that not might be the proper thing to do...)

ianyong:
> This is because Playground is operating under the false assumption that isFolderModeEnabled will always be initially consistent with whether the current language supports it or not.

This is a consequence of the design choice that Folder mode must always be toggled explicitly by the user (in the sense that we cannot not have this assumption if we want to maintain this design choice). This is because Prof @martin-henz and I felt that Folder mode being automatically toggled off when the user switches to Source 1 for example is unintuitive and confusing.

> I think the current solution for this is to do the same thing you did: disable the language dropdown when using folder mode (but that not might be the proper thing to do...)

[...]

PR #2794, opened 2024-02-19
Fixed navbar overflowing to next line

### LINKED ISSUE #2452: Right side navbar buttons wrap around improperly
nusportal

(image taken from (https://sourceacademy.nus.edu.sg/playground))

With a impractically small screen, as there is not enough space on a single bar, the buttons on the top right of the navbar will wrap below the rest of the navbar, landing on of the editor.

newbutton

(image taken from mobile (https://sourceacademy.org/playground))

With the addition of the new Language Switcher Button to the navbar, however, this problem can now be reproduced even with a relatively large mobile screen, making it more relevant now.

JovenSoh:
### Description

Closes #2452 

The right navigation bar would wrap around to the next line when on mobile devices. This would prevent users from properly selected the playground to code. Fixed it by changing the navbar component to be absolute so wrapping would not occur.

### Type of change

### How to test

1. Go to Playground on Mobile
2. Reduce the screen width till right navbar wraps over

### Checklist

- [x] I have tested this code


RichDom2185 (changes requested):
Hi, thanks for working on this.

Unfortunately, I don't think this is the right approach. The navbar should not be removed from the document flow (which is what position absolute does). The z-index is also very hacky.

Ideally, the right navbar shouldn't even have language selector (see #2491). It was a hacky and rushed implementation when multiple languages were first added. The whole design needs to be re-thought out.

Would you be keen to think of this?

martin-henz:
> Hi, thanks for working on this.
> 
> Unfortunately, I don't think this is the right approach. The navbar should not be removed from the document flow (which is what position absolute does). The z-index is also very hacky.
> 
> Ideally, the right navbar shouldn't even have language selector (see #2491). It was a hacky and rushed implementation when multiple languages were first added. The whole design needs to be re-thought out.
> 
> Would you be keen to think of this?

I think the language selection should remain on the top right corner. It allows the user to fix the programming language to be used throughout sourceacademy.org. (In the future, we might also have a natural language selection there, i.e. you can use Source Academy to study programming using Scheme in Hindi, etc.)

We all agree that the design of the language selection was done in a rush, which leads to all kinds of inconsistencies. We can't fix the design during this semester (till May 2024), so will need to live with the inconsistencies as far as the business logic is concerned.

I hope that this situation doesn't lead to a deadlock regarding this PR. As far as I can tell, the PR addresses a UI aspect of the language selection.

PR #2803, opened 2024-02-21
Finish renaming ec-evaluator to cse-machine

martin-henz:
### Description

There was still one reference to ec-evaluator in the frontend. This PR renames this reference to cse-machine, like the others that are renamed already.


RichDom2185 (approved):
LGTM

PR #2807, opened 2024-02-23
bumping js-slang

martin-henz:
### Description

bumping js-slang


RichDom2185 (commented):
Seems like according to the NPM release, the old files are still present?

PR #2714, opened 2023-11-14
Add ChatGPT dialogue box for SICP

Song-Mengfei:
### Description

Implemented an embedded ChatGPT dialogue box for SICP. Students have the ability to inquire with GPT about the content they are currently exploring within a passage.  

Need to add `openai` to dependency

Implementation:

1. As students navigate through a specific subsection(e.g. 1.2.3), the system records the current subsection and utilizes `useRef` to mark all divs containing text within that subsection.

2. When a student poses a question to the bot, `useRef` is employed to identify all paragraphs currently visible on the screen, and their text is extracted.

3. The system retrieves the summary of the current subsection from `SICPNotes` which contains short summary for subsections. Then the summary is incorporated into the prompt along with the text of the current paragraph, transmiting it to GPT as the "system," .

4. Student queries are sent to GPT, and the responses are received.

5. Code snippets in the replies are extracted and highlighted. A reminder is included at the end of the answer, specifying that the response is from GPT.

6. The responses are displayed on the screen for the students.

Concerns raised:

1. The GPT key is stored in the `.env.local` file of the project, and there's uncertainty about how to utilize it in the official version.

2. Creating `useRef` using a loop is not feasible. The manual creation and addition of `useRef` instances to an array are deemed cumbersome.

3. There's a lack of comprehension regarding Chapters 4 and 5 of SICP, leading to insufficient confidence in summarizing the content or evaluating the accuracy of GPT-generated summaries for these chapters.
### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

PR #2813, opened 2024-02-23
Louis: add chatbot to SICP textbook

yiwen101:
### Description
This is a follow up based on #2714
It depends on this change in the backend https://github.com/source-academy/backend/pull/1063

Louis is a chatbot assisting student understanding SICP.  It depends on the /chat endpoint in backend which is a wrapper of openAI client ({model: gpt4 ,  method: chat_completion}) . It is designed to serve the role of student assistant who is aware of his duty and knowledgable about the textbook content and student's query context.

Feature:

1.  Chat_completion with awareness of the chat history.

2. Good knowledge of the textbook content and the context of the queries (with help of summary notes and extracts from user's screen). 

3. Serve as the role of student assistant who prefer provide guidance and clarification over direct answer, and detect when questions are not relevant to SICP.

4. Only visible to authenticated user at frontend. (The backend end point will also only respond to authenticated users)

Future extensions: 

1. There's currently no generated summaries for chapter 4 and 5. So should student question these chapters, Louis would not know the textbook content and context as well.

2. There is currently limited logging and record of the conversation generated

3. There is currently no usage monitoring to prevent abuse of the endpoint. (Though the API is only exposed to authenticated user, and Louis will only be visible after logging in, these methods are soft).

4. There is potential of extending Louis to other areas: eg playground, providing hint to quest (to encourage student to actually attempt them) and even stories. The backend endpoint implementation can be reused. The the chatbot component can also be reused after refactoring (that takes prompt generation function as parameter).

7. OpenAI released "assistant" (Beta) feature recently. This allow user to could make interesting customisations, for example to sepcify in details the tools available to AI (what it can and can not do) or make reference to uploaded file. Could consider using this feature?

### Type of change


src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
```diff
@@ -0,0 +1,183 @@
...
+      for (let i = 0; i < contentHistory.length; i++) {
+        payload.push({ role: roleHistory[i], content: contentHistory[i] });
+      }
+      payload.push({ role: 'user', content: _userInput });
+      const tokens = { accessToken, refreshToken };
+      const response = await request(`chat`, 'POST', {
+        ...tokens,
+        body: { json: payload }
+      });
+      const text = await response?.text();
```
RichDom2185, line 88: Is it possible to move this to our Saga instead?

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { SourceTheme } from 'src/features/sicp/SourceTheme';
+
+import { request } from '../../../../commons/utils/RequestHelper';
+import SICPNotes from './SicpNotes';
+
+Buffer.from('anything', 'base64');
+
+if (!(window as any).Buffer) {
+  (window as any).Buffer = NodeBuffer;
+}
```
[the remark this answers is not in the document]
yiwen101: @Song-Mengfei could you help elaborate this? I am also confused as to what it does.

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
yiwen101: Will work on that tonight.

src/styles/_chatbot.scss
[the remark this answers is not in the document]
yiwen101: Will work on that tonight.

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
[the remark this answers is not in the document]
yiwen101: After checking with mengfei, I realize that this is for openai to work;
but since now we wrap openai client in the backend, we no longer need this.
Will get this deleted.

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
yiwen101: Could you provide some guidance on how this can be done?

Given how strongly coupled this is request is to the other local state, the furthest I can think of doing is like this: 
```
export async function chatComplete(tokens: Tokens, payload:[{role:string, content:string}]): Promise<string | undefined> {
      const response = await request(`chat`, 'POST', {
        ...tokens,
        body: { json: payload }
      });
      const text = await response?.text();
      return text;
}
```
But then I am confused by the backend saga(I need to pass in tokes). What are "yield" , "function*" and "ReturnType<typeof actions.xxx>" ? Do I need to transfer to that?

yiwen101: @RichDom2185 Hi, I have made the requested change in the commit "[Migrate request to Saga](https://github.com/source-academy/frontend/pull/2813/commits/6496254851bf2379b57f7dc3a67d4ee7a1dddbc2)". Is this what intended for? Thanks.

RichDom2185 (approved):
LGTM, thanks a lot!

PR #2789, opened 2024-02-18
Migrate to Redux Toolkit part 2

RichDom2185:
### Description

To rebase and force-push following the merge of #2758.

Also fixes a small type issue.

Please refer to commit history for list of changes, linked PR for motivation of changes.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Code quality improvements

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #2818, opened 2024-02-24
Temporary fix to resolve RTK freezing reducer states

sayomaki:
### Description

As of dcd2c69, the playground code evaluation seems to have been broken and unusable. This is caused by the fact that Redux Toolkit uses `immer` as their state manager. By design of `immer`, it will auto freeze states causing mutations to fail, even in production.

Thus, as we are still in the progress in migrating to RTK, and that various current sagas and actions still depend on these state mutations to work correctly (i.e. code evaluation), it will be good to explicitly disable auto-freezing of the state until the reducer/actions depending on state mutations get refactored.

### Type of change

### How to test

Ensure that playground code evaluates as normal, and that modules work as intended.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/pages/createStore.ts
+++ b/src/pages/createStore.ts
@@ -9,7 +9,7 @@ import MainSaga from '../commons/sagas/MainSaga';
 import { generateOctokitInstance } from '../commons/utils/GitHubPersistenceHelper';
 import { loadStoredState, SavedState, saveState } from './localStorage';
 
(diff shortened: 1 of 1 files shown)


RichDom2185:
Fixes the regression introduced in #2789.

martin-henz (approved):
Looks good!

RichDom2185 (approved):

RichDom2185:
Additional context: refer to https://immerjs.github.io/immer/freezing/

PR #2830, opened 2024-03-03
CSE Machine: Control stack display changes to accomodate two-step evaluation of block statements

NhatMinh0208:
### Type of change

### Checklist

- [X] I have tested this code
- [X] I have updated the documentation


martin-henz:
This approach looks hacky to me. We should add a simplification (clarification) step to the processing of nodes: If a node is a block without declaration, we turn it into a "sequence". (Yes, I'm proposing we call them "sequence" and not "rawBlock".) Sequences should be rendered as expected, by changing the implementation in js-slang/src/utils/astToString.ts.

NhatMinh0208:
Yes I agree, your approach makes a lot more sense. I will try to get it done this week.

PR #2832, opened 2024-03-04
Fix Playground Command Palette not closing

### LINKED ISSUE #2747: Mobile: Playground Palette does not close
The palette for the Playground on mobile devices fails to close when users navigate to a different page. It only closes upon users clicking on space outside the palette.

https://github.com/source-academy/frontend/assets/112536096/8aacbbff-544a-4207-b667-715894c778a7

Johnwz123:
### Description

Closes #2747

The 'overlayPage' function in the external Ace Editor library adds the 'div' containing the Command Palette in the document body so it is not removed when the `Editor` component is unmounted.

Override the `overlayPage` function to add an `id` to the `div` that is overlayed. This allows the `div` to be referenced and manually removed.

### Type of change

### How to test

1. In Playground, open the Command Palette (On Desktop, click `F1`. On Mobile, use the options panel)
2. Navigate back or forward page and command palette will now close automatically.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/editor/Editor.tsx
+++ b/src/commons/editor/Editor.tsx
@@ -2,6 +2,7 @@
 import { Ace, require as acequire, createEditSession } from 'ace-builds';
 import 'ace-builds/src-noconflict/ext-language_tools';
 import 'ace-builds/src-noconflict/ext-searchbox';
+import 'ace-builds/src-noconflict/ext-settings_menu';
 import 'js-slang/dist/editors/ace/theme/source';
 
 import * as AceBuilds from 'ace-builds';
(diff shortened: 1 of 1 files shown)


martin-henz (approved):
Looks good to me. As discussed in the meeting. Don't see a better way.

RichDom2185 (approved):
LGTM too, thanks a lot! I'll just have a minor comment where I think it would be good to explain why we do this in the code:

Between lines 564 and 565 together with the rest of the comments, might add a

```ts
// See https://github.com/source-academy/frontend/pull/2832
```

to link this PR from the code directly.

Thanks again!

PR #2835, opened 2024-03-06
Create contest configurations in ground control

jayjay19630:
### Description

Screenshot 2024-03-06 at 10 08 16 PM

Since contests are a unique case of assessments, we have decided to create its own tab within ground control. This will help ease access to such contests while also making preparations for future configurations and controls that will be specific to contests (i.e. voting assignment controls, exporting leaderboards, etc).

For assessments to be considered contests, course configurations will also include a new toggle button called 'Is Contest Related', which replaces 'Has Token Counter.' Note that the feature of toggling token counters will now be embedded in this new 'Is Contest Related' toggle, as the only use case for token counters are contests.

### Type of change

### Checklist
- [x] I have tested this code


DesSnowy:
Corresponding backend PR: https://github.com/source-academy/backend/pull/1085

src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
```diff
@@ -247,16 +247,16 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
...
   useEffect(() => {
-    if (props.assessmentConfiguration.hasTokenCounter) {
+    if (props.assessmentConfiguration.isContestRelated) {
       handleEnableTokenCounter();
     } else {
       handleDisableTokenCounter();
     }
   }, [
-    props.assessmentConfiguration.hasTokenCounter,
+    props.assessmentConfiguration.isContestRelated,
```
RichDom2185, line 259: Does this mean that any contest will always have a token counter enabled?

RichDom2185 (changes requested):
Hi, thanks for working on this. I have some questions I want to clarify:

> Note that the feature of toggling token counters will now be embedded in this new 'Is Contest Related' toggle, as the only use case for token counters are contests.

Are you saying that there is no more toggle for token counter (refer to the code snippet below)? I would think that if we were to encapsulate all these settings, then it would still remain as a separate toggle?

jayjay19630:
The main purpose of this PR is so that contest specific assessments can have extra configurations like a button for assigning contest entries to students (see below).

Screenshot 2024-03-07 at 4 38 12 PM

However, we've realised that by having only one 'isContestRelated' field, all contest assessmnets will have a switch for assigning entries, when we only want the contest (not voting) to have the switch. This makes it confusing as to how to actually assign the entries. We're working on a new field called "isContestVoting" to conditionally display the switch, so do expect some updates for that. In the meantime, I'll mark this PR as a draft.

PR #2844, opened 2024-03-16
Refactor `SideContentReducer`

RichDom2185:
### Description

Refactors `SideContentReducer` to use "RTK" and Immer, as well as splitting the logic of stories and non-stories side content into separate sub-reducers for better readability.

### Type of change

### Checklist

- [ ] I have tested this code

### DIFF
--- a/src/commons/sideContent/SideContentReducer.ts
+++ b/src/commons/sideContent/SideContentReducer.ts
@@ -1,10 +1,11 @@
+import { createReducer } from '@reduxjs/toolkit';
+import { Reducer } from 'redux';
+
 import { defaultSideContent, defaultSideContentManager } from '../application/ApplicationTypes';
 import { SourceActionType } from '../utils/ActionsHelper';
-import { getDynamicTabs, getTabId } from './SideContentHelper';
-import { getLocation } from './SideContentHelper';
-import { CHANGE_SIDE_CONTENT_HEIGHT } from './SideContentTypes';
+import { changeSideContentHeight, endAlertSideContent } from './SideContentActions';
+import { getDynamicTabs, getLocation, getTabId } from './SideContentHelper';
 import {
-  END_ALERT_SIDE_CONTENT,
   REMOVE_SIDE_CONTENT_ALERT,
   RESET_SIDE_CONTENT,
   SideContentManagerState,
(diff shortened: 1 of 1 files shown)


RichDom2185:
@leeyi45 could I ask for your thoughts/preliminary comments on this proof-of-concept refactoring? If it's all good, I'll proceed to refactor the rest of the reducer.

Thanks!

src/commons/sideContent/SideContentReducer.ts
```diff
@@ -1,72 +1,87 @@
...
+  const [workspaceLocation, storyEnv] = getLocation((action as any).payload.workspaceLocation);
+  if (workspaceLocation !== 'stories') {
+    return state;
+  }
+
+  const sideContentState = state.stories[storyEnv];
+  return createReducer(defaultSideContentManager, builder => {
+    builder
+      .addCase(changeSideContentHeight, (state, action) => {
+        state.stories[storyEnv].height = action.payload.height;
```
RichDom2185, line 42: Note that `sideContentState` above is outside the `createReducer` function; thus it is not a workable draft state from Immer and should not be written to, i.e. `sideContentState.XXX = YYY` is not allowed.

leeyi45:
Should we call `createReducer` every time? I think it's sufficient to perform the workspace location check, then just call the appropriate reducer

RichDom2185:
> Should we call `createReducer` every time? I think it's sufficient to perform the workspace location check, then just call the appropriate reducer

How can we capture `workspaceLocation` via closure, in that case?

leeyi45:
> > Should we call `createReducer` every time? I think it's sufficient to perform the workspace location check, then just call the appropriate reducer
> 
> How can we capture `workspaceLocation` via closure, in that case?

Do we need to capture `workspaceLocation`? Ideally the reducers should work for all workspaces right?

RichDom2185:
> > > Should we call `createReducer` every time? I think it's sufficient to perform the workspace location check, then just call the appropriate reducer
> > 
> > 
> > How can we capture `workspaceLocation` via closure, in that case?
> 
> Do we need to capture `workspaceLocation`? Ideally the reducers should work for all workspaces right?

That's true, I had originally thought that it might be not very ideal to have to write a

```ts
const workspaceLocation = // ...
const sideContentState = // ...
```

every time. But now that I think of it, the duplicated `getWorkspaceLocation` calls in `newWorkspaceLocation` make more sense.

But I still feel that the reducer should be split between stories and non-story workspace locations. What do you think of something like:

```ts
export const SideContentReducer = (state, action) => {
  state = storiesSideContentReducer(state, action);
  state = nonStoriesSideContentReducer(state, action);
  return state;
};

const storiesSideContentReducer = createReducer(
  defaultSideContent,
  builder => {
    builder.addCase(someAction, (state, action) => {
      if (action.payload.workspaceLocation === 'stories') {
        // ... logic here
      }
    });
  }
);

const nonStoriesSideContentReducer = createReducer(
  defaultSideContent,
  builder => {
    builder.addCase(someAction, (state, action) => {
      if (action.payload.workspaceLocation !== 'stories') {
        // ... logic here
      }
    });
  }
);

```

That way we group all the logic for non-story and stories together.

Though ideally, we remove the duplication entirely since the only differences would be how we access `sideContentState`.

leeyi45:
> > > > Should we call `createReducer` every time? I think it's sufficient to perform the workspace location check, then just call the appropriate reducer
> > > 
> > > 
> > > How can we capture `workspaceLocation` via closure, in that case?
> > 
> > 
> > Do we need to capture `workspaceLocation`? Ideally the reducers should work for all workspaces right?
> 
> That's true, I had originally thought that it might be not very ideal to have to write a
> 
> ```ts
> const workspaceLocation = // ...
> const sideContentState = // ...
> ```
> 
> every time. But now that I think of it, the duplicated `getWorkspaceLocation` calls in `newWorkspaceLocation` make more sense.
> 
> But I still feel that the reducer should be split between stories and non-story workspace locations. What do you think of something like:
> 
> ```ts
> export const SideContentReducer = (state, action) => {
>   state = storiesSideContentReducer(state, action);
>   state = nonStoriesSideContentReducer(state, action);
>   return state;
> };
> 
> const storiesSideContentReducer = createReducer(
>   defaultSideContent,
>   builder => {
>     builder.addCase(someAction, (state, action) => {
>       if (action.payload.workspaceLocation === 'stories') {
>         // ... logic here
>       }
>     });
>   }
> );
> 
> const nonStoriesSideContentReducer = createReducer(
>   defaultSideContent,
>   builder => {
>     builder.addCase(someAction, (state, action) => {
>       if (action.payload.workspaceLocation !== 'stories') {
>         // ... logic here
>       }
>     });
>   }
> );
> ```
> 
> That way we group all the logic for non-story and stories together.
> 
> Though ideally, we remove the duplication entirely since the only differences would be how we access `sideContentState`.

I guess if we want to maintain the difference between story and non-story side content we could do that

What I had in mind was more like:
```ts
export const SideContentReducer = (state, action) => {
  const [workspaceLocation, storyEnv] = getLocation(action.payload.workspaceLocation)
  if (workspaceLocation === 'stories') {
    const workspace = state.stories[storyEnv]
    return {
      ...state,
      stories: {
        ...state.stories,
        [storyEnv]: reducer(workspace)
      }
    }
  }

// and the same for non-story workspace locations  
}
```
The whole idea is to treat each workspace identically. But I guess that's less idiomatic?

PR #2853, opened 2024-03-17
CSE Machine: Rewrite the Config enum to be a constant object

### LINKED ISSUE #2851: CSE Machine: config should not be an enum
The CSE machine config should be a simple object, whose values can just be called like `Config.someValue` instead of having to use `Number()` or `.toString()`.

Moreover, enums are also not appropriate for the config as the enum values should be unique.

CZX123:
### Description
Fixes #2851 

### Changes
- [X] Rewrite both `Config` and `ControlStashConfig` enums to be constant objects.
- [X] Removed unneeded `toString()` and `Number`


RichDom2185 (approved):
LGTM, thanks!

PR #2855, opened 2024-03-19
Add a fullscreen button for the game div container

### LINKED ISSUE #2746: Mobile/Game: Full Screen Option
The game container is rather small on mobile devices. A "full screen" option could be added to improve accessibility for mobile users. The "full screen" option should force it into landscape mode.

image

Johnwz123:
### Description

Closes #2746 

Added a full-screen button and associated functionality for the game div container to improve accessibility on mobile devices. Fullscreen mode on mobile devices would also automatically force the screen orientation to landscape mode.

### Type of change

### How to test

1. Go to the Source Academy homepage where you can find the game. 
2. On the top left corner of the game container, you can find the "Full-screen" button. Clicking it should expand the game container to full-screen and force mobile screen orientation to landscape mode.
3. There should now be an "Exit Fullscreen" button on the top left corner. Clicking it should exit full-screen mode.

### Checklist

- [x] I have tested this code


src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +51,82 @@ function Game() {
     }
   }, [session, achievements, goals]);
 
+  // Logic for the fullscreen button to toggle fullscreen.
+  // Entering fullscreen will also lock the screen orientation to landscape.
+  const [isFullScreen, setIsFullScreen] = React.useState(false);
+  const toggleFullScreen = () => {
```
sayomaki, line 57: For this function, as we are using React, using the `ref` here would be more appropriate instead of accessing the element directly from the browser through `document`,

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +51,82 @@ function Game() {
...
+    const elem = document.getElementById('game-display') as HTMLElement;
+
+    if (!isFullScreen) {
+      if (elem.requestFullscreen) {
```
sayomaki, line 61: Here instead of checking if the function exists on the element, we should instead check whether the page is currently full screen or not (refer to [here (MDN)](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenElement) for more details)

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +51,82 @@ function Game() {
...
+        elem.requestFullscreen().then(() => {
+          if (window.screen.orientation) {
+            window.screen.orientation.lock('landscape');
+          }
+        });
+      }
+    } else {
+      if (document.exitFullscreen) {
```
sayomaki, line 69: Ditto to the previous comment.

sayomaki (dismissed):
I've taken a look at the code, and there are some changes I believe that should be made here. Other than that, everything else looks okay!

RichDom2185 (changes requested):
Hi, thanks for working on this! I am not particularly fond of this approach as it seems very imperative (unlike React's declarative-style coding).

I propose using the [`useFullscreen` hook](https://mantine.dev/hooks/use-fullscreen/) from the `@mantine/hooks` package. What are your thoughts?

Mentioning @sayomaki for comments as well.

martin-henz:
Would it be possible to generalize the full-screen option to be easily applicable to other components, such as the CSE machine, or the data visualizer, or achievements?

Johnwz123:
I have updated it so that the fullscreen functionality uses the [useFullscreen hook](https://mantine.dev/hooks/use-fullscreen/) from the @mantine/hooks package. 

Since it only handles toggling fullscreen, the functionality to lock screen orientation in landscape mode is added in a wrapper function.

sayomaki (commented):
Took a look at the updated code, and I only have one slight query below. Other than that, the changes looks good!

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +52,93 @@ function Game() {
...
+    if (window.screen.orientation) {
+      if (!isFullscreen) {
+        window.screen.orientation.lock('landscape');
+      } else {
+        window.screen.orientation.unlock();
+      }
+    }
+  };
+
+  const gameDisplayRef = React.useRef<HTMLDivElement | null>(null);
```
sayomaki, line 78: What is the purpose of this `gameDisplayRef` here, can we not just use the ref provided by the `useFullscreen` hook?

Johnwz123: I don't think so. The `gameDisplayRef` is a React RefObject that I used in line 99 to 113 to handle the resizing of the game display when entering/exiting fullscreen. But the ref provided by the `useFullscreen` hook is a callback ref function. See [https://mantine.dev/hooks/use-fullscreen/#definition](url).

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +52,93 @@ function Game() {
...
+
+  // Logic for the fullscreen button to dynamically adjust its size, position and padding
+  // based on the size of the game display.
+  const [iconSize, setIconSize] = React.useState(0);
+  const [iconLeft, setIconLeft] = React.useState('0px');
+  const [iconPadding, setIconPadding] = React.useState('0px');
+
+  React.useEffect(() => {
+    const handleResize = () => {
+      if (gameDisplayRef.current) {
```
RichDom2185, line 101: Are the values never reset otherwise?

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +52,93 @@ function Game() {
...
+        const aspectRatio = 16 / 9;
+        const height = gameDisplayRef.current.offsetHeight;
+        const width = gameDisplayRef.current.offsetWidth;
+        const size = height / 40;
+        const padding = height / 50;
+        const leftOffset =
+          isFullscreen || height * aspectRatio > width ? 0 : (width - height * aspectRatio) / 2;
+        setIconSize(size);
+        setIconPadding(`${padding}px`);
+        setIconLeft(`${leftOffset}px`);
```
RichDom2185, line 111: I don't get it:

1. Why are `iconSize`, `iconPadding`, `iconLeft` independent state variables? Aren't they all derived state?
2. What does the game height and width have anything to do with the icon positioning?
3. What is the aspect ratio for?

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +52,93 @@ function Game() {
...
+  );
+
+  // Logic for the fullscreen button to dynamically adjust its size, position and padding
+  // based on the size of the game display.
+  const [iconSize, setIconSize] = React.useState(0);
+  const [iconLeft, setIconLeft] = React.useState('0px');
+  const [iconPadding, setIconPadding] = React.useState('0px');
+
+  React.useEffect(() => {
+    const handleResize = () => {
```
RichDom2185, line 100: Please don't put this function definition inside the `useEffect` body, coupled with the resize observer below it results in so many unnecessary function objects being recreated, degrading performance.

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +52,93 @@ function Game() {
...
+        setIconPadding(`${padding}px`);
+        setIconLeft(`${leftOffset}px`);
+      }
+    };
+
+    // When exiting fullscreen, the browser might not have completed the transition
+    // at the time handleResize is called, so the height of gameDisplayRef.current
+    // is still the fullscreen height.
+    // To fix this, we delay handleResize by 100ms.
+    const delayedHandleResize = () => {
```
RichDom2185, line 119: Ditto on moving this out of the `useEffect` body.

src/pages/academy/game/Game.tsx
RichDom2185: My question is why do you need to handle the resizing manually? Doesn't `useFullscreen` already handle it for you?

RichDom2185 (changes requested):
Hi, thanks for working on this. I have quite a lot of questions with regards to the design as follows:

src/pages/academy/game/Game.tsx
Johnwz123: The fullscreen Icon element is a child of the game-display div element, which the Phaser game engine inserts the game canvas element in. The game canvas element has a fixed aspect ratio and its size is determined by the size of the game-display div element. When browser window width is too wide and height is too small, there will be empty space on the left and right of the game canvas element. So the left offset of the fullscreen Icon element is updated to ensure that it remains aligned with the top left of the game canvas element.

src/pages/academy/game/Game.tsx
```diff
@@ -50,9 +52,93 @@ function Game() {
...
+
   return (
     <>
-      <div id="game-display"></div>
+      <div id="game-display" ref={setGameDisplayRefs}>
+        <Icon
+          id="fullscreen-button"
+          icon={isFullscreen ? 'minimize' : 'fullscreen'}
+          color="white"
+          htmlTitle={isFullscreen ? 'Exit full screen' : 'Full screen'}
```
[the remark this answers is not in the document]
Johnwz123: Nope, no particular reason. I was just looking for a way to implement a simple tooltip and came across using `htmlTitle` first. Will update to use `Tooltip` instead.

src/pages/academy/game/Game.tsx
RichDom2185: I think the solution is very hacky. Why not just rely on the DOM to position the icon button correctly instead of using a resize observer? Instead of putting the button as a child you can just put it as a sibling, the wrap a parent container around both.

Maybe something like:

```tsx
{/* parent, relative, margin auto to squeeze child + align center */}
<div style={{ position: 'relative' }}>
  {/* absolute, top = left = padding */}
  <Icon /* ... */ />
  <div id="game-display"></div>
</div>
```

RichDom2185: Then you don't need to use any resize logic cause the DOM does the layouting for you

Johnwz123: I tried using `margin: auto` with the Icon button both as a child and as a sibling with a parent container wrapper. However, this causes issues with the resizing of the game canvas element since it references the game-display div element's width and height to determine its width and height. So, when you increase the size of the browser window, more margin gets added instead of the game canvas element getting enlarged.

RichDom2185: Ok, I understand it now; I still feel the solution is hacky. After looking at game code, I have an idea for a refactor. Let me discuss with @lhw-1 first.

RichDom2185 (approved):
I think we can merge this first and do the refactor later cause my idea would involve game refactoring which is out of scope for this PR.

Thanks and sorry for the wait, LGTM!

src/pages/academy/game/Game.tsx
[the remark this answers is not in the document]
RichDom2185: Any updates? I'll merge this PR first but please do change it in the future if you have the time (or create an issue if you don't).

PR #2859, opened 2024-03-20
Fix #2833

### LINKED ISSUE #2833: Unexpected behaviour about the interaction between the stepper tool and the frontend
Unexpected flashing of the stepper tool icon.
Expected behaviour: no flashing occurs (The CSE machine does not have this issue; it might be a useful source for referencing).

Order of execution to trigger the problem:
1. Run the program in stepper tool.
2. Change to Introduction setup.
3. Run the program in the setup.

Figures related to the issue:
IMG_7712
IMG_7713
IMG_7714
IMG_7715
(The figures are in the order in which the problem was triggered.)

For detailed behaviour, please check out the video clips attached.

https://github.com/source-academy/frontend/assets/97390773/f57a3ae2-5bc9-462e-9503-52d8e8afb51b

leeyi45:
Certain tabs (such as the substitution visualizer) could send their alert after dynamic tabs have already been spawned. This causes the alert to be displayed on the next evaluation, where `BEGIN_ALERT_SIDE_CONTENT` finally `take`s a `SPAWN_SIDE_CONTENT`.

This PR fixes this issue and also removes the `sideContent` object from workspaces.

Though this resolves #2833, work should probably be done to refactor and simplify how the frontend evaluates code.

### DIFF
--- a/src/commons/application/ApplicationTypes.ts
+++ b/src/commons/application/ApplicationTypes.ts
@@ -379,11 +379,7 @@ export const createDefaultWorkspace = (workspaceLocation: WorkspaceLocation): Wo
   isRunning: false,
   isDebugging: false,
   enableDebugging: true,
-  debuggerContext: {} as DebuggerContext,
-  sideContent: {
-    alerts: [],
-    dynamicTabs: []
-  }
+  debuggerContext: {} as DebuggerContext
 });
 
 const defaultFileName = 'program.js';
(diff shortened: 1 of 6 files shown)


leeyi45:
So upon further debugging, I think my final conclusion remains that how running code in the frontend needs to be refactored completely before this issue can be addressed.

PR #2865, opened 2024-03-23
Remove global variable from workspace saga

JoelChanZhiYang:
### Description
This change removes the global variable that is inside workspace saga and moves it to the global state. This is important if we want to split WorkspaceSaga into different files.

### Type of change

- [X ] Code quality improvements

### Checklist

- [x ] I have tested this code


src/commons/sagas/WorkspaceSaga.ts
```diff
@@ -595,10 +595,11 @@ export default function* WorkspaceSaga(): SagaIterator {
   );
 }
 
-let lastDebuggerResult: any;
-let lastNonDetResult: Result;
```
JoelChanZhiYang, line 599: These are the two global variables removed.

src/commons/sagas/WorkspaceSaga.ts
```diff
@@ -1099,6 +1100,9 @@ export function* evalCode(
   );
 
   const entrypointCode = files[entrypointFilePath];
+  const lastNonDetResult = yield select(
```
JoelChanZhiYang, line 1103: evalCode uses this global variable, which makes evalCode hard to refactor

RichDom2185 (approved):
LGTM, thanks!

Issue #2883, opened 2024-03-27
CSE Machine infinite loop when evaluating circular references


CZX123:
A circular reference inside the playground code leads to an infinite loop, crashing the CSE Machine. The CSE Machine does show the last step of the program, but the slider is disabled, and no other steps can be selected. An example can be seen [here](https://share.sourceacademy.nus.edu.sg/gggw3):
```js
const a = [0];
a[0] = a;
```

The console error appears to be caused by "immer" module, but the error is logged from [SafeEffects.ts](https://github.com/source-academy/frontend/blob/master/src/commons/sagas/SafeEffects.ts):
<img src="https://github.com/source-academy/frontend/assets/13569982/81a00238-9647-4aa6-899b-7f69d3dcf06e" width=400>

RichDom2185:
> The console error appears to be caused by "immer" module

Can you elaborate? I can't immediately see why Immer would be causing the error.

CZX123:
Now it's giving me a slightly different error:
<img src="https://github.com/source-academy/frontend/assets/13569982/8876aa7f-a792-41e3-8b81-cde3d3818d05" width=300>
It seems like Immer is still handling the context objects like environment trees when the CSE Machine gets run, which results in this error as there are definitely circular references inside.

RichDom2185:
> It seems like Immer is still handling the context objects like environment trees when the CSE Machine gets run, which results in this error as there are definitely circular references inside.

I see, the problem is with the design of the app: the js-slang context should have never been part of the frontend state in the first place since the golden rule of redux is to always ensure that state is immutable. The fact that js-slang context is mutable makes us resort to a lot of stuff like disabling auto-freeze and other workarounds.

Hmm, we really need to refactor code evaluation soon&hellip;.

@sayomaki @JoelChanZhiYang sorry to ping you out of nowhere but do you have any thoughts?

RichDom2185:
See also: https://immerjs.github.io/immer/pitfalls/

JoelChanZhiYang:
Just to clarify, is the circular reference inside the js-slang context intended behavior? 

Or in other words, is the root cause of the bug the fact that js-slang has a circular reference? Or is it because immer does not allow of circular references?

JoelChanZhiYang:
Ah I see. The context seems to be used all over the place. But it seems like it is mostly being consumed rather than modified.  It is modified in UPDATE_SUBLANGUAGE and I assume when the various eval functions are called. 

[...]

Issue #2716, opened 2023-11-19
CSE Machine: Overlapping drawing of pairs


deseansoh:
When running the following code and attempting to display the environmental model in the CSE machine, the pairs constructed overlap with one another. The arrows are pointing correctly though.

```
const hh1 = pair(undefined, undefined);
const hh2 = pair(undefined, undefined);
const hh3 = pair(undefined, undefined);

set_head(hh1, hh1);
set_tail(hh1, hh1);

set_head(hh2, pair(undefined, hh2));
set_head(head(hh2), head(hh2));
set_tail(hh2, hh2);

set_head(hh3, pair(hh3,hh3));
set_tail(hh3, head(hh3));
```

![Bug](https://github.com/source-academy/frontend/assets/62531591/4284a4af-9883-4be6-9d2f-728ea75187b2)

martin-henz:
I can confirm that the bug exists. Here is the share link:
https://share.sourceacademy.org/t5pi8

PR #2885, opened 2024-03-28
bumping js-slang

### LINKED ISSUE #2700: CSE machine: jumping function objects
Function objects jump too much. Example: https://share.sourceacademy.org/ljoin
Screenshot 2023-10-24 at 9 25 27 AM
Screenshot 2023-10-24 at 9 25 52 AM

### LINKED ISSUE #2705: CSE machine: display objects even when no longer accessible
We should have a way to display all objects that get created over time, i.e. all function objects and pairs.

That would be consistent with showing all frames that get created over time, even when they are no longer accessible.

### LINKED ISSUE #2781: CSE Machine: error when assigning "undefined" to a variable.
While running the below code:
```js
const x = undefined;
```
The CSE machine will get stuck at the step where `undefined` is being assigned to `x`. All subsequent steps will remain stuck on the same screen, but steps before this are still fine.
If you attempt to skip this assignment step using the slider, the steps after still will not work, but instead show whatever previous screen you were on.

Can be reproduced on https://sourceacademy.org/playground but works correctly on https://sourceacademy.nus.edu.sg/playground.

martin-henz:
See: https://github.com/source-academy/js-slang/pull/1600

Fixes #2700
Fixes #2705
Fixes #2781
This might solve #2716, but unable to test due to issue #2883.

### Changes
- Any object created in the stash now also appear inside the environment, next to the correct frame
- Revamp `removePreludeEnv()` behavior, it now completely removes any mention of the prelude environment and updates any reference to it correctly
- Revamp dummy binding behavior, it now depends on the objects in the heap of each environment
- Refactored some code to better match the new behavior
- Round values returned by `getTextWidth` so test cases do not have floating point irregularities

### Problems
- Any arrays or closures created from calling `stream_tail` may not have the correct environment, there is a workaround present but the results may be unexpected
   - This is due to `stream_tail` being a built-in function, a future js-slang version might change this

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -18,7 +18,7 @@ const cracoConfig = (module.exports = {
         plugin => plugin.constructor.name === 'InjectManifest'
       );
       if (injectManifestPlugin) {
-        injectManifestPlugin.config.maximumFileSizeToCacheInBytes = 15 * 1024 * 1024;
+        injectManifestPlugin.config.maximumFileSizeToCacheInBytes = 17 * 1024 * 1024;
       }
 
       // add rules to pack WASM (for Sourceror)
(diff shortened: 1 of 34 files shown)


RichDom2185 (approved):
I don't have the time to review this, but to not block the breaking changes for the development teams, let's merge it first since CI is passing.

PR #2889, opened 2024-03-29
Assign entries for voting

jayjay19630:
### Description

This change allows admin to control when they wish to assign / reassign entries for voting. Doing so will delete all existing vote submissions and essentially 'reboot' the assignment so that each entry has a fair chance of being voted on. 

### Type of change

### Checklist

- [x] I have tested this code


src/pages/academy/groundControl/subcomponents/GroundControlConfigureCell.tsx
```diff
@@ -117,12 +137,43 @@ const ConfigureCell: React.FC<Props> = ({ handleConfigureAssessment, data }) =>
...
+                        Cancel
+                      </Button>
+                    </ButtonGroup>
+                  </div>
+                )}
+                {isVotingPublished && (
+                  <p className="reassign-voting-warning">
+                    <b>All existing votes</b> will be <b>deleted</b> upon reassigning entries!
+                  </p>
+                )}
```
RichDom2185, line 176: Thanks for grouping the state logic together using comments. Perhaps it could be better if we can have a separate sub-subcomponent for this toggle instead for better abstraction? I can imagine once we add more configurations into this cell it can become more complex – the `GroundControlConfigureCell` just be focused on the layout, ideally.

src/styles/_groundcontrol.scss
```diff
@@ -85,6 +85,28 @@
...
+      margin-left: 15px;
+
+      .confirm-assign-text {
+        margin-top: 9px;
+      }
+
+      .voting-status-text {
+        margin-top: 10px;
+      }
+    }
```
RichDom2185, line 109: On the same vein as the above comment, these classes can be separated into a CSS module with the abstracted component name.

RichDom2185 (commented):
Just a minor comment on the refactor (abstraction). Otherwise, looks good, thanks!

Approved (in-principle), but as usual, we wait for the backend PR to pass review before merging both at the same time.

src/pages/academy/groundControl/subcomponents/GroundControlConfigureCell.tsx
jayjay19630: On that note, do you think it might be a good idea to abstract out the other switches as well (i.e. team assessment, is publish voting, etc.?) I know they're pretty small components but they still might clutter up the code once you add more and more stuff to the config panel

RichDom2185: I was thinking about it, and I came to the conclusion that we should just abstract as necessary (i.e. only this component, at least for now) – what do you think?

jayjay19630: Okay then, sounds good. I'll refactor just the assign entries portion inside a new file.

jayjay19630:
I've added a new folder called configureControls, which includes a new `AssignEntriesButton.tsx`. If any more components need to be added to the panel, they can be organised inside that configureControls folder.

Also, I've added a new scss module just for these configureControls. Let me know if there are any more issues.

RichDom2185 (approved):
LGTM, thanks a lot!

PR #2895, opened 2024-04-01
Java slang

Chang-CH:
### Type of change

### How to test

The java compiler has not yet been integrated. To run a class file enter code in the editor in the format (we should remove this when the compiler is integrated):
```
{
"<classname>.class": "<base64 encoded compiled classfile>
}
```

main class is should be named `Main.class`

Compile with `javac *.java -target 8 -source 8 `

The standard library is missing almost all native method implementations so only simple Java programs can run.

Also note that lambda functions are quite slow.
image

### Checklist

- [x] I have tested this code


src/commons/utils/JavaHelper.ts
```diff
@@ -0,0 +1,145 @@
+import setupJVM, { parseBin } from 'java-slang/dist/jvm';
+import { createModuleProxy, loadCachedFiles } from 'java-slang/dist/jvm/utils/integration';
+import { Context } from 'js-slang';
+import { initModuleContext, loadModuleBundle } from 'js-slang/dist/modules/loader/moduleLoader';
+
+import DisplayBufferService from './DisplayBufferService';
+
+const CDN = 'https://source-academy.github.io/modules/java/java-packages/src/';
```
RichDom2185, line 8: I think best to move this to the `Constants` object in `Constants.ts`?

src/commons/utils/JavaHelper.ts
```diff
@@ -0,0 +1,145 @@
...
+      initModuleContext(module, context, true);
+      const moduleFuncs = loadModuleBundle(module, context);
+      const { proxy } = createModuleProxy(module, moduleFuncs);
+      return Promise.resolve({ default: proxy });
+    }
+
+    return import(`java-slang/dist/jvm/stdlib/${path}.js`).then(m => {
+      return m;
+    });
+  };
```
RichDom2185, line 63: Is this the same as the following:

```suggestion
  const loadNatives = async (path: string) => {
    // dynamic load modules
    if (path.startsWith('modules')) {
      const module = path.split('/')[1] as string;
      initModuleContext(module, context, true);
      const moduleFuncs = loadModuleBundle(module, context);
      const { proxy } = createModuleProxy(module, moduleFuncs);
      return Promise.resolve({ default: proxy });
    }

    return await import(`java-slang/dist/jvm/stdlib/${path}.js`);
  };
```

src/commons/utils/JavaHelper.ts
```diff
@@ -0,0 +1,145 @@
...
+  const stdout = (str: string) => {
+    // java flush pushes a single newline to buffer
+    if (str === '\n') {
```
RichDom2185, line 66: What happens when you intentionally want to `System.out.println('');`?

Chang-CH: This is kind of a problem here. If we ignore this and always print whatever comes to stdout, we would get extra newlines. If we do it this way, `System.out.println('');` doesnt print a new line.

image
image

Not sure which is better.

src/commons/utils/JavaHelper.ts
Chang-CH: I am guessing the `Promise.resolve` would not be necessary as well?
```js
  const loadNatives = async (path: string) => {
    // dynamic load modules
    if (path.startsWith('modules')) {
      const module = path.split('/')[1] as string;
      initModuleContext(module, context, true);
      const moduleFuncs = loadModuleBundle(module, context);
      const { proxy } = createModuleProxy(module, moduleFuncs);
      return { default: proxy };
    }
    return await import(`java-slang/dist/jvm/stdlib/${path}.js`);
  };
  ```
I have committed this change

src/commons/utils/JavaHelper.ts
Chang-CH: updated.

src/commons/utils/JavaHelper.ts
RichDom2185: Hmmm, I tried looking at [Java documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/PrintStream.html#flush()) and StackOverflow, and the idea I'm getting is that there are two types of flushes:

* Implicit flush (done every time `\n` is seen)
* Explicit flush (done by calling `.flush()`)

And seems like `.flush()` does not print an additional newline?

Though the screenshots definitely show otherwise, so I'm not sure of the implementation detail; what exactly is causing this flush to happen for us?

Chang-CH: I am not sure how `println` is implemented, but as far as the standard library is concerned we only implement `FiledOutputStream::writeBytes`: [implementation](https://github.com/source-academy/java-slang/blob/main/src/jvm/stdlib/java/io/FileOutputStream.ts#L14)

`println` is not implemented by us, that the java standard library.

What I notice is when we `println("a")` we get `writeBytes("a")` and `writeBytes("\n")`. `DisplayBufferService` logs this as two separate lines rather than a single `a\n`

Chang-CH: image
Think I have a better solution, we just split newlines and push each line to `DisplayBufferService`.

RichDom2185 (approved):
LGTM, thanks!

PR #2904, opened 2024-04-07
Make frontend work with new module loading

leeyi45:
Make the frontend compatible with the PR that was just merged.

I did have to modify the code for `CToWasmHelper.ts`, not sure who should be responsible for reviewing that


martin-henz (approved):
Great to see the module loading all asynchronous now. Tested that module loading works in the frontend.

PR #2906, opened 2024-04-08
CSE Machine: Fixes for variadic function & global closures, added faded unreferenced objects

CZX123:
### Description
See: https://github.com/source-academy/js-slang/pull/1645
Fixes source-academy/js-slang#1641

### Changes
- Bump js-slang
- Differentiating between global functions and non-global ones is now more precise
  - Removes old behavior where removing `functionName` from a closure will result in a global function
- Closures in global frame now display correctly
- Closure are drawn further away from the environment frames
- Variadic functions now correctly show the argument array
- Functions created from calling `stream` now have a special type `StreamFn`
  - They are drawn like closures but only have one property `environment`, and they have their implementation hidden
- `apply_in_underlying_javascript` now works correctly with the environment
- Unreferenced objects in the environment are now drawn with a lighter shade
  
   - Achieved by making every object unreferenced by default, and marking them as referenced when the bindings or array units get added one at a time
- Objects like runes are now represented as a new type `SourceObject`, and should be displayed as a string, e.g. `<Rune>`, similar to how the REPL and the stepper displays it
  - Unable to test due to errors in module loading, but if this works, will solve #2908


martin-henz (approved):
All looks good. The method for finding unreferenced objects needs revisiting later; will make an issue.

PR #2925, opened 2024-04-10
Integrate Java type checker and compiler into frontend

1001mei:
### Type of change

### How to test

Input a simple Java program (such as the following) into the code editor and click on the `Run` button:
```
public class Main {
    public static void main(String[] args) {
        System.out.println("Hello world");
    }
}
```
Make sure to choose `Java` as the language option.

### Checklist

- [x] I have tested this code


martin-henz:
Great! Here is an example:

```
class HelloWorld {
    public static void main(String[] args) {
        System.out.println("Hello, World!"); 
    }
}
```

Screenshot 2024-04-10 at 6 51 28 PM

martin-henz (changes requested):

src/commons/utils/JavaHelper.ts
```diff
@@ -129,7 +135,11 @@ export async function javaRun(javaCode: string, context: Context) {
               resolve(
                 context.errors.length
                   ? { status: 'error' }
-                  : { status: 'finished', context, value: '' }
+                  : {
+                      status: 'finished',
+                      context,
+                      value: new ProgramTerminatedSuccessfully()
+                    }
```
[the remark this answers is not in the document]
bryanlohxz: Done!

src/commons/utils/JavaHelper.ts
```diff
@@ -9,6 +11,33 @@ import DisplayBufferService from './DisplayBufferService';
...
+    return Promise.resolve({ status: 'error' });
+  }
+
+  try {
+    const classFile = compileFromSource(javaCode);
+    compiled = {
+      'Main.class': Buffer.from(new BinaryWriter().generateBinary(classFile)).toString('base64')
+    };
+  } catch (e) {
+    stderr('TypeCheck', e);
```
martin-henz, line 37: Looks strange: Shouldn't this be runtime error instead of TypeCheck?

martin-henz (changes requested):
Found a small issue; otherwise looks good!

src/commons/utils/JavaHelper.ts
bryanlohxz: Will is be ok if I changed this to `"Compile"`? Since it catches errors during compiler time.

bryanlohxz: The `RuntimeErrors` are caught at the return of the `javaRun` method.

bryanlohxz: Resolved.

image

martin-henz (approved):

PR #2928, opened 2024-04-12
Add stories moderation

keyansheng:
### Description

source-academy/stories-backend#120

### Type of change

### Checklist

- [ ] I have tested this code


src/commons/sagas/StoriesSaga.ts
```diff
@@ -33,11 +35,34 @@ import { safeTakeEvery as takeEvery } from './SafeEffects';
...
+      const resp = await getStories(tokens, StoryStatus.Published);
       return resp ?? [];
     });
 
+    const allStories: StoryListViews = {
+      draft: draftStories,
+      pending: pendingStories,
+      rejected: rejectedStories,
+      published: publishedStories
+    };
```
RichDom2185, line 64: This design doesn't look right to me. To the user/the frontend, the "listing" of stories should be transparent.

In other words, the FE just makes one "list stories" query, and the backend returns a list of stories that the user has access to. The permissions/different types of stories should be transparent and the frontend can subsequently later filter the story by the status. You are adding a status field inside the story model in the backend but it's not being utilised here at all.

RichDom2185 (changes requested):
Hi, thanks for working on this, my comments are as follows (mainly about the feature design):

PR #2940, opened 2024-04-15
Bump almost every dependency

RichDom2185:
### Description

Recreated the lockfile and bumps a lot of things.

Need to manually check and run through features for regressions.

### Type of change

### Checklist

- [ ] I have tested this code


RichDom2185:
Still one remaining runtime error with the following cryptic message:

image

@chownces have you encountered this before?

chownces:
> Still one remaining runtime error with the following cryptic message:
> 
> image
> @chownces have you encountered this before?

Nope, not yet. How can I repro this?

RichDom2185:
> Nope, not yet. How can I repro this?

Just need to run `yarn start` on playground-only mode.

chownces:
I think probably one of the resize libraries is causing this issue. Will need to investigate further. This only happens during browser force refreshes on my end.

RichDom2185:
> I think probably one of the resize libraries is causing this issue. Will need to investigate further. This only happens during browser force refreshes on my end.

I suspect might be the router, but also, seems like ace is bugging out as well (despite being **not** updated):

image

PR #2943, opened 2024-04-16
Fix double window prompt when uploading users

### LINKED ISSUE #2177: Add user prompts two file select window
Behaviour as described in the title. 

Reason: not investigated.

cheehongw:
### Description

Fix #2177 

Previously, two window prompts were shown when uploading a csv file. The first window was caused by the rootProps from the CSVReader component. The second window was caused by the FileInput's input element.

Can't see any neat way to keep using `<FileInput>` without its `<input>` behavior, so I replaced it entirely with something that looks roughly the same

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/pages/academy/adminPanel/subcomponents/AddStoriesUserPanel.tsx
+++ b/src/pages/academy/adminPanel/subcomponents/AddStoriesUserPanel.tsx
@@ -1,7 +1,7 @@
 import {
   Button,
   Callout,
-  FileInput,
+  Classes,
   FormGroup,
   H2,
   H4,
@@ -177,7 +177,10 @@ const AddStoriesUserPanel: React.FC<Props> = props => {
               >
                 {({ getRootProps, acceptedFile, ProgressBar, getRemoveFileProps }: any) => (
                   <>
-                    <FileInput text="Upload CSV" inputProps={getRootProps()} />
+                    <label className={Classes.FILE_INPUT} {...getRootProps()}>
+                      <div style={{minWidth: 250}}> </div>
+                      <span className={Classes.FILE_UPLOAD_INPUT}>Upload CSV</span>
+                    </label>
                     <Popover
                       content={
                         <div>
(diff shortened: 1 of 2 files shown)


RichDom2185 (commented):
Hi, thanks for working on this and also providing a UI screenshot of the changes!

I'd like to bring to your attention #2722: while it does solve the problem, I'm not sure if it's the best way given they are deprecated. Having said that, since we are using the `Classes` object as opposed to raw `bp5-`-prefixed strings directly, any breaking changes should be caught as type errors.

What do you think? Looking at https://blueprintjs.com/docs/#core/components/file-input, it does seem that the cause of this is not Blueprint, but simply `<input type="file" />` which is beyond our control&hellip;

cheehongw:
@RichDom2185  Thanks for the prompt review and bringing my attention to #2722 

The original intention was to fix the bug while somehow preserving the styling used in `FileInput` as much as possible. 

However, I believe using Blueprint's react component API (aka `<FileInput>`) is out of the question here since we cannot get rid of the `<input type="file" />` that is baked into the `FileInput` component. Unfortunately, the solution for moving out of Blueprint's CSS API is to start using their React components, which is the work being done in #2722.

Therefore, I see two possibilities here:

1. keep styling consistent with Blueprint and use the CSS API 
2. don't follow blueprint's styling and write the styling ourselves in `_academy.scss`

The later is a big decision to make, so I think using Blueprint's `Classes` object is likely the best option.

RichDom2185 (approved):
Thanks for the reply!

It's good that this rationale is communicated somewhere (i.e. this PR) so that future developers will be able to reference this. I agree that using `Classes` seems like a good middle ground. The CSS namespace is abstracted away, and though it can cause breaking changes when removed, the breaking changes will be caught with a type error, so we don't have to worry about it going unnoticed.

LGTM, thanks and congrats on your first contribution to Source Academy!

PR #2945, opened 2024-04-16
Java upload tab

Chang-CH:
### Description

Allows users to bypass compiler + type checker **in local development** by uploading class files to run directly.

Supports multiple file uploads.

image

### Type of change

### How to test

Compile java file(s) with `javac *.java -target 8 -source 8` and upload

### Checklist

- [x] I have tested this code


RichDom2185 (commented):
Will this be safe (considering one can upload any file)? Any potential social engineering attack vector? Is it possible to access the JS environment from Java (and e.g. steal browser cookies)?

Chang-CH:
> Will this be safe (considering one can upload any file)? Any potential social engineering attack vector? Is it possible to access the JS environment from Java (and e.g. steal browser cookies)?

At the moment most of the standard library is not yet implemented, there are no native functions that access network.

Ideally we only need the run tab, but due to the limitations of the type checker and compiler this was requested as a feature.

martin-henz:
Yes, maybe this feature should only be available in the command line tool and not in sourceacademy.org? I agree that for now it's ok. Maybe we include it for the project presentation next week and then remove it?

Chang-CH:
> Yes, maybe this feature should only be available in the command line tool and not in sourceacademy.org? I agree that for now it's ok. Maybe we include it for the project presentation next week and then remove it?

I am planning to use localhost for presentation. I think mei tin is using the localhost approach as well. Is there a way we only enable it for dev and not production? do we have feature flags?

Chang-CH:
> > Is there a way we only enable it for dev and not production?
> 
> Not sure, does `process.env.NODE_ENV` work?

looks like it does, added the check.

src/commons/sideContent/content/SideContentUpload.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <br />
+          Compile the files with the following command:
+          <code>javac *.java -target 8 -source 8</code>
+          <br />
+          <br />
+          Avoid running class files downloaded from unknown sources.
+          <br />
+          <br />
+          <strong>Main class must be named Main and uploaded as Main.class.</strong>
+        </p>
```
RichDom2185, line 46: Semantically, does it make sense to have everything in a single paragraph with so many line breaks, instead of multiple `p`/`pre` tags?

src/commons/sideContent/content/SideContentUpload.tsx
```diff
@@ -0,0 +1,83 @@
...
+          }}
+        />
+      </div>
+    );
+  }
+}
+
+const makeUploadTabFrom = (
+  location: SideContentLocation,
+  onUpload: (files: { [key: string]: any }) => void
```
RichDom2185, line 75: Or perhaps this type could be aliased instead (or both).

src/commons/utils/JavaHelper.ts
```diff
@@ -106,7 +91,34 @@ export async function javaRun(
     }
   };
 
-  if (isUsingCse) return await runJavaCseMachine(javaCode, targetStep, context);
+  // #endregion
```
RichDom2185, line 94: Can we remove this comment?

src/commons/workspace/WorkspaceReducer.ts
```diff
@@ -841,6 +843,18 @@ const newWorkspaceReducer = createReducer(defaultWorkspaceManager, builder => {
...
       };
+    })
+    .addCase(toggleUsingUpload, (state, action) => {
+      const { workspaceLocation } = action.payload;
+      if (workspaceLocation === 'playground' || workspaceLocation === 'sicp') {
+        state[workspaceLocation].usingUpload = action.payload.usingUpload;
+      }
+    })
+    .addCase(uploadFiles, (state, action) => {
+      const { workspaceLocation } = action.payload;
```
RichDom2185, line 854: `workspaceLocation` should not be accessed via destructuring of `action.payload`. Refer to the other functions for how (i.e. `getWorkspaceLocation`).

src/commons/sideContent/content/SideContentUpload.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+type Props = {
+  workspaceLocation: SideContentLocation;
+  onUpload: (files: UploadResult) => void;
+};
+
+/**
+ * This component is responsible for uploading Java class files to bypass the compiler.
+ */
+const SideContentUpload: React.FC<Props> = ({ onUpload, workspaceLocation }) => {
```
RichDom2185, line 31: The `workspaceLocation` prop is not used. Is this intentional?

src/commons/sideContent/content/SideContentUpload.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <code>javac *.java -target 8 -source 8</code>
+      </pre>
+      <p>Avoid running class files downloaded from unknown sources.</p>
+      <p>
+        <strong>Main class must be named Main and uploaded as Main.class.</strong>
+      </p>
+      <FileInput
+        inputProps={{ multiple: true, accept: '.class' }}
+        onInputChange={handleFileUpload}
+      />
```
RichDom2185, line 70: Can we keep track of the state by updating the `text` prop? See https://blueprintjs.com/docs/#core/components/file-input

RichDom2185 (changes requested):
Hi, I've fixed some of the bugs and typings as per my above comments but I still have some comments below:

src/commons/sideContent/content/SideContentUpload.tsx
Chang-CH: Not quite sure what you mean by this. Do we want to show some kind of text to indicate what files are uploaded like "4 files uploaded"?

RichDom2185 (approved):
LGTM, thanks and sorry for the delayed review!

Issue #2468, opened 2023-05-08
Internationalise Source Academy


RichDom2185:
Let us create a roadmap/issue tracker to track the progress of supporting multiple languages in Source Academy:

* [ ] Settle on an internalisation framework (probably [`react-i18next`](https://react.i18next.com/))
* [ ] Settle implementation design/typing/etc.
* [ ] Migrate basic strings (UI button labels, etc.) into the framework
* [ ] Migrate remaining strings (textual descriptions, etc.) into the framework
* [ ] Add more languages

Issue #2426, opened 2023-04-06
i18n: Internationalization


martin-henz:
We should (finally) look at internationalization. If we set up the infrastructure, we can gradually externalize our strings and text snippets.

RichDom2185:
One proposed design was #2496.

Issue #1943, opened 2021-08-23
Help text of sourceacademy.org


martin-henz:
Currently we have:

<img width="627" alt="Screenshot 2021-08-23 at 4 35 13 PM" src="https://user-images.githubusercontent.com/30220768/130417539-6b120e45-823d-48d8-a046-3e93b446b0d1.png">

We need to make this more generic:

As of August 2021, we recommend the browsers Google Chrome or Microsoft Edge, Version 83 or higher, or Mozilla Firefox, Version 78 or higher, to visit the Source Academy. If you encounter issues with the Source Academy using these browsers, please use the issue system of the Source Academy repositories on [GitHub](https://github.com/source-academy)

martin-henz:
Of course the help text of the multi-tenant version is configurable  on a per-course basis. A simple solution for now would be to make the "server-less" frontend take its help ext from the .env file. Any better idea, anyone?

martin-henz:
The latter option should be done together with many other strings shown in the academy. i18n will take care of this.

PR #2946, opened 2024-04-17
i18n framework

cheehongw:
### Description

Partially addresses #2426, #2468 

This PR adds an i18n framework that we can work with.

- automatically detect browser's language using the navigator detector in `i18next-browser-languagedetector`
- adds a selector in the settings to choose between languages
- persists chosen language in local storage automatically through `i18next-browser-languagedetector`. 
- type safe translation keys so that we can only specify keys that exist in the default language
image

- some demo translations for `NavigationBar` and `Dropdown` and `Login`
- updated test for `Login` page

### Things to note

Some translations were done for  `NavigationBar` and `Dropdown` and `Login` to demonstrate how the strings might be internationalized over time and some of the potential problems we might encounter.

1. **Structure for translation files** 
    ```
    /locales
      /en
        commons.json     //for all the components in commons
        academy.json
        achievement.json
        ...        //one json for each page
    ```
  
2. **Applying translation at the correct level** 

   In `Dropdown`, the `useTranslation` hook and the translate function was applied directly to the strings since these are located within the component itself.

    ```tsx
    const Dropdown: React.FC = () => {
      ...
      const { t } = useTranslation('translation', { keyPrefix: 'dropdown' })
      ...
      const myCourses = isLoggedIn ? (
        <MenuItem icon={IconNames.PROPERTIES} onClick={toggleMyCoursesOpen} text={ t('My Courses') } />
      ) : null;
    ```

   However, the strings in `NavigationBar` are not located in the component. For example, some of the strings are found hardcoded as part of the many list entries in `AcademyNavigationBar.tsx > getStaffNavLinkInfo`. Furthermore, these entries are then passed into `filterNotificationsByType`, where a string comparison is performed between the entry's string and a hardcoded string. If we applied translation directly to the entry's string, a translated string would always fail this comparison, causing some logic to break. Therefore, translation is only applied at the rendering level. 
   
    ```tsx
    <div className={classNames(navbarEntry.hiddenInBreakpoints?.map(bp => `hidden-${bp}`))}>
      <Translation keyPrefix={"navigationBar"}>
        {(t) => t(navbarEntry.text), { defaultValue: navbarEntry.text } 
      </Translation>
    </div>
    ```
   
    ```tsx
      export function filterNotificationsByType(
        assessmentType: AssessmentType
      ): NotificationFilterFunction {
        return (notifications: Notification[]) =>
          notifications.filter(n => {
            if (assessmentType === 'Grading') {
              return n.submission_id !== undefined;
            }
            return !n.submission_id && assessmentType === n.assessment_type;
          });
      }
      ```

https://github.com/source-academy/frontend/assets/72195240/847f3196-4bb6-4df0-a22f-3e11008d2734

### Type of change

### Checklist

- [x] I have tested this code


src/i18n/i18n.ts
```diff
@@ -0,0 +1,21 @@
...
+    fallbackLng: i18nLanguageCode.ENGLISH,
+    debug: true,
+    resources,
+    interpolation: {
+      escapeValue: false, // not needed for react as it escapes by default
+    }
+  });
+
+
+export default i18n;
```
RichDom2185, line 21: Can we add a blank line to the end of this file? Thanks!

src/i18n/locales/en/commons.json
```diff
@@ -0,0 +1,29 @@
...
+  },
+  "dropdown": {
+    "My Courses": "My Courses",
+    "Create Course": "Create Course",
+    "Logout": "Logout",
+    "Settings": "Settings",
+    "About": "About",
+    "Help": "Help"
+  }
+}
```
RichDom2185, line 29: Ditto on newline at EOF.

src/i18n/locales/index.ts
```diff
@@ -0,0 +1,18 @@
...
+export const resources: { [language: string]: any } = {
+  [i18nLanguageCode.ENGLISH]: {
+    name: 'English',
+    translation: translationEN,
+  },
+  [i18nLanguageCode.PSEUDO]: {
+    name: 'Pseudo',
+    translation: translationPseudo
+  }
+}
```
RichDom2185, line 18: Ditto.

src/i18n/locales/pseudo/commons.json
```diff
@@ -0,0 +1,29 @@
...
+  },
+  "dropdown": {
+    "My Courses": "Mÿ Côûrsês",
+    "Create Course": "Crêâtê Côûrsê",
+    "Logout": "Lôgôût",
+    "Settings": "Sêttîñgs",
+    "About": "Âbôût",
+    "Help": "Hêlp"
+  }
+}
```
RichDom2185, line 29: Ditto.

src/commons/dropdown/LocaleSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <Tooltip
+        position={PopoverPosition.TOP}
+        className="form-field-help-text"
+        content="Choose your language."
+      >
+        <Icon icon={IconNames.Help} />
+      </Tooltip>
+    </FormGroup>
+  )
+}
```
RichDom2185, line 41: Ditto.

RichDom2185 (changes requested):
Hi, thanks for working on this! I will comment on the design and implementation later, but for now, can you fix the lint issues (`yarn format`) as well as the comments below? Thanks a lot!

src/i18n/locales/en/login.json
```diff
@@ -0,0 +1,4 @@
+{
+    "Logging In...": "Logging In...",
```
RichDom2185, line 2: Nit: Can we try to avoid using dots in keys to avoid confusion with nested keys?


```suggestion
    "Logging In": "Logging In...",
```

src/i18n/locales/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const resources = {
+  [i18nLanguageCode.ENGLISH]: {
+    name: 'English',
+    ...translationEN
+  },
+  [i18nLanguageCode.PSEUDO]: {
+    name: 'Pseudo',
+    ...translationPSEUDO
+  }
```
RichDom2185, line 24: Maybe we only enable this in development mode using a `process.env.NODE_ENV` check?

src/pages/login/Login.tsx
```diff
@@ -111,7 +113,7 @@ const LoginButton = ({
       rightIcon={IconNames.LOG_IN}
       onClick={useCallback(() => handleClick(id), [handleClick, id])}
     >
-      {`Log in with ${name}`}
+      <Translation ns={'login'}>{t => t('Log in with', { name: name })}</Translation>
```
RichDom2185, line 116: This string can use a normal attribute:

```suggestion
      <Translation ns="login">{t => t('Log in with', { name: name })}</Translation>
```

src/commons/navigationBar/NavigationBar.tsx
```diff
@@ -75,7 +76,17 @@ const NavigationBar: React.FC = () => {
...
       >
         <Icon icon={navbarEntry.icon} />
-        <div>{navbarEntry.text}</div>
+        <div>
+          {
+            <Translation ns={'commons'} keyPrefix={'navigationBar'}>
+              {
+                // @ts-expect-error the t function expects a specific set of strings
+                // as valid translation keys but we are passing in a string type
+                t => t(navbarEntry.text)
```
RichDom2185, line 85: Do you know if it's possible to use a typecast `as SomeType` instead of ignoring the compile error here?

RichDom2185 (commented):
Thank you for the type-safety and autocomplete additions! I don't have much issue with the implementation details as it seems pretty easy to extend or even reorganise should we need to, as we bring in more strings into translation files.

In addition to my comments below, it's also good to note that we plan to bring more than just basic UI strings in the long run (see #1943 for an example). Or even the  sublanguage intro tab (with dynamic links):

image

Seems like it makes more sense to follow your design of using separate namespaces for different parts (e.g. "commons", "login", etc. – on an almost per-component basis) instead of nesting top-level keys on a single namespace.

What do you think of the design?

src/commons/navigationBar/NavigationBar.tsx
cheehongw: we could do something like this:

                t => t(navbarEntry.text as keyof CustomTypeOptions['resources']['commons']['navigationBar'])

src/i18n/locales/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    ...translationEN
+  },
+  ...(process.env.NODE_ENV === 'development'
+    ? {
+        [i18nLanguageCode.PSEUDO]: {
+          name: 'Pseudo',
+          ...translationPSEUDO
+        }
+      }
+    : {})
```
RichDom2185, line 28: I was originally thinking of using an `if (process.env.NODE_ENV === 'development')` and setting the property value, but this actually looks cleaner.

Destructuring like this would also allow only enabling languages that are "in beta" (e.g. very incomplete/in-progress translation maps) to be turned on in development only. Which makes me think: Perhaps we can take this out into a constant `developmentOnlyLanguages`?

src/i18n/locales/pseudo/login.json
```diff
@@ -0,0 +1,4 @@
+{
+    "Logging In": "Lògging Ìn...",
+    "Log in with": "Lòg in with {{name}}"
+}
```
RichDom2185, line 4: Missing newline at EOF.

src/i18n/locales/en/login.json
```diff
@@ -0,0 +1,4 @@
+{
+    "Logging In": "Logging In...",
+    "Log in with": "Log in with {{name}}"
+}
```
RichDom2185, line 4: Missing newline at EOF.

src/i18n/i18nForTests.ts
RichDom2185, line 1: Nit: the formatting of this file with the extra newlines makes it inconsistent with the non-test i18n config. Can we make it consistent?

src/commons/navigationBar/NavigationBar.tsx
```diff
@@ -75,7 +77,13 @@ const NavigationBar: React.FC = () => {
         key={navbarEntry.text}
       >
         <Icon icon={navbarEntry.icon} />
-        <div>{navbarEntry.text}</div>
+        <div>
+          {
+            <Translation ns="commons" keyPrefix="navigationBar">
+              {t => t(navbarEntry.text as keyof i18nDefaultLangKeys['commons']['navigationBar'])}
+            </Translation>
+          }
```
RichDom2185, line 85: Unnecessary extra curly braces.

src/commons/dropdown/LocaleSelector.tsx
```diff
@@ -0,0 +1,41 @@
+import { FormGroup, HTMLSelect, Icon, PopoverPosition, Tooltip } from '@blueprintjs/core';
+import { IconNames } from '@blueprintjs/icons';
+import { useState } from 'react';
+import i18n from 'src/i18n/i18n';
+import { i18nLanguageCode, resources } from 'src/i18n/locales';
+
+const languageOptions = Object.keys(resources)
+  .filter(String)
```
RichDom2185, line 8: Can I check what's this filter guarding against?

cheehongw: because we conditionally spread the psuedo language resources into `resources`, there could be an undefined` key if we spread an empty object `{ }`

im not _entirely_ certain of this behavior though, the type hints suggests that this could be a possibility, hence the filter.

cheehongw:
thanks for the review @RichDom2185 

> we plan to bring more than just basic UI strings in the long run (see https://github.com/source-academy/frontend/issues/1943 for an example). Or even the sublanguage intro tab (with dynamic links):

i18next supports complex UI strings with the [`Trans` component ](https://react.i18next.com/latest/trans-component). The next step would definitely be identifying and externalizing all the English strings in the repo to follow i18n. 

Actual translation can take place alongside this process.

> design of using separate namespaces for different parts 

Originally, i wanted to have a json file for each component containing strings, however, combining them into a nested structure is tricky and each separate json file is considered a namespace by i18next. Furthermore, the location of a string and where the translation is applied might be different, so there is less meaning to having one json file to each component.

I think namespacing will become clearer as more strings get externalized. My only concern is that the `"commons"` namespace will contain majority of the translated strings since most components are located under the `commons` directory, which might make it difficult for a translator to track down a string belonging to which component.

RichDom2185 (approved):
> I think namespacing will become clearer as more strings get externalized. My only concern is that the `"commons"` namespace will contain majority of the translated strings since most components are located under the `commons` directory, which might make it difficult for a translator to track down a string belonging to which component.

Let's keep this in mind as we anticipate future PRs to add more translation maps. LGTM, thank you for laying the groundwork for internationalization!

PR #2950, opened 2024-04-21
Java CSEC Bug Fix

### LINKED ISSUE #2949: Java CSEC supports Inheritance but not Compiler
Inheritance is supported in the Java CSEC machine so the code below is valid syntax. However, nothing happens when the run button is clicked. This is because the call to the CSEC machine, i.e., `runJavaCseMachine()` happens after the call to the compiler, i.e., `compileFromSource()`, but the compiler currently does not support inheritance. Note that the CSEC machine and compiler are using different parsers.

image

xyliew25:
### Description

Fixes #2949 

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/utils/JavaHelper.ts
+++ b/src/commons/utils/JavaHelper.ts
@@ -35,6 +35,8 @@ export async function javaRun(
     return Promise.resolve({ status: 'error' });
   }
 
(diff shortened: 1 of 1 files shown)


RichDom2185 (approved):
LGTM, thanks!

PR #2952, opened 2024-04-21
Migrate to Redux Toolkit part 6

RichDom2185:
### Description

Now that we are using the builder pattern, we can more easily split the code for better clarity and also better enforce some separation.

### Type of change

### Checklist

- [x] I have tested this code


RichDom2185 (commented):
@chownces @JoelChanZhiYang can I get your thoughts on the splitting? Does it make sense (as well as the file naming/structure)? If yes, I'll proceed to split the other cases as well.

Thanks!

JoelChanZhiYang:
It looks good! The naming of "multiFileReducer" might be a bit confusing to people newer to Source Academy. I would not explicitly separate the multi-file and the single file (since the single file is simply a subset of the multi-file).

I would just call it the editorReducer.

Some parts of the multiFileReducer would be good contenders for a future refactor. It seem strange that cursor location is stored in global state.

chownces:
Debugged and fixed the failing tests following the RTK migration as discussed offline with @RichDom2185 

Analysis:
- RTK discourages the use of non-serializable structures like Map due to use of the Immer library
- As a result, the map data structures in the redux store were frozen
- The lodash.merge operation in `createStore.ts` fails silently when it is unable to merge in the new map
- Thus, the mock store.session.assessments was an empty default map, resulting in the snapshots failing

- Also added `await` to the backend saga tests to ensure that the redux-saga-test-plan tests were actually running to completion

chownces (approved):

src/commons/assessmentWorkspace/__tests__/AssessmentWorkspace.tsx
```diff
@@ -71,7 +72,13 @@ const mockContestVotingAssessmentWorkspaceProps: AssessmentWorkspaceProps = {
 
 const mockStore = mockInitialStore({
   session: {
-    assessments: new Map(mockAssessments.map(assessment => [assessment.id, assessment]))
+    assessments: mockAssessments.reduce(
+      (acc, assessment) => {
+        acc[assessment.id] = assessment;
+        return acc;
+      },
+      {} as { [id: number]: Assessment }
```
RichDom2185, line 80: Instead of a typecast, might be better to use a type argument instead.

src/commons/sagas/__tests__/BackendSaga.ts
```diff
@@ -131,7 +131,13 @@ import {
...
 const mockAssessment: Assessment = mockAssessments[0];
 
-const mockMapAssessments = new Map<number, Assessment>(mockAssessments.map(a => [a.id, a]));
+const mockMapAssessments = mockAssessments.reduce(
+  (acc, assessment) => {
+    acc[assessment.id] = assessment;
+    return acc;
+  },
+  {} as { [id: number]: Assessment }
+);
```
RichDom2185, line 140: Ditto

src/commons/sagas/__tests__/BackendSaga.ts
```diff
@@ -333,6 +333,9 @@ describe('Test FETCH_AUTH action', () => {
   const courseRegistration = mockCourseRegistration1;
   const assessmentConfigurations = mockAssessmentConfigurations;
 
+  // API call is made when dispatching subsequent action, causing console warning
+  jest.spyOn(global, 'fetch').mockReturnValue(Promise.resolve({} as Response));
```
RichDom2185, line 337: Ditto on using type argument.

RichDom2185: Whoops, tested and this doesn't work. Nevermind.

src/commons/assessmentWorkspace/__tests__/AssessmentWorkspace.tsx
RichDom2185: Resolved in 1ea8f148484e25ceeb80e71080ca82784e702cba.

src/commons/sagas/__tests__/BackendSaga.ts
RichDom2185: Resolved in 1ea8f148484e25ceeb80e71080ca82784e702cba.

RichDom2185 (commented):
LGTM, thanks for the help! @chownces

PR #2971, opened 2024-05-02
Migrate to Redux Toolkit part 7

RichDom2185:
### Description

Based on @leeyi45's work done in #2874.

Intentionally split and only migrated some stuff to keep the diff readable.

I strongly recommend going through the changes commit by commit instead of going through the full diff of this PR. Please refer to the commit history for a list of changes.

### Type of change

### How to test

Everything should still work. Tests involving `createStore` are still failing at the moment, will fix in the next few days.

### Checklist

- [ ] I have tested this code


chownces:
Turns out `saferTakeEvery` is not so safe 🤓. Added the fix.

RichDom2185:
> Turns out `saferTakeEvery` is not so safe 🤓. Added the fix.

Thanks! My bad, I missed out the fact that `saferTakeEvery` takes in the action and not the action type (like `takeEvery`) 😅

src/commons/mocks/BackendMocks.ts
RichDom2185, line 1: @chownces Technically, it's not a really big deal if we want to maintain this file a bit longer, but&hellip;

chownces: Is it tedious to migrate this file over to RTK and Sagas? If so, i think we can just remove it and update the docs

src/commons/utils/TestUtils.ts
```diff
@@ -47,3 +47,19 @@ export function deepFilter<T>(
   helper(nestedObject);
   return matches;
 }
+
+// TODO: Fix type inference for this function,
```
chownces, line 51: Nit

chownces (approved):

src/commons/mocks/BackendMocks.ts
RichDom2185: Not tedious to migrate to RTK, but tedious to migrate to our new helper functions. I'll just migrate to RTK using the `actionName.type` instead of `ACTION_TYPE` for now. I'll migrate to the new helper functions later.

src/commons/utils/TestUtils.ts
RichDom2185: Will be done in Part 8

RichDom2185 (commented):
Tested locally and everything still works, will merge now and make the rest of the changes in subsequent parts.

Issue #2484, opened 2023-05-16
Upgrade react-router-dom to v6


chownces:
Part of the dependency updates required to migrate to react v18 properly

chownces:
Initial groundwork done in #2495.

Further work needed in the future:
- #2505 
- hoisting of Academy routes to routerConfig.tsx

PR #2978, opened 2024-05-02
Refactor Academy router

RichDom2185:
### Description

* Create `GuardedRoute` class for builder pattern
* Reformat app router config for readability
* Migrate Academy routes to use new Data Loader API
* Use lazy loading + code-split Academy routes for performance #2499
* Refactor routing logic (assessment configurations, not found behavior)

Please refer to commit history for details.

See #2507
See #2484

**TODO:**

- [x] Fix infinite render update loop when visiting admin panel
- [x] Fix tests (tried following https://reactrouter.com/en/main/routers/create-memory-router without luck)
- [x] Potentially refactor `routerConfig.tsx` to use `GuardedRoute` where possible

### Type of change

### How to test

Everything should still work.

### Checklist

- [x] I have tested this code


src/pages/academy/academyRoutes.tsx
```diff
@@ -0,0 +1,108 @@
...
+  { path: 'groundcontrol', lazy: GroundControl },
+  { path: `grading/${gradingRegExp}`, lazy: Grading },
+  { path: 'sourcereel', lazy: Sourcereel },
+  { path: 'gamesimulator', lazy: GameSimulator },
+  { path: 'teamformation', lazy: TeamFormation },
+  { path: 'teamformation/create', lazy: TeamFormationForm },
+  { path: `teamformation/edit/${teamRegExp}`, lazy: TeamFormationForm },
+  { path: 'teamformation/import', lazy: TeamFormationImport },
+  { path: 'dashboard', lazy: Dashboard }
+].map(r => new GuardedRoute(r).check(s => s.session.role !== Role.Student, notFoundPath).build());
```
chownces, line 91: Let's change this to a whitelist instead.

chownces (dismissed):

src/pages/academy/academyRoutes.tsx
RichDom2185: Resolved in 90dba2e1f75fad50fc5842915413181fa44dc5b0.

RichDom2185:
> * [x]  Fix infinite render update loop when visiting admin panel
> * [x]  Fix tests (tried following https://reactrouter.com/en/main/routers/create-memory-router without luck)
> * [x]  Potentially refactor `routerConfig.tsx` to use `GuardedRoute` where possible

@chownces I've fixed the infinite rerender issue, and also refactored `ensureUserAndRole` to use `GuardedRoute`, thereby eliminating the dependency on `role`. I think in the future we should move towards a static router (thereby removing the need for `UPDATE_REACT_ROUTER` action) to decrease complexity – just access the store directly via `store.getState()` within loader functions whenever we need to have dynamic routing logic.

This will reduce the complexity of the implementation.

chownces (commented):
There's currently a bug in the academy routes when the route is non-existent. Instead of 404-ing, the page hangs (likely in an infinite loop).

Try something like `http://localhost:8000/courses/11/sourcecastt`

RichDom2185:
> There's currently a bug in the academy routes when the route is non-existent. Instead of 404-ing, the page hangs (likely in an infinite loop).
> 
> Try something like `http://localhost:8000/courses/11/sourcecastt`

Somehow the new 404 behaviour seems to be causing the infinite loop. I've restored the original behaviour of rendering `<NotFound />` as a catch-all, instead of redirecting first to a `/not_found` path, then rendering the component there. This fixes the issue.

chownces (approved):
LGTM. Thanks

PR #2979, opened 2024-05-03
Update waitFor usage

### LINKED ISSUE #2975: Inconsistent snapshot test between different test runs
Following changes to test files in #2952 and #2961, I've been finding myself having to re-trigger workflows often because some tests would just randomly fail.

When they do fail, it's always due to a snapshot mismatch, specifically regarding the icon component SVGs not being populated.

Restarting the workflow solves it.

This affects local test environments as well the CI environment. Sometimes the PR passes CI and gets merged only for it to fail the `master` branch CI later (until the workflow is reran), which is annoying.

Reference logs:

https://github.com/source-academy/frontend/actions/runs/8912047561/attempts/1
https://github.com/source-academy/frontend/actions/runs/8912188643/attempts/1

chownces:
### Description

Attempting to fix the non-det snapshot testing which fails randomly.

### Type of change

### Checklist

- [ ] I have tested this code

Closes #2975.

### DIFF
--- a/src/commons/assessmentWorkspace/__tests__/AssessmentWorkspace.tsx
+++ b/src/commons/assessmentWorkspace/__tests__/AssessmentWorkspace.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
 import { require as acequire } from 'ace-builds';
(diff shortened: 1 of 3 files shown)


RichDom2185 (approved):
LGTM, thanks! Let's try this, and if the problem still persists, we'll reopen the issue.

PR #2988, opened 2024-05-05
Migrate to Redux Toolkit part 8

RichDom2185:
### Description

Migrated more sagas and action creators to use the new helpers.

Codebase is still a bit messy as it's in the midst of refactoring but it will be cleaned up in subsequent parts.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Code quality improvements

### How to test

Everything should still work.

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #2989, opened 2024-05-05
Migrate to Redux Toolkit part 9

RichDom2185:
### Description

To rebase and merge onto `master` once #2988 is merged.

Mainly a PR to clean up things, but also migrated to RTK for Sourcecast and Sourcereel reducers.

Removes the compatibility layer of having named exports for every action creator. This forces callers to use the default export (and also fully qualified names) of the actions they want to dispatch. Hopefully, this will reduce the chances of any name conflicts, and also enforce better abstractions and grouping of the individual actions. Currently, I feel that some action creators don't belong in the correct file. By forcing the usage of the fully qualified name (especially as the number of actions keeps increasing), people might think twice before putting an action creator inside a specific file.

Technically a breaking change.

### Type of change

### How to test

I just changed the imports, without any implementation change (except for the migration of `GitHubPersistenceSaga` to use the new helper) – so it should be a non-issue.

### Checklist

- [x] I have tested this code


src/commons/application/actions/SessionActions.ts
RichDom2185, line 1: @sayomaki @chownces Just wanted to check if you're agreeable to removing the named exports to favour the default exports (and fully qualified names).

src/features/sourceRecorder/sourcecast/SourcecastReducer.ts
```diff
@@ -1,69 +1,49 @@
...
-  SAVE_SOURCECAST_DATA,
-  SET_CODE_DELTAS_TO_APPLY,
-  SET_CURRENT_PLAYER_TIME,
-  SET_INPUT_TO_APPLY,
-  SET_SOURCECAST_DATA,
-  SET_SOURCECAST_PLAYBACK_DURATION,
-  SET_SOURCECAST_PLAYBACK_STATUS
-} from '../SourceRecorderTypes';
-import { SourcecastWorkspaceState, UPDATE_SOURCECAST_INDEX } from './SourcecastTypes';
+  saveSourcecastData,
```
chownces, line 5: Im agreeable to removing the named exports.

src/features/sourceRecorder/sourcereel/SourcereelReducer.ts
```diff
@@ -1,80 +1,53 @@
...
-  RECORD_INPUT,
-  RESET_INPUTS,
-  SourcereelWorkspaceState,
-  TIMER_PAUSE,
-  TIMER_RESET,
-  TIMER_RESUME,
-  TIMER_START,
-  TIMER_STOP
-} from './SourcereelTypes';
+  recordInit,
```
chownces, line 6: .

chownces (approved):
`SessionActions` seems quite large and cross-cutting. Perhaps we can consider splitting it up into multiple actions with well-defined names. But that can be done in another PR too. We can add an issue to track it in the meantime

src/features/sourceRecorder/sourcereel/SourcereelReducer.ts
RichDom2185: As above.

RichDom2185:
> We can add an issue to track it in the meantime

Done in #2992.

RichDom2185 (commented):
Went through the changes one more time and looks good, thanks!

PR #2998, opened 2024-05-12
Refactor academy navigation bar

### LINKED ISSUE #2722: Move out of Blueprint's CSS API

RichDom2185:
### Description

By moving the rendering logic to use conditional rendering instead of CSS classes, we can ensure correct margin behaviour (as the margin is dependent on the whether a child exists).

Ideally, we remove the old `flexbox-grid` dependency as well in the future.

image

Also refactored to use components instead of render functions.

Refer to commit history for more detailed list of changes.

Also resolves (partially) #2722.

### Type of change

### Checklist

- [x] I have tested this code


RichDom2185 (commented):
Looked through the changes again and while it's not the most ideal separation, the logic seems cleaner now and I think it's okay for an incremental improvement.

We can always refactor further later together with the rest of the routes.

src/commons/navigationBar/NavigationBar.tsx
```diff
@@ -42,10 +42,52 @@ export type NavbarEntryInfo = {
...
+      {renderDrawer && (
+        <Button
+          onClick={() => setMobileSideMenuOpen(!mobileSideMenuOpen)}
+          icon={IconNames.MENU}
+          large={true}
+          minimal={true}
+        />
+      )}
+      <NavLink
+        className={classNames('NavigationBar__link', Classes.BUTTON, Classes.MINIMAL)}
```
sayomaki, line 64: Can we refactor this to use the new Button component as well? (since we are trying to move away from CSS)

src/commons/utils/Hooks.ts
```diff
@@ -119,7 +119,18 @@ export const useDimensions = (ref: RefObject<HTMLElement>): [width: number, heig
  */
 export const useResponsive = () => {
   const isMobileBreakpoint = useMediaQuery(`(max-width: ${Constants.mobileBreakpoint}px)`);
+
+  // Based on values from flexboxgrid-helpers
```
sayomaki, line 123: With the new breakpoints added for the various sizes, can we consider replacing the `isMobileBreakpoint` and `isDesktopBreakpoint` to use these breakpoints instead?

sayomaki (commented):
I have taken a quick look at the changes, and here are some comments

src/commons/utils/Hooks.ts
RichDom2185: Maybe not now, seems out of scope for this PR. But definitely we should try to simplify all these and combine stuff together.

src/commons/navigationBar/NavigationBar.tsx
RichDom2185: Resolved in c1c8532fc4ad887900e7c7c87df77114fcaea20d.

sayomaki (approved):
Looks good to me! In that case, we will need to refactor the mobile breakpoint media queries in a future PR.

RichDom2185 (commented):
Looked through the changes one more time and all seems good!

PR #3016, opened 2024-08-11
Create NUS login page

RichDom2185:
### Description

As per title. For compliance with requirements.

### Type of change

### Checklist

- [x] I have tested this code


src/pages/login/LoginPage.tsx
```diff
@@ -4,22 +4,29 @@ import classNames from 'classnames';
...
 import SessionActions from 'src/commons/application/actions/SessionActions';
 import classes from 'src/styles/Login.module.scss';
 
 import Constants from '../../commons/utils/Constants';
 
-const providers = [...Constants.authProviders.entries()].map(([id, { name }]) => ({
-  id,
-  name
-}));
+const providers = [...Constants.authProviders.entries()]
```
sayomaki, line 13: Would it be better to just use a setting for NUS deployment? In that case we would not need to overcomplicate processing the providers

src/pages/login/NusLogin.tsx
```diff
@@ -0,0 +1,87 @@
...
+import sourceAcademyLogo from '../../assets/SA.jpg';
+
+const nusProviders = [...Constants.authProviders.entries()]
+  .map(([id, { name }]) => ({
+    id,
+    name
+  }))
+  .filter(e => e.name.includes('NUS'));
+
+const NUS_LOGO =
```
sayomaki, line 19: This should ideally be in its own asset file instead of being hardcoded

src/pages/login/NusLogin.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const BLUE_BG =
```
sayomaki, line 22: This should also not be hardcoded

sayomaki (commented):
Some nitpicks and feedback, other than that everything is looking good!

src/pages/login/NusLogin.tsx
RichDom2185: It's only used in one page, that's why I thought no need to put in an asset. Also, it's NUS-specific, not Source Academy, so it felt weird to put under Source Academy assets.

src/pages/login/LoginPage.tsx
RichDom2185: Yeah, makes sense. Will refactor.

sayomaki (approved):
Looks good to me! Let's have this deployed, thanks for the work!

PR #3017, opened 2024-08-12
Add payload size limit

yiwen101:
### Type of change

### Checklist

- [ ] I have tested this code


yiwen101:
Corresponding BE PR: https://github.com/source-academy/backend/pull/1150

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
```diff
@@ -42,6 +48,10 @@ const ChatBox: React.FC<Props> = ({ getSection, getText }) => {
     if (userInput.trim() === '') {
       return;
     }
+    if (userInput.length > maxContentSize) {
+      setMessages(prev => [...prev, MESSAGE_TOO_LONG_MESSAGE]);
+      return;
+    }
```
RichDom2185, line 54: I don't think we should error when the user submits. Instead, the input field should not allow more than `maxContentSize` characters.

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
```diff
@@ -22,6 +22,11 @@ const BOT_ERROR_MESSAGE: Readonly<ChatMessage> = {
   role: 'assistant'
 };
 
+const MESSAGE_TOO_LONG_MESSAGE: Readonly<ChatMessage> = {
+  content: 'Your message is too long. Please try again with a shorter message.',
+  role: 'assistant'
+};
```
RichDom2185, line 28: That way, this will never be necessary. From UX perspective, it's easier for the user to edit their message before sending in order to fit the character limit, as opposed to having to copy and paste their previous message and send it again. Not to mention, if you do it like this, there will be a mismatch between the conversation history stored in the FE state with the one stored in the DB

RichDom2185 (changes requested):

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
RichDom2185: This can be a simple indicator of `maxContentSize - userInput.length` and set the `onChange` to ignore extra characters beyond `maxContentSize`

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
yiwen101: Hi, I think it's better to stay with the current design. By not reseting "user input" to empty string, use can still edit their message as you suggested;

This would be similar to the behavior of chatGPT bot. If the message is too long, after sending the message to chatGPT, user receives an error message and the input field remains the same. So I find this a more intuitive behavior.

I feel it is OK for the mismatch between the conversation history as blocking messages that are too long before it reaching backend is the purpose of this feature.

One customization I made is to not show the "overly long" message in the chat field so it is less painful to scroll up and refer to earlier conversation.

yiwen101: You are right; 
it is more intuitive to do let the user trigger this in the first place;
this also pollute the chat history, making it less readable.

src/pages/sicp/subcomponents/chatbot/ChatBox.tsx
yiwen101: Thanks for the comment. Feature added.

yiwen101:
@RichDom2185 Could you review this again. Is it ready for merge?

PR #3025, opened 2024-08-21
feat: Show bonus xp end date

### LINKED ISSUE #355: Display time countdown/deadline for max attainable XP per mission & quest
To aid students to prioritize different assignments accordingly to the deadline where they can still get max XP

### LINKED ISSUE #1964: Display full bonus XP deadline in Assessment overview card
^ Suggestion by a student. For students to explicitly know when the bonus XP starts to decay.

GabrielCWT:
This change adds a icon which gives a tooltip for when the bonus xp will start to decay. 
Relevent Backend PR: https://github.com/source-academy/backend/pull/1156

Closes #1964

Closes #355


src/styles/_academy.scss
```diff
@@ -189,6 +189,11 @@
     flex-direction: column;
   }
 
+  .listing-xp {
+    display: flex;
+    gap: 0.5rem;
+  }
```
RichDom2185, line 195: Can we start to move newly created styles to a separate CSS module for better abstraction (instead of using global CSS)?

src/commons/utils/DateHelper.ts
```diff
@@ -41,3 +41,9 @@ export const getStandardDate = (dateString: string): string => {
   const prettyDate = date.format('MMMM Do YYYY');
   return prettyDate;
 };
+
+export const getPrettyDateAfterHours = (dateString: string, hours: number): string => {
+  const date = moment(dateString).add(hours, 'hours');
+  const prettyDate = date.format('Do MMMM, HH:mm');
+  return prettyDate;
+};
```
RichDom2185, line 49: Would it be better if we use Moment's `.fromNow()` to get an even "prettier" format? E.g. "in 4 hours", "in 2 days"

RichDom2185: What do you think? Not a suggestion per se, but more to spark a discussion

RichDom2185 (changes requested):
Just some minor nits :)

src/commons/utils/DateHelper.ts
GabrielCWT: I think this is up to personal preference of the user. Personally, I would like to see the absolute deadline but I definitely understand that there wil be users who prefer a relative deadline. Maybe we can do both? e.g. " 22nd August, 21:00 (in 3 hours)"

RichDom2185: That's fair, let's try it out

src/styles/Academy.module.scss
GabrielCWT, line 1: Would this be what you are refering to for CSS Module. Also I placed it in the Academy module to follow the current convention that Assessment stylings are found in _academy

PR #3045, opened 2024-09-13
bumping js-slang

martin-henz:


RichDom2185:
Hi Prof, I think you forgot to commit `yarn.lock` resulting in CI failure

RichDom2185 (changes requested):
Per comment above

RichDom2185 (approved):
LGTM, thanks!

PR #3049, opened 2024-09-22
Use `DOMPurify` to sanitize and mitigate XSS attacks

TheMythologist:
### Description

An XSS vulnerability is present in the `Markdown` component, where unsanitized user input is directly passed into `dangerouslySetInnerHTML`. For more info, refer to [here](https://github.com/showdownjs/showdown/wiki/Markdown's-XSS-Vulnerability-(and-how-to-mitigate-it)).

_Note_: The changes in UI test snapshots are merely the order of the `a` tag attributes - they are essentially the same.

### Type of change

### How to test

An example would be entering the text `` in the "Module Help Text" of the Admin Panel page when creating a new course. After sanitization, the `alert('xss')` command should not be run.

### Checklist

- [x] I have tested this code


sayomaki (commented):
Hi, thanks for catching this and working on this! The changes looks good to me, and I think we can go ahead and have this merged once @RichDom2185 takes a look as well.

Issue #2814, opened 2024-02-23
Export CSV function regression after transfer of pagination of grading submissions from frontend to backend


josh1248:
Relevant PRs:
Frontend: https://github.com/source-academy/frontend/pull/2787
Backend: https://github.com/source-academy/backend/pull/1065

The backend pagination has diminished the export CSV function, which can now only save the backend entries it gets.

Possible fixes:
- An updated function could take the total number of grading submissions, suppose X, from the backend response. Then, it can submit a query with {offset: 0, pageSize: X} to circumvent the LIMIT implementation in the backend.

PR #3064, opened 2024-10-27
Resolve loading indicator issues in grading table

### LINKED ISSUE #3048: Grading table refresh button bugged
Clicking refresh on the grading table causes it to send a backend request with query param `group = false` even though the drop down is still set to "submissions from my group"

josh1248:
Resolves #3048

Good to do before tackling #2814 so that loading can be more explicity indicated.

This PR resolves the loading indicator bug in the grading table, whereby a request for a new submissions page does not properly get reflected as "loading" until backend data arrives. This problem likely arised due to the use of a deprecated API from AGGrid.

To accompany this change, additional measures have been added to prevent users from spamming new requests while awaiting for data - this disabling is now made possible by the proper loading state.

### Type of change

### How to test

- In the backend, set a high number of students in `priv/repo/seeds.exs`, example 20000 students and 100 questions. It should be enough such that backend fetching takes at least 1 second.
- Go to the grading table and attempt to load the next page. Notice that the loading logo stays until new data is received.
- Attempt to "spam" the next page button. Notice that you are now unable to do so due to the page change locking while data is being loaded.
- Attempt to continuously hit the refresh button. Notice that it is also disabled while data is being loaded.
- Note that this lock is not applied on the dropdown list as there is no HTML API to disable it easily - but this shouldn't be an issue since it cannot be repeatedly clicked in a short time.

### Checklist

- [ ] I have tested this code


josh1248:
ready for review!

sayomaki (commented):
Did a preliminary review of this PR, and it looks great! Thanks for the improvements!

Just to clarify but from what I see this does not seem to resolve #3048, does it?

RichDom2185 (approved):
Also pushed some cosmetic changes for better UI and code quality, LGTM, thanks!

josh1248:
> Did a preliminary review of this PR, and it looks great! Thanks for the improvements!
> 
> Just to clarify but from what I see this does not seem to fully resolve #3048, does it?

Hi sayo, apologies for the late reply - it does solve refresh issues for next page / previous page / filters, but not for the dropdown menus for page size or grouping. This is because there wasnt an API to disable them - and I thought it wouldn't be an issue since you couldn't spam backend requests through the dropdown menus easily. Thank you for the review!

PR #3076, opened 2024-11-13
Fix export CSV to retrieve all assessement entries

### LINKED ISSUE #2814: Export CSV function regression after transfer of pagination of grading submissions from frontend to backend

josh1248:
Resolves #2814 

### Description

Resolves the issue of export CSV being unable to export all entries due to backend pagination. The solution is to connect to a backend endpoint outside of the Redux loop so that data can be exported independently of what is shown to the user. (This will be an admin-only button to avoid excess server fees)

### Type of change

### How to test

Start up both the updated backend and updated frontend. 
- Enter the 'Grading' tab.
- Click the 'Export CSV' button at the grading table.
- If you are logged in as an admin, you will receive a notification to download the CSV file.
- If you are logged in as a staff, you will receive a 401 error.

### Checklist

- [ ] I have tested this code

### DIFF
--- a/src/pages/academy/grading/Grading.tsx
+++ b/src/pages/academy/grading/Grading.tsx
@@ -7,6 +7,7 @@ import SessionActions from 'src/commons/application/actions/SessionActions';
 import { Role } from 'src/commons/application/ApplicationTypes';
 import GradingFlex from 'src/commons/grading/GradingFlex';
 import GradingText from 'src/commons/grading/GradingText';
+import { getAllGradingOverviews } from 'src/commons/sagas/RequestsSaga';
 import SimpleDropdown from 'src/commons/SimpleDropdown';
 import { useSession, useTypedSelector } from 'src/commons/utils/Hooks';
 import WorkspaceActions from 'src/commons/workspace/WorkspaceActions';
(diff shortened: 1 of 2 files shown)


src/pages/academy/grading/Grading.tsx
```diff
@@ -58,6 +59,10 @@ const Grading: React.FC = () => {
   const allColsSortStates = useTypedSelector(state => state.workspaces.grading.allColsSortStates);
   const hasLoadedBefore = useTypedSelector(state => state.workspaces.grading.hasLoadedBefore);
   const requestCounter = useTypedSelector(state => state.workspaces.grading.requestCounter);
+  const tokens = useTypedSelector(state => ({
+    accessToken: state.session.accessToken || 'Invalid',
+    refreshToken: state.session.refreshToken || 'Invalid'
+  }));
```
RichDom2185, line 65: I'm not a big fan of this, can we just do the following?

```suggestion
  const accessToken = useTypedSelector(state => state.session.accessToken);
  const refreshToken = useTypedSelector(state => state.session.refreshToken);
```

src/pages/academy/grading/Grading.tsx
```diff
@@ -165,7 +170,9 @@ const Grading: React.FC = () => {
                 <Button
                   minimal
                   icon={IconNames.EXPORT}
-                  onClick={() => exportGradingCSV(gradingOverviews.data)}
+                  onClick={() => {
+                    getAllGradingOverviews(tokens).then(resp => exportGradingCSV(resp?.data));
+                  }}
```
RichDom2185, line 175: Then, we use a non-null assertion instead (which should be safe since the page will auto-redirect if the token is not in the state)

```suggestion
                  onClick={() => {
                    const tokens = {
                      accessToken: accessToken!,
                      refreshToken: refreshToken!
                    };
                    getAllGradingOverviews(tokens).then(resp => exportGradingCSV(resp?.data));
                  }}
```

RichDom2185 (changes requested):
Just some minor nits :)

RichDom2185 (approved):
LGTM, thanks!

PR #3080, opened 2025-01-14
feat: add vscode slice for message passing with extension

heyzec:
### Description

FYP: Source Academy as Visual Studio Code Extension

This PR makes the frontend's behavior change conditionally when it detects itself to be running as an iframe in the VS Code extension. This is required so that it can send and receive the editor text contents, encapsulated as messages, to the VS Code extension.

### Type of change

### Checklist

- [ ] I have tested this code


RichDom2185 (commented):
Hi, thanks for working on this! Could you update the test snapshots?

src/commons/application/actions/VscodeActions.ts
```diff
@@ -0,0 +1,10 @@
+import { createActions } from 'src/commons/redux/utils';
+
+const VscodeActions = createActions('vscode', {
+  setVscode: () => ({})
```
RichDom2185, line 4: Nit: falsy values mean empty payload by default

```suggestion
  setVscode: 0
```

src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
```diff
@@ -184,11 +185,11 @@ const AssessmentWorkspace: React.FC<AssessmentWorkspaceProps> = props => {
...
-  useEffect(() => {
-    // TODO: Hardcoded to make use of the first editor tab. Refactoring is needed for this workspace to enable Folder mode.
-    handleEditorValueChange(0, '');
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  // useEffect(() => {
+  //   // TODO: Hardcoded to make use of the first editor tab. Refactoring is needed for this workspace to enable Folder mode.
+  //   handleEditorValueChange(0, '');
+  //   // eslint-disable-next-line react-hooks/exhaustive-deps
+  // }, []);
```
RichDom2185, line 192: Will this break anything with assessments? Have you tested locally?

src/commons/workspace/Workspace.tsx
```diff
@@ -222,7 +223,11 @@ const Workspace: React.FC<WorkspaceProps> = props => {
         </Resizable>
         <div className="row content-parent" ref={contentContainerDiv}>
           <div className="editor-divider" ref={editorDividerDiv} />
-          <Resizable {...editorResizableProps()}>{createWorkspaceInput(props)}</Resizable>
+          {isVscode ? (
+            <div style={{ width: '0px' }}>{createWorkspaceInput(props)}</div>
```
RichDom2185, line 227: It must still be visible is it? We can't use `visibility: hidden`? If we can't, then in addition to `width: 0`, should we also use `overflow: hidden`?

src/features/vscode/messages.ts
```diff
@@ -0,0 +1,58 @@
+// This file also needs to be copied to source-academy/frontend
```
RichDom2185, line 1: Maybe for the one committed to the frontend, you can change the comment to point to the original file in the extension repo?

RichDom2185 (commented):
Just some minor nits :)

src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
heyzec: From my local testing, there's no unintended behavior. Assessments still loading as per normal.

handleEditorValueChange is ultimately called with the editor contents due to the useEffect [here](https://github.com/source-academy/frontend/blob/4e43c66fd338b9bd2870651853fb3fd177e03a22/src/commons/assessmentWorkspace/AssessmentWorkspace.tsx#L252).

src/commons/workspace/Workspace.tsx
heyzec: While `visibility: hidden` wouldn't work, I think we can completely remove it now that messages are passed directly to the redux store rather than to the editor.

src/commons/application/Application.tsx
```diff
@@ -70,6 +73,47 @@ const Application: React.FC = () => {
...
   }, [isPWA, isMobile]);
 
+  // Effect to fetch the latest user info and course configurations from the backend on refresh,
+  // if the user was previously logged in
+  React.useEffect(() => {
+    // Polyfill confirm() to instead show as VSCode notification
+    window.confirm = () => {
+      console.log('You gotta confirm!');
+      return true;
+    };
```
RichDom2185, line 83: Should we use `if (window.confirm) ...` instead?

src/commons/assessmentWorkspace/AssessmentWorkspace.tsx
RichDom2185: Let's just delete and clean up all this unused code then

src/commons/application/Application.tsx
```diff
@@ -70,6 +73,47 @@ const Application: React.FC = () => {
...
+
+    const message = Messages.WebviewStarted();
+    sendToWebview(message);
+
+    window.addEventListener('message', event => {
+      const message: MessageType = event.data;
+      // Only accept messages from the vscode webview
+      if (!event.origin.startsWith('vscode-webview://')) {
+        return;
```
RichDom2185, line 92: I know this is a guard, but shall we log such an event?

Technically we might be able to configure our deployment with `X-Frame-Options: DENY` to prevent embedding but let's look into it another time

RichDom2185 (changes requested):
Just some minor nits :)

src/commons/application/Application.tsx
heyzec: Note to self: also add Content-Security-Policy frame-ancestors 'none';

heyzec: Adding header `X-Frame-Options: DENY` does not compromise the usability of the VSC extension, but `Content-Security-Policy: frame-ancestors 'none';` will cause the request to be blocked within VSC.

RichDom2185 (approved):
LGTM, thanks!

PR #3083, opened 2025-01-30
update session details with latest course information before switchin…

### LINKED ISSUE #3040: Switching to course as an admin bring to game page, even when game is not enabled
This results in 404 not found error

C-likethis123:
…g courses

### Description

Fixes the bug where switching courses redirects the user to the 'game' page, even when games are not enabled for the course.

The reason for this bug is because the loader function at `src/pages/academy/academyRoutes` checks for the `enableGame` flag in the Redux store before deciding whether to redirect, but the `enableGame` flag value belongs to the previous course and not the course it's about to switch to. The solution adopted in this MR is to update the latest viewed course to the course that the user is about to navigate to, before actually navigating to that course.

Closes #3040.

### Type of change

### How to test

1. Course configuration: create two courses - one with games enabled and one without games enabled.
2. Switch to the other course using the 'My courses' dropdown - it should redirect to the games page if the course has games enabled, and to the missions page otherwise.

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/dropdown/DropdownCourses.tsx
+++ b/src/commons/dropdown/DropdownCourses.tsx
@@ -1,10 +1,13 @@
 import { Dialog, DialogBody, HTMLSelect } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useDispatch } from 'react-redux';
 import { useNavigate } from 'react-router';
 
+import SessionActions from '../application/actions/SessionActions';
 import { Role } from '../application/ApplicationTypes';
 import { UserCourse } from '../application/types/SessionTypes';
+import { useTypedSelector } from '../utils/Hooks';
 
 type Props = {
   isOpen: boolean;
@@ -15,6 +18,8 @@ type Props = {
 
 const DropdownCourses: React.FC<Props> = ({ isOpen, onClose, courses, courseId }) => {
   const navigate = useNavigate();
+  const dispatch = useDispatch();
+  const latestCourse = useTypedSelector(state => state.session.courseId);
 
   const options = courses.map(course => ({
     value: course.courseId,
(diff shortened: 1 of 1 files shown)


C-likethis123:
Hi @RichDom2185, can you review this again? Thank you!

src/commons/dropdown/DropdownCourses.tsx
```diff
@@ -15,18 +18,26 @@ type Props = {
...
   const options = courses.map(course => ({
     value: course.courseId,
     label: course.courseName.concat(!course.viewable ? ' - disabled' : ''),
     disabled: !course.viewable && course.role !== Role.Admin
   }));
 
-  const onChangeHandler = (e: React.ChangeEvent<HTMLSelectElement>) => {
-    navigate(`/courses/${e.currentTarget.value}`);
+  const onChangeHandler = async (e: React.ChangeEvent<HTMLSelectElement>) => {
+    await dispatch(SessionActions.updateLatestViewedCourse(Number(e.currentTarget.value)));
```
RichDom2185, line 31: image

I don't think you can await a `dispatch` call? We should be handling the async logic via sagas instead.

What was the idea behind this change?

C-likethis123: Hi @RichDom2185, thank you for the feedback and sorry for the late reply.

I agree that the logic is odd - here's my reasoning behind the change:
1. when switching between courses, the router uses `isGameEnabled` from the state of the current course to decide whether to route to the game page, or to route to another page.
2. this `isGameEnabled` state is based on the current course, not the course that the user wants to navigate to.
3. the state is only updated via an API call after the routing logic kicks in and the user finishes navigation to a page.
4. the idea behind awaiting the dispatch call is to update the state of the latest viewed course to the course that I am trying to navigate to, before the routing logic kicks in

I can try to explore handling the async logic via sagas, but another suggestion from me is to enforce a navigation to a page that is common among all courses (eg the home page) so we don't have to use this conditional logic.

PR #3084, opened 2025-02-04
integrate conductor into frontend

tsammeow:
### Description

integrated [conductor](https://github.com/tsammeow/sa-conductor) system for a standard language implementation interface

implemented basic feature flag system

new conductor execution system is hidden behind the `conductor.enable` flag.

also upgraded yarn (4.6.0) to enable prepack script to be run automatically when installing raw dependencies

### Type of change

### How to test

old behaviour was left untouched; no changes to tests.


RichDom2185:
#3091 will fix the CI issues (hopefully)

package.json
```diff
@@ -163,9 +165,6 @@
     "url": "^0.11.1",
     "webpack-bundle-analyzer": "^4.9.0"
   },
-  "resolutions": {
-    "**/gl": "^8.0.2"
-  },
```
RichDom2185, line 168: We still need this, right? Just that for Yarn 4, we need to change the format and remove the `**/`:

```
"resolutions": {
  "gl": "^8.0.2"
},
```

RichDom2185 (changes requested):
Can also change `ci.yml`, and `build-development.yml` to use `yarn install --immutable` instead of `yarn install --frozen-lockfile`?

package.json
tsammeow: i didn't touch this, it seems yarn removed that automatically (probably because gl isn't a dependency)

RichDom2185: > i didn't touch this, it seems yarn removed that automatically (probably because gl isn't a dependency), but i'll check again

Yes it's because yarn 4 doesn't understand the `**/gl` format so it think it's not a dependency. You need to manually change it to `gl` so that Yarn will pick it up as a transitive dependecy

RichDom2185: The lockfile isn't the issue, it's package.json

tsammeow: hm, can we try this first? if we actually need 8.0.2 then it should be added as a dependency instead of forcing a resolution like this

tsammeow:
hold on, there's a small bug with the repl eval button, will fix in a bit

tsammeow:
eval bug is fixed and there should be no change in behaviour when conductor.enable is off

there's a ui bug in conductor mode where running something in the repl while the runner isn't running will make it look like the runner is running, even though it will not be started (this is due to isRunning being set to true in replReducer/evalRepl - i cannot check the feature flag in this reducer to switch behaviour off)

martin-henz (approved):
Reviewed with Sam. Looks good!

RichDom2185 (approved):
Implementation wise, LGTM, thanks!

But please try to follow the code style and conventions (e.g. import style, variable names, file locations) next time (or even in a follow-up PR)

PR #3105, opened 2025-02-28
Game: Allow consistent, automatic switching between Game and Assessments

JustATin555:
### Description

- Cadets are able to submit minigames from within the Assessment Workspace and are then prompted to close the window and return to game
- userstate.assessments.id and navigateToAssessment() use the `number` instead of `id`
- An assessment type can be marked as a minigame from the Admin Panel

### Type of change

### How to test

- Passed all testcases
- Properly loads assessments in both Assessment Overview and Assessments Workspace
- Can load assessments from GroundControl

### Checklist

- [x] I have tested this code


src/features/game/scenes/gameManager/GameGlobalAPI.ts
```diff
@@ -450,15 +452,23 @@ class GameGlobalAPI {
...
-      window.open(`/courses/${courseId()}/missions/${assessmentId}/0`, 'blank');
-      window.focus();
+      const assessments = await getAssessmentOverviews(
+        SourceAcademyGame.getInstance().getAccountInfo()
+      );
+      if (assessments) {
+        const { id, type } = assessments.filter(
+          assessment => assessment.number === assessmentId
+        )[0];
+        window.open(`/courses/${courseId()}/${type.toLowerCase()}/${id}/0`, 'blank');
```
RichDom2185, line 469: We open in a new window?

JustATin555: That's the current implementation. We have considered changing it but decided to leave it for a separate PR if time permits.

src/features/game/state/GameUserStateManager.ts
```diff
@@ -55,7 +55,7 @@ export default class GameUserStateManager {
       (assessments || [])
         .filter(assessment => assessment.status === 'submitted')
         .sort((a, b) => (a.closeAt <= b.closeAt ? -1 : 1))
-        .map(assessment => assessment.id.toString())
+        .map(assessment => assessment.number || assessment.id.toString())
```
lhw-1, line 58: Why is this an `||`? Are there cases where it does not have a valid `assessment.id.toString()` but has a valid `assessment.number`?

lhw-1 (commented):
There does not seem to be major issues with the PR. @JustATin555 Are you able to include screenshots of UI changes?

Also, once the PR is merged, we'll need to update the game dev guide and the storywriter guide for relevant portions on the id vs number.

src/features/game/state/GameUserStateManager.ts
JustATin555: The other way round. By rights all story writers should include a unique "number" identifier for each chapter, but if they forget this provides a fallback id based on backend id.

src/features/game/scenes/gameManager/GameGlobalAPI.ts
JustATin555: Confirmed with prof that we will stick to this implementation for now.

src/features/game/state/GameUserStateManager.ts
JustATin555: With regards to updating the game dev guide and storywriter guide, both currently do not contain documentation for `navigate_to_assessment()`

lhw-1: > With regards to updating the game dev guide and storywriter guide, both currently do not contain documentation for `navigate_to_assessment()`

That is not ideal... could we take this opportunity to add it into the documentation with the updated implementation?

lhw-1: My mistake - looks like we've updated the story chapters after 21/22 to eliminate the need to update assessment IDs. 

In that case, this is a breaking change by technicality but there shouldn't be any changes needed on the chapters. 

Still, the documentation will be good to have - I think we can now start using `navigate_to_assessment()` again in our chapters :)

JustATin555:
image
Direct submission button shown for minigames.

JustATin555:
image
Message shown after submission.

lhw-1:
> Message shown after submission.

Hmm... I wonder if the window might be a bit too different from the current assessment windows? (IIRC there is no window for submission on the workspace itself, but there is a window for submission from assessment list that looks a bit different)

@RichDom2185 Any thoughts on standardizing styles for popup windows? I am ok with this for now since this will only be the case for the minigame, though I suspect that in the long run we want to minimize edge cases like this..

JustATin555:
@lhw-1 the window is a modified copy of the assessment briefing card. Will likely refactor into a more standardised React component if time permits later on.

lhw-1:
My main concern is that assessments opened by the minigame are functioning differently from the regular assessments when I'm not sure if there is a need for that - can I check if the plan is to show these minigame assessments via the assessment list, or if they are only accessible through the game chapters?

JustATin555:
Should only be accessible through the game chapters - the category would have "display in dashboard" disabled.

lhw-1:
> Should only be accessible through the game chapters

Alright, in that case we can go ahead with it. We can revisit the standardization at a later date as needed.

RichDom2185 (approved):
LGTM, thanks!

PR #3107, opened 2025-03-04
bumping js-slang

martin-henz:
test cases need to be fixed


RichDom2185:
js-slang has retirement chances, need to update dependencies as well as  the CI workflows as well

PR #3108, opened 2025-03-04
bumping js-slang

martin-henz:
test cases need fixing: lazy and gpu variants are removed in the newest js-slang

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "i18next-browser-languagedetector": "^8.0.0",
     "java-slang": "^1.0.13",
     "js-cookie": "^3.0.5",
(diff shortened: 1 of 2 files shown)


RichDom2185 (changes requested):
need to update other dependencies and CI workflow as well due to js-slang changes

sayomaki (approved):
LGTM, should work after #3109 is merged.

RichDom2185:
> LGTM, should work after #3109 is merged.

Let's remove gpu JS from CI since JS-slang doesn't need it anymore

RichDom2185:
@martin-henz hi Prof, actually it was still not handled yet...

PR #3116, opened 2025-03-23
Refactor and fix group filter

GabrielCWT:
- Refactored `showAllGroups` to `showUserGroups` to follow the params being passed to the backend.
- Fixed the initial fetch not being filtered by group leading to long initial load time.


RichDom2185 (approved):
LGTM, thanks!

PR #3118, opened 2025-03-25
added continue instr

alsonleej:
resolves https://github.com/source-academy/js-slang/issues/1753 by adding continue instr support in frontend

image


sayomaki (approved):
LGTM

PR #3119, opened 2025-03-31
CSE Machine: Visual garbage collector

alsonleej:
### Description

Currently, frames that will no longer be referenced still show up as white in the environment, along with their associated bound objects. There was previously no mechanism to visually "garbage collect" frames which should have been dereferenced. This PR fixes this, by having the default color be faded, and for each step in the CSE Machine, we mark the current environment as _reachable_, and then recursively mark its parent frames (with their bindings) as reachable, up till the root parent (Global).

### Type of change

### How to test
`// frames being abandoned
// let x = 0;
// let y = 10;
// while (x < 1) {
//     x = x + 1;
//     {
//         let y = 20;
//         continue;
//     }
// }
// pair(x, y);
// 3;

// frames being abandoned 2
// let x = 3;
// let i = 4;
// for (let i = 0; i < 30;  i = i + 1) {
//     x = x + i;
// }
// const k = [1,2,3];

// function closures dereferncing
// function a(b,c) {
//     function f(c) {
//         return 3*c;
//     }
    
//     return f(b)+c;
// }

// function j(b,c) {
//     function f(c) {
//         return 3*c;
//     }
    
//     return f(b)+c;
// }

// a(5,2);
// j(6,2);

//test array referencing
// const f = x => 1;
// {
// const a =[1,2,f];
// }
// f(3);
`
## Demo

Frames and bindings (such as function closures) will indicate as grey, once they will be no longer accessed.

https://github.com/user-attachments/assets/242e5436-8848-48d2-8ba9-2bd472c2e364

Nested referencing within ArrayUnits are also updated.

https://github.com/user-attachments/assets/b079a7b9-8e14-426c-a69b-18116c2867c1

### Checklist

- [ ] I have tested this code

### DIFF
--- a/src/features/cseMachine/CseMachineAnimation.tsx
+++ b/src/features/cseMachine/CseMachineAnimation.tsx
@@ -152,6 +152,7 @@ export class CseAnimation {
     ) {
       return;
     }
+
     if (isNode(lastControlItem)) {
       CseAnimation.handleNode(lastControlItem);
     } else if (isInstr(lastControlItem)) {
--- a/src/features/cseMachine/CseMachineUtils.ts
+++ b/src/features/cseMachine/CseMachineUtils.ts
@@ -412,13 +412,13 @@ export function setUnhoveredStyle(target: Node | Group, unhoveredAttrs: any = {}
     node.setAttrs({
       stroke: node.attrs.stroke
         ? node instanceof Text
-          ? defaultTextColor()
-          : defaultStrokeColor()
+          ? reachedTextColor()
+          : reachedStrokeColor()
         : node.attrs.stroke,
       fill: node.attrs.fill
         ? node instanceof Text
-          ? defaultTextColor()
-          : defaultStrokeColor()
+          ? reachedTextColor()
+          : reachedStrokeColor()
         : node.attrs.fill,
       ...unhoveredAttrs
     });
@@ -942,16 +942,16 @@ export const isStashItemInDanger = (stashIndex: number): boolean => {
 export const defaultBackgroundColor = () =>
   CseMachine.getPrintableMode() ? Config.PrintBgColor : Config.BgColor;
 
-export const defaultTextColor = () =>
+export const reachedTextColor = () =>
   CseMachine.getPrintableMode() ? Config.PrintTextColor : Config.TextColor;
 
-export const fadedTextColor = () =>
+export const defaultTextColor = () =>
   CseMachine.getPrintableMode() ? Config.PrintTextColorFaded : Config.TextColorFaded;
 
-export const defaultStrokeColor = () =>
+export const reachedStrokeColor = () =>
   CseMachine.getPrintableMode() ? Config.PrintStrokeColor : Config.StrokeColor;
 
-export const fadedStrokeColor = () =>
+export const defaultStrokeColor = () =>
   CseMachine.getPrintableMode() ? Config.PrintStrokeColorFaded : Config.StrokeColorFaded;
 
 export const defaultActiveColor = () =>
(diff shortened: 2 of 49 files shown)


arnav-goel10:
I noticed some cases where the **reachability logic doesn’t fully reflect actual references**:

### 1. Indirect referencing through nested arrays

```js
function test_func() {
    const x = [];
    const y = [x];
    const a = [x, y];
    return a;
}

const result = test_func();
result[1][0] = 5;
```

In this case, `x`, `y` and `a` are still indirectly accessible via the returned `result` array, `result[0]` gives `x`, and `result[1][0]` also accesses `x`. However, the visualization currently marks them as unreachable (faded out), even though they are still referenced.


https://github.com/user-attachments/assets/98cda54b-aad6-4e0c-9abd-441fc3fea23f

---

### 2. Function still referenced after return

```js
function test_func() {
    const x = [];
    const a = d => d + 1;
    return a;
}

const result = test_func();
result(5);
```

In this example, `test_func()` returns a function `a`, which is saved in the variable `result`. Even though `test_func` finishes running, the function `a` still exists and can be used as it hasn’t been lost or deleted. This means the frame where `a` was created is **still needed** and should not be marked as unreachable.

Right now, the visualization **marks the frame as unreachable as soon as `test_func()` returns**, even though `a` is still being held in `result`. This gives the wrong impression, it looks like the frame has been garbage collected, even though the function is still accessible and working.

Also, when `result(5)` is called, the visualization updates and **marks the frame as reachable again**. But this shouldn't happen. Once a frame is marked as unreachable (like garbage collected), it shouldn't become reachable again.


https://github.com/user-attachments/assets/729d138d-20d0-4b45-86ec-63d8b7e9c124

alsonleej:
unfortunately, after 12 consecutive hours, i still can't figure out why even when isReachable is true, the object does not change its color. debugging is pain.

possible ideas for future:

1) each object to inherit a setReachable() method, and calls each other in a chain. Frame to call setReachable in bindings (key and values) and arrows (to parent frames). ArrayValue calls setReachable for each of its ArrayUnits (key and value (which may be references)). FnValue to setReachable on its ArrrawFromFn and enclosing frame (only if enclosing frame is not already reachable, to avoid an infinite loop).

Unfortunately changing the color of the enclosing frame for some reason doesn't work no matter what I do, yet everything else updates its color by setting isReachable. (also a word of warning: i cannot set the target of an arrow to be reachable, as the target remains undefined until countingSteps or later. console.log() did not help me, turns out it givesobject references instead of object state. i wasted way too long to find out.  thanks google.)


2) perhaps there could be some pre-processing, where before any drawing, we already do 3x DFS (from current environment, from control env instruction, and from pair/array and closure from stash) to mark all reachable objects. Then when the actual drawing starts, it is colored based on this. I'm not sure where to do this though.

alsonleej:
> ### 2. Function still referenced after return
> ```js
> function test_func() {
>     const x = [];
>     const a = d => d + 1;
>     return a;
> }
> 
> const result = test_func();
> result(5);
> ```
> 
> In this example, `test_func()` returns a function `a`, which is saved in the variable `result`. Even though `test_func` finishes running, the function `a` still exists and can be used as it hasn’t been lost or deleted. This means the frame where `a` was created is **still needed** and should not be marked as unreachable.
> 
> Right now, the visualization **marks the frame as unreachable as soon as `test_func()` returns**, even though `a` is still being held in `result`. This gives the wrong impression, it looks like the frame has been garbage collected, even though the function is still accessible and working.
> 
> Also, when `result(5)` is called, the visualization updates and **marks the frame as reachable again**. But this shouldn't happen. Once a frame is marked as unreachable (like garbage collected), it shouldn't become reachable again.
> 
>  Source.Academy.-.Personal.-.Microsoft.Edge.2025-04-02.18-36-28.mp4

After way too long debugging this, I found that the enclosing environment of the bottom `FnValue` **does not actually refer to the existing frame, but re-creates it.** So when I was setting `setReachable` on the `FnValue`'s enclosing frame, it actually sets a different object.

More details here https://github.com/source-academy/frontend/issues/3127

PR #3123, opened 2025-04-05
Language options

tohlh:
This PR's objective to add language options support to Source Academy. The use case for now is to allow instructors to configure the strictness of Source Typed variants.

This PR allows the assessments workspace to correctly fetch and process the new `Library` schema from the backend.


src/commons/assessment/AssessmentTypes.ts
```diff
@@ -177,6 +177,7 @@ export type Library = {
     2?: string; // For mission control
   }>;
   moduleParams?: any;
+  languageOptions?: any;
```
RichDom2185, line 180: Double check, can we avoid any here?

RichDom2185 (commented):
Just the one comment, thanks!

src/commons/assessment/AssessmentTypes.ts
tohlh: Have fixed this on the latest push. Thanks!

PR #3130, opened 2025-04-24
Remove Concurrent and Interpreter

leeyi45:
Removal of the concurrent variant as well as the legacy interpreter.

I'm not sure if the functionality from `manualToggleDebugger` (and thus `interrupt` or `paused`) is still required within `WorkspaceSaga`


martin-henz (approved):
Looks good. More cleanup possible later, as discussed in telegram "SA 2526 Leadership Charm"

PR #2996, opened 2024-05-12
Migrate to Vite

RichDom2185:
### Description

Supersedes #2327.
Supersedes #2508.

Closes #2933.
Will need to remove CRA compatibility layers after (part of #2990).

TODO:

- [x] Devserver
- [x] Build
- [x] Somehow Admin Panel renders as an iframe that blocks all subsequent interactions (resolved in ce5a1c6517fa7743d9df5d2fb9643b0bb42b1245)
- [ ] GameSimulator crashes with runtime error (fixed by #2997)
- [ ] Test scripts
- [ ] Manually test everything
- [ ] Java-slang not working
- [ ] C-slang not working
- [ ] Service worker

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Code quality improvements

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #3131, opened 2025-04-26
Migrate to rsbuild and swc

### LINKED ISSUE #2933: Migrate to Vite
We really need to move away from CRA because it's really starting to show its age...

Past attempts:

* #2327
* #2508

### LINKED ISSUE #2990: Clean up codebase post-migration to Vite/React19
Includes:

- Lint rules https://github.com/source-academy/frontend/pull/2986#discussion_r1589093567
- Dependencies https://github.com/source-academy/frontend/pull/2986#discussion_r1590365190
- &hellip; _more to be added as we discover them_

RichDom2185:
* Closes #2933.
* Closes #2990.
* Supersedes #2996.

---

### Description

CRA is dead. Switch to rsbuild (which uses rspack &rarr; drop-in for webpack &rarr; powers CRA), but written in Rust so much faster.

* Switch to rsbuild for performance and maintenance
* Upgrade jest from v27 to v29
* Switch to swc parser (also happens to be written in rust lol) for jest for performance
* Remove CRA and CRACO
### Type of change

### How to test

Everything should (and does) still work since it's effectively drop in.

### Checklist

- [x] I have tested this code


package.json
```diff
@@ -20,8 +21,7 @@
     "format:tsx": "prettier --list-different \"src/**/*.{js,jsx,ts,tsx}\"",
     "format:scss": "prettier --list-different --parser scss \"src/**/*.scss\"",
     "format:ci": "yarn run format:tsx && yarn run format:scss",
-    "start": "cross-env DISABLE_ESLINT_PLUGIN=true BROWSER=none PORT=8000 craco start",
```
sayomaki, line 22: Can we consider having `start` alias `dev` for those users still used to the old way of launching? Or do you think that is not necessary?

src/features/remoteExecution/RemoteExecutionEv3Types.ts
```diff
@@ -5,6 +5,34 @@ import SensorGyro from 'src/assets/sGyro.svg';
 import SensorTouch from 'src/assets/sTouch.svg';
 import SensorUltrasonic from 'src/assets/sUltrasonic.svg';
 
+// (18 March 2022)
```
sayomaki: Does this get resolved with SWC? I assume not, that's why you've added it.

Might need to take a deeper look into this.

package.json
```diff
@@ -25,7 +25,8 @@
     "test-coveralls": "./scripts/test-coveralls.sh",
     "update-ui-snapshots": "jest --updateSnapshot",
     "eslint": "eslint src",
-    "prepare": "husky"
+    "prepare": "husky",
+    "preview": "rsbuild preview"
```
sayomaki, line 29: Maybe we can have this as the new `yarn start` command behaviour instead?

src/env.d.ts
sayomaki, line 1: Should this file be named `globals.d.ts` or `imports.d.ts` instead? Since it will be required at quite a high level, `env` doesn't really make too much sense as this handles the import types.

package.json
```diff
@@ -100,6 +100,9 @@
     "sourceror": "^0.8.5",
     "unified": "^11.0.0",
     "uuid": "^11.0.2",
+    "workbox-core": "^7.3.0",
```
sayomaki, line 103: Is this required for anything in particular that we need to have it explicitly as a dependency?

sayomaki (commented):
Did a preliminary quick review of the code changes, although I'm not too sure for the testing updates. As long as they are passing (which they appear to do so), I think it should not be too far off.

I have also yet to run the code locally, so here are just some questions and ideas from me after glancing over the changes, let me know what you think!

src/features/remoteExecution/RemoteExecutionEv3Types.ts
RichDom2185: Yep it didn't get resolved. SWC is only to parse TS/X files, so it won't do anything to these other imports

package.json
RichDom2185: Yeah, needed in the service worker. Previously it was a transitive dependency under CRA so we didn't need to explicitly put it in package.json

package.json
RichDom2185: I think we can just use start instead of dev

src/env.d.ts
RichDom2185: Sure, I'll make the changes later. Thanks!

leeyi45:
Previously with Webpack it would compile away the dynamic import used to import modules (because there was no way to `webpack-ignore` something in `js-slang`). Should I check if `js-slang` can be switched away from using the function constructor, or should we leave things as is?

RichDom2185:
> Previously with Webpack it would compile away the dynamic import used to import modules (because there was no way to `webpack-ignore` something in `js-slang`). Should I check if `js-slang` can be switched away from using the function constructor, or should we leave things as is?

@leeyi45 I think I saw this comment somewhere, but wasn't too sure what it entailed in terms of testing.

I had tried running some module code in the playground and it seems to work, though I'm not sure if that is what you mean.

Would you mind checking as well? Thank you!

leeyi45:
> > Previously with Webpack it would compile away the dynamic import used to import modules (because there was no way to `webpack-ignore` something in `js-slang`). Should I check if `js-slang` can be switched away from using the function constructor, or should we leave things as is?
> 
> @leeyi45 I think I saw this comment somewhere, but wasn't too sure what it entailed in terms of testing.
> 
> I had tried running some module code in the playground and it seems to work, though I'm not sure if that is what you mean.
> 
> Would you mind checking as well? Thank you!

Did a quick review. I think on the `js-slang` side we should move toward building for ESM rather than CommonJS, which would make this change easier. Either ways, I've put the ignore warnings config back in for the dynamic module import

rsbuild.config.ts
```diff
@@ -0,0 +1,144 @@
...
+        cwd: __dirname,
+        configType: 'flat'
+      }
+    })
+  ],
+  server: {
+    port: 8000
+  },
+  tools: {
+    // TODO: See if still needed
```
heyzec, line 28: Will we need these commented code in the future?

RichDom2185: It's mainly kept as reference since the original file was deleted (so it would be hard to get a "file history" since the old webpack config won't be present in the current repo anymore).

Now that this PR is merged, it's free to be deleted as one can always look at the file history of rsbuild.config.ts to easily see what the original webpack config was

PR #3201, opened 2025-06-27
feat: OAuth flow for VS Code 

heyzec:
### Description

VSC PR: https://github.com/source-academy/vscode/pull/47

To support the "Log in with NUSNET ID" option.

Unlike the SAML flow in https://github.com/source-academy/backend/pull/1240, the changes needed to support this requires more code change on the frontend and none on the backend.

The OAuth 2.0 standard mandates single use for authorization codes (https://stackoverflow.com/a/26564117). Hence, we do not need to make use of the exchange table created in the earlier backend PR for SAML.

### Modified login flow for VSC
1. User clicks any of the provider buttons on the login page within the Webview of VSC. This triggers the opening of browser window to the auth code link (NUS VAFS).
    - Change is made here to `redirect_uri` = `/login/vscode_callback` instead of the usual `/login/callback`. (We need a way to tell frontend to act differently later on)

2. Normal OAuth flow
    - User authorises application
    - Auth code received via callback

3. Handling auth code
    - An additional step is added here. Instead of immediately contacting `/login` of the backend, the frontend uses VS Code deeplink to send `code` and `client-request-id` to the extension.
      image
    - VSC forwards these again by telling embedded frontend to visit `/login/vscode_callback` with the relevant query params. From here on, we resume the typical flow of contacting backend `/login`.

### Type of change

### How to test

 

Tested by pushing (a modified variant) of this branch to deploy-stg.

### Checklist

- [x] I have tested this code


src/commons/sagas/LoginSaga.ts
```diff
@@ -9,12 +10,17 @@ import { showWarningMessage } from '../utils/notifications/NotificationsHelper';
 
 const LoginSaga = combineSagaHandlers({
   [SessionActions.login.type]: function* ({ payload: providerId }) {
-    const epUrl = computeEndpointUrl(providerId);
+    const isVscode = yield select(state => state.vscode.isVscode);
```
heyzec, line 13:
```suggestion
    const isVscode = yield select(state => state.vscode?.isVscode);
```

All the tests failures would have been simply gone if I were to do this... but I'm not sure about this

RichDom2185: Why is the slice typed as optional?

heyzec: Some test failures are caused by their mock states not including this vscode slice.

src/pages/login/LoginVscodeCallback.tsx
```diff
@@ -13,10 +27,36 @@ const LoginVscodeCallback: React.FC = () => {
...
+    'client-request-id': clientRequestId
+  } = parseQuery(location.search);
+  const isVscode = useTypedSelector(state => state.vscode.isVscode);
   const { access_token: accessToken, refresh_token: refreshToken } = parseQuery(location.search);
 
+  // `code` parameter from OAuth2 redirect, `ticket` from CAS redirect (CAS untested for VS Code)
+  const authCode = code || ticket;
+
+  const launchVscode = () => {
+    window.location.href = `vscode://source-academy.source-academy/sso?code=${authCode}&client-request-id=${clientRequestId}`;
```
RichDom2185: Understand that this is copied from the other login callback, nbut we won't ever use SAML here right? So ticket etc. is not needed.

src/pages/login/LoginVscodeCallback.tsx
```diff
@@ -27,10 +67,9 @@ const LoginVscodeCallback: React.FC = () => {
       dispatch(SessionActions.fetchUserAndCourse());
       navigate('/welcome');
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [isVscode]);
```
RichDom2185: Can you double check this? Seems like the eslint-disable is still needed?

heyzec: ~~Doesn't seem to raise any errors on my side, neither on the pipelines too~~

I wasn't looking in the right place. Added eslint-disable.

src/pages/login/LoginVscodeCallback.tsx
heyzec: In the modified flow for SAML, the last step involves redirecting the user to this endpoint.

I'm not sure how a modified CAS flow would be like, so I can remove `ticket` for now.

src/pages/login/LoginVscodeCallback.tsx
```diff
@@ -13,10 +27,32 @@ const LoginVscodeCallback: React.FC = () => {
...
+  const {
+    code,
+    provider: providerId,
+    'client-request-id': clientRequestId
+  } = parseQuery(location.search);
+  const isVscode = useTypedSelector(state => state.vscode.isVscode);
   const { access_token: accessToken, refresh_token: refreshToken } = parseQuery(location.search);
 
+  const launchVscode = () => {
+    window.location.href = `vscode://source-academy.source-academy/sso?code=${code}&client-request-id=${clientRequestId}`;
```
RichDom2185: Nit: Can use a value in `Constants.ts` for the base URL instead of being hard coded? But since unlike a frontend base URL, the extension is technically universal (deployment-agnostic), should be fine too. Just DRY if you need to do this in the future

RichDom2185 (approved):
LGTM, thanks!

src/pages/login/LoginVscodeCallback.tsx
heyzec: should it be a environment variable?

RichDom2185: I think no need? Do you foresee it to be configurable? There's only ever one extension (even when developing locally) right?

heyzec: Makes sense

src/pages/login/LoginVscodeCallback.tsx
```diff
@@ -41,6 +78,30 @@ const LoginVscodeCallback: React.FC = () => {
...
       </Card>
     </div>
+  ) : (
+    <div className={classNames(classes['Login'], Classes.DARK)}>
+      <Card elevation={Elevation.FOUR}>
+        <div>
+          <div className={classes['login-header']}>
+            <H4>
+              <Icon className={classes['login-icon']} icon={IconNames.LOG_IN} />
+              Sign in with SSO
```
heyzec, line 89: Do you know why this page is only shown when the user is not logged in? Otherwise, the frontend is redirected to `/` immediately. I can't figure this out

PR #3250, opened 2025-07-15
Use new test environment format for Jest

RichDom2185:
### Description

Actually, this is the expected way to define environments in the current Jest major version. But I think there might be some backwards compatibility of sorts, thus it still ran on Linux CI as well as my Windows. But it wouldn't on my Mac.

Now it does.

### Type of change

### Checklist

- [x] I have tested this code


sayomaki (approved):
LGTM, thanks!

PR #3113, opened 2025-03-08
Leaderboard

tohyzhong:
### Description
- Added Overall Leaderboard and Contest Leaderboards
- XP and contest score fetching from backend
- podium display for each leaderboard
- customise number of entries to display in overall leaderboard in course config panel and create courses
- dropdown to switch between leaderboards
- routing to different leaderboards (url + dropdown)
- links for contest entries
- added contest scores update button in ground control (under voting features)
- export as CSV for admins and staff
- contest score updating
- contest xp issuing for contest winners
- Fixed code links from leaderboard without using setTimeout
- Customise XP issuing for contest winners in contest voting XML file
- Default Source Academy Profile pictures in Leaderboard based on NUSNet ID

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [ ] Code quality improvements

### Checklist

- [ ] I have tested this code
- [ ] I have updated the documentation

PR #3274, opened 2025-08-07
Leaderboard refactor

sayomaki:
### Description

- Updated endpoints and requests to match updated backend routes
- Directly call requests for exporting of leaderboard instead of using store
- Remove toggles and fields for leaderboard during course creation
- Only show leaderboard in contest assessment workspace after assessment is published
- Combined contest leaderboard may not work as expected due to breaking changes (need to disable contest leaderboard for now)

This PR is an improvement on #3113

### Type of change

### How to test

Requires backend PR: https://github.com/source-academy/backend/pull/1281

### Possible planned improvements (not in PR)
- Consider feasibility of keeping overall contest leaderboards, improve design if keeping
- Improve achievements system and incorporate achievements with leaderboards
- Combined contest leaderboards should be retrieved directly from assessment instead of additional endpoint
- QoL improvements to voting features and token counter


src/pages/leaderboard/subcomponents/LeaderboardExportButton.tsx
```diff
@@ -15,60 +18,50 @@ type Props = {
...
-    if (type == 'overall') dispatch(LeaderboardActions.getAllUsersXp());
-    else if (type == 'score')
-      dispatch(LeaderboardActions.getAllContestScores(contestID as number, visibleEntries));
-    else
-      dispatch(LeaderboardActions.getAllContestPopularVotes(contestID as number, visibleEntries));
-    setExportRequest(true);
-  };
+  const onExportClick = async () => {
+    const accessToken = store.getState().session.accessToken!;
+    const refreshToken = store.getState().session.refreshToken!;
```
RichDom2185: Why not destructure from Redux `useTypedSelector` earlier?

src/pages/leaderboard/subcomponents/LeaderboardExportButton.tsx
```diff
@@ -15,60 +18,50 @@ type Props = {
...
-        return (store: { leaderboard: { contestPopularVote: any } }) =>
-          store.leaderboard.contestPopularVote;
-    }
-  };
-
-  const selector = useMemo(() => selectData(type), [type]);
-  const data = useTypedSelector(selector);
-
-  const visibleEntries = Number.MAX_SAFE_INTEGER;
+  const role = useTypedSelector(store => store.session.role);
```
RichDom2185: Can use `useSession`.

RichDom2185 (changes requested):
Just some minor nits :)

src/pages/leaderboard/subcomponents/LeaderboardExportButton.tsx
sayomaki: It's a one-time operation which would unlikely need keeping track of the tokens as state, but if you think its better we can change that.

src/pages/leaderboard/subcomponents/LeaderboardExportButton.tsx
sayomaki: For this, the primary concern is still that using `useSession` would cause the hook to change when any session value changes (but here we are only interested in role). 

However, we could combine this together with the previous change in that case for better readability.

RichDom2185: Fair concern, but I think as long as we don't have side effects, the performance difference would be negligible, react and modern browsers are performant enough that we can afford to focus on readability?

sayomaki: Changes have been added.

PR #3278, opened 2025-08-08
Fix stepper errornously evaluating non-stepper output

### LINKED ISSUE #3279: Source §2: list program crashes frontend
```
pair(list(1), 2);
```
currently crashes sourceacademy.org, see https://share.sourceacademy.org/caxir

Note that the stepper now handles this program correctly.  The problem might be type-related; console shows:

```
96669.eebb44cf.js:2897 Error handled by React Router default ErrorBoundary: TypeError: Cannot read properties of undefined (reading 'type')
    at e (index.d601d906.js:438:6400)
    at index.d601d906.js:438:6925
    at ec (index.d601d906.js:438:7209)
    at ad (lib-react.917565be.js:2:58521)
    at i (lib-react.917565be.js:2:119389)
    at oO (lib-react.917565be.js:2:99066)
    at oF (lib-react.917565be.js:2:98936)
    at oC (lib-react.917565be.js:2:96041)
    at r8 (lib-react.917565be.js:2:44927)
    at lib-react.917565be.js:2:93621
```
Image

sayomaki:
### Description
It appears that the check for ensuring whether the evaluation output is for the stepper is missing (presumably removed as it is no longer in use), and the check falsely assumes an array with first element being an object (which unfortunately, arrays are objects as well) is stepper output, and then tries to render the stepper with invalid information

Some examples include:
`pair(pair(1, 2), 3);`
`list(list(1), 2);`
`list(pair(1, 2), 3);`
`pair(list(1), 2);`

Resolves #3279.

### Type of change

### Checklist

- [x] I have tested this code


RichDom2185 (approved):
LGTM, thanks

PR #3304, opened 2025-08-14
Fix incorrect assessment overview memoization

RichDom2185:
### Description

Wow, this bug is 4 years old.

Indices should not be used as memoization keys. When switching between assessment types, the old image is still shown while the new one loads. Thus if the image load fails, nothing happens.

This fixes it, ensuring the entire assessment card is tied to the assessment ID, thus switching immediately updates the UI to loading state.

Also moves components to use FC instead of function calls as part of the fix (so that `key` prop works without drilling).

### Type of change

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/commons/assessment/Assessment.tsx
+++ b/src/commons/assessment/Assessment.tsx
@@ -1,15 +1,9 @@
 import {
   Button,
-  Card,
   Collapse,
   Dialog,
   DialogBody,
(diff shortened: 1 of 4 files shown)


sayomaki (approved):
LGTM, thanks! It seems that coveralls is working again.

PR #3316, opened 2025-08-18
Add Feature flag for language-directory

loyaltypollution:
Simple proof-of-concept for adding additional languages via a `language-directory`.
* Wrote as feature flag to avoid breaking changes.

To-Do? 
* Add UI support to allow users to view language directory URLs in frontend.

Questions:

* Is SourceAcademy planning to make *conductors* a “first-class” concept?
  - removing the hardcoding of languages in drop down?
  - ensuring that all `-slang` variants support conductors?

Happy to collaborate with relevant people on making conductors first-class if that’s the direction we are taking 👍


s-kybound:
>Should that be done in a config file at build time, or should this also come from a github page?
Any proposals?

@martin-henz best to be done in a config file i think, less layers of indirection especially when setting up new frontends.

package.json
```diff
@@ -41,6 +41,7 @@
     "@reduxjs/toolkit": "^1.9.7",
     "@sentry/react": "^10.5.0",
     "@sourceacademy/c-slang": "^1.0.21",
+    "@sourceacademy/language-directory": "https://github.com/loyaltypollution/language-directory.git#866cc934372a19e514af0025947ce7f1f5a6e1fc",
```
s-kybound, line 44: is it necessary for the language directory to point to your version?

src/commons/controlBar/ControlBarChapterSelect.tsx
```diff
@@ -1,22 +1,17 @@
-import { Button, Menu, MenuItem, Tooltip } from '@blueprintjs/core';
+import { Button, Menu, MenuItem } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import { ItemListRenderer, ItemRenderer, Select } from '@blueprintjs/select';
 import { Chapter, Variant } from 'js-slang/dist/types';
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useDispatch } from 'react-redux';
 
-import {
```
s-kybound, line 7: not a review comment - this is beautiful. thanks!

src/commons/controlBar/ControlBarChapterSelect.tsx
```diff
@@ -31,81 +26,77 @@ type StateProps = {
...
+
+  const onSelectEvaluator = (evaluator: IEvaluatorDefinition) => {
+    dispatch(LanguageDirectoryActions.setSelectedEvaluator(evaluator.id));
+  };
 
   return (
-    <ChapterSelectComponent
-      items={choices.filter(({ mainLanguage }) => mainLanguage === selectedLang)}
-      onItemSelect={handleChapterSelect}
-      itemRenderer={chapterRenderer(isFolderModeEnabled)}
```
s-kybound, line 98: is it right to say that folder mode will be disabled for the new evaluators for now? if so please doc it as a limitation

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
```diff
@@ -1,49 +1,49 @@
...
-  SUPPORTED_LANGUAGES,
-  SupportedLanguage
-} from 'src/commons/application/ApplicationTypes';
+import { useFeature } from 'src/commons/featureFlags/useFeature';
 import SimpleDropdown from 'src/commons/SimpleDropdown';
 import { useTypedSelector } from 'src/commons/utils/Hooks';
-import WorkspaceActions from 'src/commons/workspace/WorkspaceActions';
-import { playgroundConfigLanguage } from 'src/features/playground/PlaygroundActions';
-
-// TODO: Hardcoded to use the first sublanguage for each language
```
s-kybound, line 20: again, beautiful

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
```diff
@@ -1,49 +1,49 @@
...
-  [SupportedLanguage.PYTHON]: pyLanguages[0],
-  [SupportedLanguage.SCHEME]: schemeLanguages[0],
-  [SupportedLanguage.JAVA]: javaLanguages[0],
-  [SupportedLanguage.C]: cLanguages[0]
-};
+import { flagLanguageDirectoryEnable } from 'src/features/languageDirectory/flagLanguageDirectory';
+import LanguageDirectoryActions from 'src/features/languageDirectory/LanguageDirectoryActions';
+import type { ILanguageDefinition } from 'src/features/languageDirectory/LanguageDirectoryTypes';
+
+// Remove when conductors.languageDirectory is default behaviour
```
s-kybound: please mark this explicitly as a TODO, if possible with name and date as well

```javascript
// TODO <name> <date>: Remove when conductors.languageDirectory is default behaviour
```

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
```diff
@@ -1,49 +1,49 @@
...
+import LegacyNavigationBarLangSelectButton from './LegacyNavigationBarLangSelectButton';
+
+function useDirectoryOptions() {
+  const langs = useTypedSelector(s => s.languageDirectory.languages) as ILanguageDefinition[];
```
s-kybound: correct me if i'm wrong, but i'm pretty sure you don't need this type assertation here

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
```diff
@@ -1,49 +1,49 @@
...
+  const dirOptions = useDirectoryOptions();
+  const languagesLoaded = useTypedSelector(s => s.languageDirectory.languages.length > 0);
+  useEffect(() => {
+    if (!languagesLoaded) {
+      dispatch(LanguageDirectoryActions.fetchLanguages());
+    }
+  }, [languagesLoaded, dispatch]);
+
+  const directoryEnabled = useFeature(flagLanguageDirectoryEnable);
+  if (!directoryEnabled) {
```
s-kybound, line 32: nit: this should be moved higher as a guard clause to cleanly separate logic. to my understanding the code above and below this correspond to the new Navigation bar lang select button, but this corresponds to the original button.

s-kybound (changes requested):
minor nits, after resolution LGTM! please make sure the frontend has been tested first.

package.json
dingf3ng: It seems that his version is newer than the version in SA. Might migrate to SA version after merging.

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
dingf3ng: Add explict TODO

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
dingf3ng: Removed.

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
dingf3ng: React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render.eslintreact-hooks/rules-of-hooks.
Seems some linting issues if move it higher. So we have to put it here. Correct me if I am wrong.

src/commons/controlBar/ControlBarChapterSelect.tsx
dingf3ng: The folder mode works for new evaluators

src/commons/navigationBar/subcomponents/NavigationBarLangSelectButton.tsx
s-kybound: got it, thanks!

package.json
s-kybound: will okay this, but need to make sure that the SA version is quickly updated after :sweat_smile:

Issue #3377, opened 2025-09-20
Getting "Error: The context {} is empty, but we are looking for display" for any program


andreihodoroaga:
Trying to run even the most basic program results in this error:

`Error: The context {} is empty, but we are looking for display`

The issue seems to be happening in all programming languages.

loyaltypollution:
Ah, this error is caused by the conductor flag being turned on automatically. The default evaluator is set to pie-slang, so the error message seen is an internal error created by pie-evaluator.js

For now, a quick fix would be to turn off the conductor flag in "Settings -> Feature Flags -> conductor.enable"

I will quickly write a short PR to ensure the flag is not automatically switched on!

PR #3355, opened 2025-09-06
persistent feature flags and language directory support

tsammeow:
### Description

make feature flags persistent across reloads (localstorage) and add support for language directory

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #3378, opened 2025-09-21
Fix languageDirectory feature flag

loyaltypollution:
### Description

The error #3377 is caused by the conductor flag being turned on automatically. The default evaluator is set to pie-slang, so the error message seen is an internal error created by pie-evaluator.js

This solves #3377 by refactoring the feature flag toggle code.


tsammeow:
should be fixed by #3355

PR #3416, opened 2025-11-09
Support sticky sessions configuration

RichDom2185:
### Description

Cookies were being set from the API response but not being sent back in the cross-origin request because of unset `credentials: true` in fetch options, thus sticky sessions/session affinity wasn't working.

This PR adds a support to always force inclusion of such credentials via an environment variable, to keep it configurable on a per-deployment basis.

### Type of change

### Checklist

- [x] I have tested this code


sayomaki (approved):
LGTM, although I'm not sure if this will work. Will need to test with multiple backend instances to confirm.

PR #3570, opened 2026-02-11
CSE Machine Liveness

Akshay-2007-1:
### Summary of change:
This PR implements **liveness analysis and visual highlighting** for the CSE Machine environments and values.

It introduces:
- A liveness pass over the environment tree and heap at each step
- Live flags in `Frame`, `Value`, and `Object` components
- Visual differentiation: live entities render normally, dead entities are greyed out

This effectively functions as a **visual garbage collector**.

---

### Motivation and Context:
Large CSE Machine traces often display many overlapping environments and heap objects, making it difficult for students to determine which bindings are still relevant.

Liveness visualisation:
- Highlights only entities reachable from the current machine state
- Helps learners focus on meaningful state
- Aligns with the CP3108B goal of improving the pedagogical UI for Source program execution

---

### Issues fixed:
This implementation addresses:

- **#2750** – CSE Machine: Ability to toggle between "show all frames created" and "show current frames only"

It also supersedes a previous incomplete attempt:

- **#3119** – CSE Machine: Visual garbage collector

---

Dependencies:
No new external dependencies; changes are confined to src/features/cseMachine (layout, utils, and frame/value components) and reuse existing CSE Machine structures for environments, control, stash, and heap.

---

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [ ] Code quality improvements

---

## How to test
Run code that causes frames and value objects (such as function objects and arrays) to become not live / garbage collected.

Code snippets:
```javascript
let a = 2;
{
    function fun (x) { return () => x;}
    let arr = [fun(1), fun(2), fun(3), 1];
    arr = [arr[0], arr[2]]; 
}
a = 3;
```

https://github.com/user-attachments/assets/9f375481-f9d9-48d0-bc31-508fd7dc2910

The above code highlights the garbage collection of aspect of previously allocated dataArrays which are now useless and should be grayed out!

---

```javascript
let bo = 1;

function make_account(initial_balance) {
 let balance = initial_balance;
 bo = () => initial_balance;
 
 function withdraw(amount) {
  if (balance >= amount) {
   balance = balance - amount;
   return balance;
  } else {
   display("bo");
   return balance;
  }
 }
 return withdraw;
}

const w1 = make_account(100);
w1(40);
let w2 = make_account(200);
w2(40);
w2 = 1;
[...]

PR #3575, opened 2026-02-12
Revert incorrect changes

RichDom2185:
### Description

Partially reverts #3570.

### Type of change

### Checklist

- [x] I have tested this code


Akshay-2007-1:
I have actually made a branch fork of the changes that have been shown here since there were some files that were actually broken in the earlier "merge" at #3570. In that fork, I ran the following tests
```
yarn run format:ci
yarn run eslint
yarn run tsc
yarn test --bail=1
yarn run build
```

In the above tests, the first three work. But the yarn test --bail=1 causes this error : 
image

In fact, I was getting a similar error in MANY of the earlier commits I have done before Prof Martin merged the PR. This is the reason I had to mess with the files that are not related to my project at all. So, how do we fix that? Or do we not fix it?

There is some good news though, when I try to deploy the current reverted implementation highlighted by #3575, _yarn start_ does not give any weird build errors as it does usually and has the Liveness computation to be working quite well!

PR #3577, opened 2026-02-15
Fixing the auto-saving from location changing

ItsByt:
### Description

Summary of change: 
This PR fixes the issue of the SA Game not saving when changing locations if a location has no actions.

### Type of change

###Issues fixed

This implementation addresses the issue of not saving the game state when we move to a location that does not have any game actions. This causes the player to be reset to the last location that was saved from a game action rather than the last location they were in if they reload the game.

### Motivation and Context:
Some locations do not have game actions associated with them, so if a student leaves the game while in this location, they may find it confusing and annoying/frustrating to return at a previous location.

### How to test

In the Game Simulator, Simulate Chapter 0 (or any chapter where in the JSON file for the chapter, some location lacks any actions).

(Example of Chapter 0's "room" location that has actions)
image

(Example of Chapter 0's "commonArea" location that does not have any actions)
image

Once the chapter starts, use the developer tool and track the game states under network after moving between locations:

https://github.com/user-attachments/assets/f78d31b4-189a-4c54-ae23-ad46553d70be

### Check
- [x] I have tested this code

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "react-dom": "^18.3.1",
     "react-drag-drop-files": "^3.0.0",
     "react-draggable": "^4.4.5",
-    "react-dropzone": "^14.2.3",
+    "react-dropzone": "^15.0.0",
     "react-i18next": "^16.0.0",
     "react-konva": "^18.2.10",
     "react-latex-next": "^3.0.0",
--- a/src/commons/sagas/WorkspaceSaga/helpers/updateInspector.ts
+++ b/src/commons/sagas/WorkspaceSaga/helpers/updateInspector.ts
@@ -14,6 +14,10 @@ export function* updateInspector(workspaceLocation: WorkspaceLocation): SagaIter
       state.workspaces[workspaceLocation].lastDebuggerResult,
       state.workspaces[workspaceLocation].context.chapter
     ]);
+    if (!lastDebuggerResult) {
+      yield put(actions.setEditorHighlightedLines(workspaceLocation, 0, []));
+      return;
+    }
     if (chapter === Chapter.FULL_JAVA) {
       const controlItem = lastDebuggerResult.context.control.peek();
       let start = -1;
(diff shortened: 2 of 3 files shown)


src/features/game/scenes/gameManager/GameManager.ts
```diff
@@ -264,6 +272,14 @@ class GameManager extends Phaser.Scene {
...
     // Update state after location is fully rendered, location has been visited
     this.getStateManager().triggerInteraction(locationId);
+
+    // Triggers a location Save at the end of every location change,
+    // except for locations where we have saved from running Location Actions
+    if (this.actionJustSaved) {
+      return;
+    }
+
+    await GameGlobalAPI.getInstance().saveGame();
```
RichDom2185, line 282: We are resetting the `actionJustSaved` before saving, so the `if` clause will always be false?


```suggestion
// Transition to the new location
await blackFade(this, 300, 500, async () => {
await this.getLayerManager().clearAllLayers();
await this.renderLocation(locationId, startAction);
});

// Update state after location is fully rendered, location has been visited
this.getStateManager().triggerInteraction(locationId);

    // Triggers a location Save at the end of every location change,
    // except for locations where we have saved from running Location Actions
    if (this.actionJustSaved) {
      return;
    }
    await GameGlobalAPI.getInstance().saveGame();
    // Reset the actionJustSaved flag
    this.actionJustSaved = false;
```

rsbuild.config.ts
```diff
@@ -23,7 +23,7 @@ export default defineConfig({
     })
   ],
   server: {
-    port: 8000
+    port: 8080
```
RichDom2185: Can we undo this change?


```suggestion
    port: 8000
```

ItsByt: Yup ok

src/features/game/scenes/gameManager/GameManager.ts
ItsByt: The if clause will not always be false.

TLDR:
We first reset the flag to be false.
renderLocation then calls processGameActions, which then changes the flag to be true if actions have been processed


The previous problem was that if a location had no actions, we would not save at all. 
The actionJustSaved flag will only be updated if (an array of) actions for a location is run when we change locations.

So in short, the clause will be TRUE if the location has actions, and this is to prevent double saving since running the location actions triggers a save through the processGameActions method in frontend/src/features/game/action/GameActionManager.ts, which is called when we call this.renderLocation
In this method, we update the flag to be TRUE so the clause is TRUE and we do not do the 2nd save at the end of the changeLocationTo method.

In the case that the location has no actions, the clause will then be FALSE, and thus we will save at the end of the method.

RichDom2185: I see, I didn't realise `processGameActions` is not a pure function. Got it, thanks!

PR #3591, opened 2026-02-19
Asset Missing Issue Fix

chenyuzhen2007-source:
### Motivation and Context
When we are running source academy game locally without deployment, the prefix for the assets of the game will be given wrongly. Every time we test the game, we will have to open the admin panel to load the config to change the prefix. In order to make testers' life easier, this bug is now fixed.

### Summary of the Change
This pull request fixes the asset missing bug while running source academy game on local source academy.
To fix this issue, we added another condition for using the asset prefix. Now the asset prefix will not be used if we are running our local source academy which will have the prefix "courses-dev/2/".

## Type of change

## How to test

Run local source academy and check if the assets are successfully loaded.

## Checklist

- [ ] I have tested this code


chenyuzhen2007-source:
> I don't think we should be hardcoding it, seems like it's a security feature to not give the bucket ID?

I have changed the code, is it better now?

Issue #2750, opened 2024-01-29
CSE Machine: Ability to toggle between "show all frames created" and "show current frames only"


CZX123:
The current CSE Machine implementation of the environment shows all the frames that are created during the execution of the program. While this is in line with how the environment is drawn as part of the CS1101S syllabus, it can become quite messy in situations where a lot of frames are drawn, such as in loops and recursive functions. Many frames are no longer referenced, and it would be nice to have a toggle that removes the display of such frames.

A simple example of this problem would be the following code in Source:
```js
let i = 0;
while (i < 10) {
    const a = 0;
    i = i + 1;
}
```

The output of the CSE Machine during the middle of execution:
![image](https://github.com/source-academy/frontend/assets/13569982/81b5e498-112a-4ea7-8249-7cba0638e582)
In this example, the current frame is very far right, and the user would have to manually move the canvas to find the current frame.

martin-henz:
Alternatively, without creating a new "mode", we could just work on a consistent grey color for frames that are "garbage". There is already a grey color implemented for garbage data structures:

<img width="1070" alt="Image" src="https://github.com/user-attachments/assets/00535a2f-74ea-4827-9ae8-bb0c8f92246a" />

Relevant here: "modeless software" see https://en.wikipedia.org/wiki/Larry_Tesler

PR #3592, opened 2026-02-19
Option for "show all frames created" vs "show current frames only"

Akshay-2007-1:
### Description

A simple change to the existing logic to add an option that can temporarily (for the current step ONLY which resets on changing the step of evaluation on the slider) switch and show ONLY the live frames. This also removes all the dead components on the Layout and fills in the gap by moving the live frames closer! This cleans up the Layout during complex code analysis quite a bit.

We use the pre-existing Liveness logic we built recently here : #3570.

This PR will close the existing open issue here : #2750.

A future PR will be made to work on the animations for these sudden movements to further aid the user!

### Type of change

### How to test
I am attaching a few code snippets along with the result achieved when tested.
```javascrpt
let a = 2;
{
    function fun (x) { return () => x;}
    let arr = [fun(1), fun(2), fun(3), 1];
    arr = [arr[0], arr[2]]; 
}
a = 3;
```
Video Project 3

---
```javascript
function scale_stream(c, stream) {
    return stream_map(x => c * x, stream);
}

const C = pair(1, () => scale_stream(2, C));

stream_ref(C, 2);
```

https://github.com/user-attachments/assets/4d55a143-9b10-48f2-99b4-793aeabd78b4

---

The rest of the code snippets to test out will work as intended and can be taken from #3570 which already has multiple test codes.

### Checklist

- [X] I have tested this code


src/commons/sideContent/content/SideContentCseMachine.tsx
```diff
@@ -300,8 +302,29 @@ class SideContentCseMachineBase extends React.Component<CseMachineProps, State>
                 onClick={this.stepNextBreakpoint}
               />
             </ButtonGroup>
+
             {!this.isJava() && (
               <ButtonGroup>
+                <Tooltip content="Hide Dead Frames" compact>
```
gigopogo: One UX observation while testing: I found myself instinctively clicking this "Hide Dead Frames" button a second time, expecting it to act as a toggle and bring the dead frames back. Right now, the only way to revive the dead frames is to scrub the timeline back a step and return.

I did see the Prof's comment specifying that this should act as a one-time clear action rather than a toggle, so the current behavior definitely aligns with that!

martin-henz: Maybe we can change the tooltip a bit to avoid this. "Hide" may sound reversible: "Unhide". If we say: "Clear dead frames" or "Remove dead frames" maybe the suggestion of reversibility will be less strong.

Akshay-2007-1: Will do so Prof!
That does make a lot of sense!

src/commons/sideContent/content/SideContentCseMachine.tsx
```diff
@@ -42,6 +42,7 @@ type State = {
   lastStep: boolean;
   stepLimitExceeded: boolean;
   chapter: Chapter;
+  hideDeadFrames: boolean;
```
sayomaki: Consider renaming this to something else, as the discussion earlier implies that using the term "hide" indicates that it is reversible.

src/commons/sideContent/content/SideContentCseMachine.tsx
```diff
@@ -300,8 +302,29 @@ class SideContentCseMachineBase extends React.Component<CseMachineProps, State>
...
+                          () => {
+                            CseMachine.setHideDeadFrames(this.state.hideDeadFrames);
+                            CseMachine.clearCachedLayouts();
+                            CseMachine.redraw();
+                          }
+                        );
+                      }
+                    }}
+                    icon="eraser"
+                    disabled={!this.state.visualization}
```
sayomaki: You might also want to consider disabling the button once it is pressed, to further indicate that it is one-time action and cannot be undone.

src/features/cseMachine/components/values/ArrayValue.tsx
```diff
@@ -118,6 +118,9 @@ export class ArrayValue extends Value implements IHoverable {
   };
 
   draw(): React.ReactNode {
+    if (Layout.hideDeadFrames && !this.isLive) {
+      return null;
+    }
```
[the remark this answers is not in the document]
sayomaki: Do double check the usage of `isLive` here.

src/features/cseMachine/CseMachineLayout.tsx
```diff
@@ -91,6 +92,8 @@ export class Layout {
   /** all environment and value IDs that are live in the current context */
   static liveEnvIDs: Set<string> = new Set();
   static liveObjectIDs: Set<string> = new Set();
+  /** hide non-live frames temporarily for the current step */
+  static hideDeadFrames: boolean = false;
```
sayomaki: Ditto here, maybe reconsider the naming for this boolean flag.

PR #3627, opened 2026-03-04
CSE Machine : Function Descriptions modification for Frame overlap

### LINKED ISSUE #3602: CSE Machine: Function descriptions during print mode overlaps with frames
Descriptions for function objects displayed when print mode is on would result in overlaps, as pointed out by @Akshay-2007-1. This issue exists in the current version of Source Academy, which makes it difficult to make out the names declared in frames that were overlaid.

Example 1:
```
function foo(n) {
    let arr = [];
    let i = 0;
    while (i < n) {
        function bar() {
            return 3;
        }
        arr[i] = bar;
        i = i + 1;
    }
}
foo(3);
```
Image

Example 2:
```
function foo() {
    function f(x) {
        return y => x + 3;
    }
    return f;
}
foo()(4);
foo()(5);
```
Image

### LINKED ISSUE #3624: CSE Machine : Implement Function description box width capping and text truncation
## Introduction
This issue, I am raising is extremely similar to #3606 even to the very title of it! We (@Oofky and I) arise at this issue while trying to fix the issue #3602 along with keeping in mind the new implementation of #3594, #3592 and the issues they themselves have in their working together illustrated by #3623 and #3603. 
 
My motivation and solution to the descriptions of functions overlapping with frames in the Print mode actually came from the issue raised at #3606 and #3607.  After a long discussion, @Oofky and I came to a conclusion that it is in our best interest to put the _function description boxes_ to the bottom of the function eyeballs instead of to the right. This is solely because more often than not, we find ourselves scrolling horizontally a LOT more than vertically while using the CSE machine due to the excessive widths of the frames and some other times, frames stacking together horizontally in each level! 

Hence, instead of adding to the already high horizontal width in the Printable mode, we thought it would be better to put the boxes underneath the eyeball objects representing functions. This does not completely eradicate the problem raised by #3602 since even if you put it down, if the description box is long enough, it still becomes longer than the function object as shown in the picture below : 
Image

---

The contrast of how it would look for the boxes to be to the under the function objects instead of to the right (if there was no overlap in either of the case) is shown below : 

Image

## Central Issue

This new implementation of the boxes underneath the function object quickly falls apart when you realize that there is NO limit to how long the function descriptions are in the printable mode. This is shown as follows : 

Image

Therefore, we want to limit i.e., cap this length at some value.

## TODO after solution
Hence, we would like to patch this just like how extremely wide frames would be as per #3606. Once that is achieved, we can successfully make sure that the gap between frames in each level in the printable mode is FIXED at the cap set, and according to the Layout team's PR #3594 invariant, fix the x-coordinates there and there is NO possibility that the function descriptions overlap with the frames. Additionally we also have to make sure that if the boxes are underneath the function objects, the frame's length i.e., y-coordinates are increased and fixed in the Printable mode versus the normal mode where its not required!

In fact, this is where #3607 comes in, we do adhere to fixing the y-coordinates, its just that in the printable mode, they will be shifted down a little bit and then fixed.

### LINKED ISSUE #3639: CSE Machine : Closing parantheses during truncation of descriptions
During the Printable mode or while hovering on the normal mode, the function objects (eyeball - like structures) result in their descriptions' boxes through tooltips underneath them. These descriptions have a specific format as can be seen in the image : 

Image

What we need when long functions like that get truncated due to their sheer length (as can be seen above) is that there is a closing parentheses, ")" after the ellipse "..." describing the rest of the parameters in the Printable mode. Moreover, we also need a closing curly bracket "}" after the truncation  of the body!

Do note that this issue gets raised only when #3627 gets approved and hence #3624 gets fixed!

Akshay-2007-1:
### Description

This PR addresses the printable mode overlap issue identified in #3602 and is relevant to the issues #3624. 

This PR:
- Moves function descriptions in Print mode to the bottom of function objects (instead of being to the right of function objects), which is also consistent with the textbook. (Refer to [@martin-henz's comment](https://github.com/source-academy/frontend/issues/3624#issuecomment-3989997721) ) This allows the CSE machine to be more compact.
- Implements a cap for function descriptions in Print mode, which was previously not implemented as seen below:

| Before | After |
|--------|--------|
| image | image |

- Assigns more vertical space for function objects to account for edge cases of overlapping function descriptions that involve arrays and nested recursive arrays/lists with closures. (Do note that though these scenarios seldom occur in the actual course work, we assume the worst case!)
- Do note that this elongation happens only in the **Printable** mode so that unnecessary visual gaps are not present in the normal mode where descriptions only appear on hovering!

| Normal | Printable |
|--------|--------|
| image | image |

Clearly, in the Printable mode we dynamically make sure that the height is MORE than the normal mode as illustrated above

- Before passing on the frames to the Layout team's implementation of the _ghost_ algorithm, we change the widths of frames that includes the bindings and total widths of arrays including recursive calculations as well. This way, before the x-coordinates are fixed, we ensure that they are fixed in the way we want them to be making sure that no overlaps happen!

| Before | After |
|--------|--------|
| image | image |
| image | image |

- In the normal mode, for function descriptions that are very wide, even on hovering, we will only see the compact form of the description of that function object while for the full view, you can click on the object! This is demonstrated below : 

https://github.com/user-attachments/assets/b229a180-09bd-4a0f-aa82-6f5b1991b0dc

- Earlier, the parantheses and the body closing curly bracket were not in place due to truncation but a recent commit fixed that!
image

Closes #3602
Closes #3624
Closes #3639

### Concerns:
- Since more vertical space is allocated in frames to account for function descriptions being below function objects, there is now a new issue. After clicking _Clear Dead Frames_, if function objects are dead, it creates a very ugly elongated frame. This can be fixed in a future PR after all issues regarding the function description's layout is fixed and an issue will be made for the same soon.
For the code below,
```javascript
function foo(x) {
    return x;
}
const b = x => x => x => 5;
b(1)(1)(1);
function e() {
    return [() => 1, () => 1, () => 1];
}
e();
foo(1);
foo(() => 1);
foo(1);
const a = () => 5;
const c = () => 5;
const d = () => 5;
```
| Before erasure | After erasure |
|--------|--------|
| image | image |

- Sometimes, the width calculation done before passing on to the _ghost_ algorithm for coordinate fixing goes a bit beyond what is expected and the visual gap increases more than expected in both the normal and the print modes! We are working towards fixing that. The same is illustrated below. (Credits to @ThatLi for the code being used as the test case here)
```javascript
function foo() {
    function f(x) {
        return y => x + 3;
    }
    return f;
}
foo()(4);
foo()(5);
```

| Before | After |
|--------|--------|
| image | image |

### Type of change

### How to test
Test our code out with any of the following code snippets. The expected result also is given right underneath!
### Test 1
```javascript
// Test B: sibling frames with closures + nested arrays
function make_frame(id, n) {
  const arr = [];
  let i = 0;
  while (i < n) {
    const k = id * 100 + i;
    function add(x) {
      return x + k;
    }
    function combo(x) {
      return add(x) + k;
    }
    arr[i] = [add, combo, [list(add, combo), [i, i + 1, i + 2]]];
    i = i + 1;
  }
  return () => arr;
}

const frames = [];
let j = 0;
while (j < 4) {
  frames[j] = make_frame(j, 4);
  j = j + 1;
}

frames[0]();
frames[1]();
frames[2]();
frames[3]();
``` 
| Before | After |
|--------|--------|
| image | image |

Do note that for the above case, I am attaching only a fraction of the Layout for demonstration sake, there is a lot more frames actually. Moreover, note that one must increase the maximum steps to 4000 before running the above code.

### Test 2
```javascript
const arr = [
  x => x + 1,
  [y => y * 2, [z => z - 3, 99]],
  5
];
arr;
```
| Normal | Printable |
|--------|--------|
| image | image |

### Test 3
```javascript
function pack(n) {
  return [x => x + n, [y => y * n, [z => z - n]]];
}
const a1 = pack(2);
const a2 = pack(3);
[a1, a2];
```
| Normal | Printable |
|--------|--------|
| image | image |

For more test cases, please refer to the ones given by @ThatLi at #3602. You will notice that they dont overlap anymore!

### Checklist

- [X] I have tested this code

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11877,12 +11877,12 @@ __metadata:
   linkType: hard
 
 "react-simple-keyboard@npm:^3.6.27":
-  version: 3.8.167
-  resolution: "react-simple-keyboard@npm:3.8.167"
+  version: 3.8.171
+  resolution: "react-simple-keyboard@npm:3.8.171"
   peerDependencies:
     react: ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0
     react-dom: ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0
-  checksum: 10c0/0e41a516cc73f2986f95804fd5e3c72e14e0d11885b559a8585cd45dac100d462e7017bdd7c43122593c6e84d55623f62bb8d1b76380e2eebbc011dcefc5d7b8
+  checksum: 10c0/eea988f176a22533321dff8b83e75c06d643c6ecf5f6b83d27df0283713233e7016d8fa334c821f1296f4bbda9c2ffbfd9672be5f3f7c6f2621e4a3f34038016
   languageName: node
   linkType: hard
 


src/features/cseMachine/components/Frame.tsx
```diff
@@ -160,66 +165,6 @@ export class Frame extends Visible implements IHoverable {
...
-
-        while (isDataArray(curr)) {
-          if (seenSpine.has(curr)) break;
-          seenSpine.add(curr);
-          const blockWidth = curr.length * Config.DataUnitWidth;
-          const head = curr[0];
-
-          if (isDataArray(head)) {
-            const branchWidth = getChainWidth(head);
-            maxWidth = Math.max(maxWidth, currentSpineX + branchWidth);
```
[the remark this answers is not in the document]
Akshay-2007-1: In printable mode, using value.totalWidth for function values in Frame.tsx is intentional, not a spacing bug. FnValue.totalWidth and GlobalFnValue.totalWidth include the visible printed description box to the right of the eyeball, so the frame’s totalDataWidth needs that full width budget to avoid overlap with whatever comes next.

src/global.ts
RichDom2185, line 1: ? Can we revert?

.coverage-fix.tmp
RichDom2185, line 1: ???

et --hard 25755a4c
RichDom2185, line 1: ???!?!??!!

RichDom2185 (commented):
@Akshay-2007-1 please make a PR to revert these changes.

PR #3629, opened 2026-03-05
1088

martin-henz:
### Description

This PR bumps js-slang to 1.0.88, which does not include support for the language Scheme. As a result, this PR removes any scheme implementation.

### Type of change

### How to test

I tested that everything builds and runs locally, and checked that Scheme is no longer available.

### Checklist

- [x] I have tested this code

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4971,9 +4971,9 @@ __metadata:
   linkType: hard
 
 "ace-builds@npm:^1.36.3, ace-builds@npm:^1.4.12, ace-builds@npm:^1.42.1":
-  version: 1.43.6
-  resolution: "ace-builds@npm:1.43.6"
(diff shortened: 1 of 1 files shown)


JayeJyn (dismissed):

src/features/cseMachine/CseMachineLayout.tsx
```diff
@@ -51,7 +51,7 @@ import {
   isUnassigned,
   setDifference
 } from './CseMachineUtils';
-import { Continuation, isContinuation, isSchemeNumber, isSymbol } from './utils/scheme';
+import { Continuation, isContinuation } from './utils/scheme';
```
s-kybound: Continuation and isContinuation should be moved away from utils/Scheme

s-kybound (changes requested):
LGTM except for the utils/scheme folder, should be renamed to something like utils/continuations, or logic should be moved

RichDom2185 (dismissed):
LGTM, thanks!

RichDom2185 (commented):
Why are we reverting the updates back to older versions?

RichDom2185:
https://github.com/source-academy/frontend/pull/3643

PR #2881, opened 2024-03-26
Fix and Improve Ground Control

RichDom2185:
### Description

The mutable state should have been migrated to use `useRef` when the component was transitioned to a functional component.

This was a regression introduced in #2875.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### How to test

It should now work, aka fixes this bug:

telegram-cloud-photo-size-5-6059905916867819461-y

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #2875, opened 2024-03-25
Clean up team assessments

RichDom2185:
### Description

To rebase and force-push following the merge of #2548.

Breaking change by technicality.

Summary of changes:

* Cleans up some unnecessary styles
* Reverts Ground Control back to AG Grid
* Remove ground control and team formation from "workspaces"
* Favor Blueprint components over custom ones

More cleanup will be done in the future.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Code quality improvements

### Checklist

- [x] I have tested this code
- [ ] I have updated the documentation

PR #3637, opened 2026-03-06
Fix Ground Control: migrate gridApi to useRef, remove connect() container

martin-henz:
### Description

[This PR is an experiment in using Claude Code to revive old PRs and clean up the repo. The description below was generated by Claude Code, Sonnet 4.6. I will post the prompt conversation below.]

Revives and re-applies #2881, updated for the current state of the codebase.

The mutable `gridApi` variable was never properly persisted across re-renders after the component was converted to a functional component in #2875. It was declared as a plain `let` inside the component body, which means it gets reset to `undefined` on every render — so any call to `resizeGrid()` after the initial mount was a no-op.

**Changes:**
- Migrate `let gridApi` → `useRef<GridApi>()` (the actual bug fix)
- Remove `DispatchProps` type and `GroundControlContainer.ts`; dispatch actions directly via `useDispatch()`, consistent with the rest of the codebase
- Update `academyRoutes.ts` to import `GroundControl` directly instead of via the container
- Hook up `onGridSizeChanged` on the grid so `resizeGrid` is actually called on window resize

### Type of change

### How to test

Open Ground Control as an admin. The assessment grid should render and columns should size correctly. Resizing the window should also resize the grid columns.

### Checklist

- [x] I have tested this code


RichDom2185 (approved):
LGTM, tested, thanks!

PR #3642, opened 2026-03-09
Fixed liveness bug for null array units 

Oofky:
### Description

This PR fixes a small bug we found where null array units sometimes still show the default stroke color despite being dead/unreferenced. It is a very minor fix and follows the same logic as other objects such as regular array units, so it will not break anything.

### Type of change

### How to test

Run the following code (or any code where the frame that references an array with a null unit is dead), and observe that the stroke representing the null unit is greyed out when appropriate:
```
{
    let a = list(2, null, null);
}
const c = 3;
```
Before:

https://github.com/user-attachments/assets/5ff335ba-7923-449b-bade-af2fa2c7db9f

After:

https://github.com/user-attachments/assets/c5f7c699-edf1-4aad-8f35-6f0b90f72f39

### Checklist

- [x] I have tested this code

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "private": true,
   "type": "module",
   "name": "frontend",
-  "packageManager": "yarn@4.6.0",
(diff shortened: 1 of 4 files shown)


martin-henz (approved):
Looks good!

Issue #3645, opened 2026-03-10
Invalid Playground Language State in localStorage


martin-henz:
## Issue

If a user's browser localStorage contains an invalid chapter/variant combination (e.g. chapter 1 + `explicit-control`, which only applies to chapter 4), the playground crashes on load with:

```
Error: Language config not found for chapter 1 variant explicit-control
```

This can happen when switching feature flags (e.g. the conductor flag) causes the playground to temporarily use a different language setting, which then gets persisted to localStorage.

## Fix

When loading stored state in `createStore.ts`, validate that the stored chapter+variant combination exists in `ALL_LANGUAGES`. If not, fall back to the default language config.

---

*Generated by Claude Code, version 2.1.72*

PR #3646, opened 2026-03-10
Fix crash when stored playground language state is invalid

martin-henz:
AI-generated PR, fixing https://github.com/source-academy/frontend/issues/3645

Generated with Claude Code, Sonnet 4.6.

### Description

Validate the stored chapter+variant combination against ALL_LANGUAGES when loading from localStorage. Falls back to the default language config if the combination is invalid (e.g. chapter 1 + explicit-control). 

### Type of change

### How to test

Start frontend using localhost, switch to conductor in feature flags, choose Source 1. Then reload in the browser. This would crash the frontend when the feature flag is reset to false, because the "Source 1" language does not have a CSE machine implementation.

### Checklist

- [x] I have tested this code


RichDom2185 (changes requested):
This works but behaves differently than what the issue suggests.

martin-henz:
> > > This works but behaves differently than what the issue suggests.
> > 
> > 
> > In what way is this different from what the issue suggests?
> 
> The issue suggests:
> 
> > invalid combination of lang + variant → fallback to default lang + variant combination
> 
> But the logic implemented is:
> 
> > invalid lang → fallback to default lang && invalid variant → fallback to default variant

Hmm, the implementation checks:

```
          const validCombination = ALL_LANGUAGES.some(
            lang => lang.chapter === chapter && lang.variant === variant
          );
```
so it looks for a valid combination. It falls back on the default lang/variant only if the combination is not valid.

martin-henz:
Revived this: merged `master` in (branch was ~227 commits behind) and resolved the conflict in `createStore.ts`.

@RichDom2185 — to address your note that it behaved differently from what the issue suggests: the invalid-combination fallback now resets to `defaultLanguageConfig` (a known-valid chapter+variant pair from `ALL_LANGUAGES`) rather than the individual default context fields, so it explicitly falls back to *the default language config* as the issue describes. When the combo is invalid we reset the whole language wholesale rather than trying to preserve a valid chapter and only fix the variant — since this is a crash-fix, dropping to the known-good default seemed the safest, most predictable behavior. Let me know if you had the chapter-preserving variant in mind instead.

Prettier/eslint/tsc clean; CI re-running.

PR #3732, opened 2026-04-03
Addition of the Skip Feature in the Source Academy Game

ItsByt:
### Summary of change:
This PR proposes the new implementation of the skip feature in the Source Academy Game, which enables users to skip through dialogue.

### Type of change

### Motivation and Context:
Users may feel that some dialogues are particularly too lengthy, and would want some way to skip through it conveniently without having to repeatedly click through the dialogue. This feature enables users to skip through dialogue with the click of a button (the skip button), or by clicking the "s" button on their keyboard. To prevent accidental usage, users can also toggle in their settings whether to have a confirmation appear whenever they use the skip button. If on, they will be prompted with the confirmation on their usage, and if off, any use of the skip feature immediately skips through the dialogue. 

We also note that the skip feature only skips until the next required prompt by the user (e.g. location changes, questions that require the user to give an answer etc.), or the end of the dialogue. Trying to re-use the skip feature will not skip any further in such cases.

Also note that the default setting for the confirmation setting is for the confirmation to be turned ON. Users will have to manually turn this confirmation off should they want no confirmation. However, once turned off, the settings will be saved between chapters and users need not re-turn it off every time.

### How to test

Choose any chapter (I chose Chapter 0 for simplicity and as it is the easiest to illustrate)

https://github.com/user-attachments/assets/33ff3f6c-ba79-4e94-aa10-0ce63ed0f777

When in a dialogue, there should now be blue button at the top right corner of the dialogue box. Click it to skip through dialogue. The above is with the confirmation settings turned on.

https://github.com/user-attachments/assets/1509435b-4f92-465c-a1d4-6b1670d78493

The above is when the confirmation settings is turned off.

https://github.com/user-attachments/assets/88d35727-d674-4083-92f2-28310378a9ff

The above is when we use the "s" keyboard shortcut to skip through the dialogue. Note that this works regardless of whether we have the confirmation turned on or off. It is simply the equivalent of clicking the skip button.

https://github.com/user-attachments/assets/915c7c5c-9fa2-4f21-961f-4b75276b1717

We note that trying to use the skip button is useless if they are already at a point in the dialogue that cannot be progressed any further without their direct prompt. In the case where a user is explicitly required to answer a question, I have hidden the button so that it is not clickable directly on their screen until the prompt is answered, and temporarily disabled the "s" keyboard shortcut so clicking it does nothing in that instance.

### Checklist

- [x] I have tested this code


src/features/game/dialogue/GameDialogueManager.ts
```diff
@@ -37,39 +44,192 @@ export default class DialogueManager {
...
+        await new Promise(resolve => setTimeout(resolve, 50));
+      }
+    } finally {
+      this.isSkipping = false;
+    }
+  }
+
+  private async peekNextLine() {
+    const currentPart = (this.getDialogueGenerator() as any).currPart;
+    const currentLineNum = (this.getDialogueGenerator() as any).currLineNum;
```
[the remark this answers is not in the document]
ItsByt: Have moved peekNextLine to DialogueGenerator

public/assets/skip-icon.png
sayomaki, line 1: It would be preferred to use the S3 for the skip icon instead, just like any other game asset. It should not be put into frontend.

src/features/game/scenes/settings/Settings.ts
```diff
@@ -177,7 +177,11 @@ class Settings extends Phaser.Scene {
       : 1;
 
     // Save settings
-    await this.getSaveManager().saveSettings({ bgmVolume: bgmVol, sfxVolume: sfxVol });
+    await this.getSaveManager().saveSettings({
+      bgmVolume: bgmVol,
+      sfxVolume: sfxVol,
+      skipConfirm: false
```
sayomaki: Please double check this, it seems that this will override skipConfirm to false every time the settings are saved.

rsbuild.config.ts
```diff
@@ -23,7 +23,7 @@ export default defineConfig({
     })
   ],
   server: {
-    port: 8000
+    port: 8080
```
sayomaki: Please revert this, this should not be included in the change.

src/features/game/dialogue/GameDialogueManager.ts
```diff
@@ -38,39 +45,191 @@ export default class DialogueManager {
...
+    }
+  }
+
+  private createSkipButton(dialogueContainer: Phaser.GameObjects.Container) {
+    const gameManager = GameGlobalAPI.getInstance().getGameManager();
+
+    this.skipButton = new Phaser.GameObjects.Image(
+      gameManager,
+      screenSize.x - 62.5,
+      screenSize.y * 0.73,
```
sayomaki: A minor nit here but the position of the skip button should follow the dialogue constants, or be placed in the dialogue constants instead.

src/features/game/dialogue/GameDialogueManager.ts
```diff
@@ -38,39 +45,191 @@ export default class DialogueManager {
...
+      'skip-icon'
+    ).setInteractive({ useHandCursor: true });
+
+    this.skipButton.setDisplaySize(45, 45);
```
sayomaki: The size of the skip button should also be a constant set in `GameDialogueConstants` instead.

sayomaki (dismissed):
Some changes to consider for the team, refer to my comments below. Also, I think the current skip duration seems a bit slightly short, maybe can consider slowing it down a little bit (just my personal opinion on it from the videos), but otherwise great job!

src/features/game/scenes/settings/Settings.ts
```diff
@@ -74,18 +75,33 @@ class Settings extends Phaser.Scene {
...
+      SettingsConstants.volContainerOpts,
+      bgmVolIdx,
+      optHeaderPos[1][1]
+    );
+    // Create Skip Confirm Radio Buttons
+    this.skipConfirmRadioButtons = this.createOptRadioOptions(
+      ['ON', 'OFF'],
+      skipConfirmIdx,
+      optHeaderPos[2][1]
+    );
```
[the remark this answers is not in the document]
sayomaki: Please double check this, it should not just be ignored.

sayomaki:
Thanks for updating the PR! As noted by sentry (and I do think its a valid point), there appears to be an issue when trying to access the 3rd item (index 2) of the `optHeaderPos`, as it is reading from `getSettingsHeader`. You should add the missing entry to the header to fix the problem.

public/assets/skip-icon.png
ItsByt: Hi, could someone help me move this asset into the S3? Thank you!

src/features/game/scenes/settings/Settings.ts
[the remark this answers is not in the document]
ItsByt: Have fixed this

RichDom2185 (approved):
Looks good, thank you!

PR #3739, opened 2026-04-05
Feat: Manhattan Routing for Control and Stash Arrows

RogerZhang888:
### Description

Initially, control and stash arrows are all straight arrows that point directly to the destination. Now, they follow manhatting routing, just like all the arrows in the environment model. 

Note: In order to allow the control arrows to have enough space to terminate, I shifted the control to the left slightly using a new constant `ControlShiftX` inside `CseMachineControlStashConfig.ts`. 

### Type of change

### How to test

Run any Source program and enable control and stash to see the new arrows. 

### Checklist

- [ ] I have tested this code

### DIFF
--- a/src/features/cseMachine/components/arrows/ArrowFromControlItemComponent.tsx
+++ b/src/features/cseMachine/components/arrows/ArrowFromControlItemComponent.tsx
@@ -18,6 +18,10 @@ export class ArrowFromControlItemComponent extends GenericArrow<
     this.isLive = true; // Control items are always live
   }
 
+  protected getOriginFilterKey() {
+    return 'control' as const;
+  }
+
   protected calculateSteps() {
     const from = this.source;
     const to = this.target;
--- a/src/commons/application/__tests__/ApplicationWrapper.test.tsx
+++ b/src/commons/application/__tests__/ApplicationWrapper.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { Provider } from 'react-redux';
 import { createMemoryRouter, RouterProvider } from 'react-router';
 import { createStore } from 'src/pages/createStore';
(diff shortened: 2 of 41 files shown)


src/i18n/i18next.ts
Akshay-2007-1: I think this was accidental and should be renamed back!

RichDom2185, line 1: Please revert.

.coverage-fix.tmp
RichDom2185, line 1: Please delete.

.gitignore
```diff
@@ -5,9 +5,11 @@
 
 # testing
 /coverage
+.coverage-fix.tmp
```
RichDom2185, line 8: Not meant to be gitignored because you're not meant to run `yarn test-coveralls` locally.

RichDom2185: But I'm ok with adding it to gitignore if it helps

src/commons/sideContent/content/SideContentSubstVisualizer.tsx
```diff
@@ -678,16 +677,15 @@ function renderNode(currentNode: StepperBaseNode, renderContext: RenderContext):
   };
 
   // Entry point of rendering
-  const renderer = renderers[currentNode.type as keyof typeof renderers];
+  const renderer = (
+    renderers as unknown as Record<string, (node: StepperBaseNode) => React.ReactNode>
```
RichDom2185, line 681: Let's restore the original comment to explain why this cast is safe


```suggestion
    // All subclasses of stepper base node has its corresponding renderes
    renderers as unknown as Record<string, (node: StepperBaseNode) => React.ReactNode>
```

src/features/cseMachine/components/arrows/ArrowFromControlItemComponent.tsx
```diff
@@ -22,16 +23,29 @@ export class ArrowFromControlItemComponent extends GenericArrow<
     const to = this.target;
     if (!to) return [];
 
+    const terminalSegmentLength = Math.max(Config.ArrowHeadSize, Config.MinTerminalSegmentLength);
+    const postSourceStraightLength = Config.ArrowPostFrameStraightLength;
+    const targetY =
+      to.y() +
+      (to instanceof Frame && to.x() > ControlStashConfig.ControlItemWidth + 100
```
RichDom2185, line 35: Restore comment


```suggestion
      // Draw arrow slightly below frame corner if frame is far enough
      (to instanceof Frame && to.x() > ControlStashConfig.ControlItemWidth + 100
```

Issue #3607, opened 2026-02-26
CSE Machine: Stabilize Vertical Layout via Persistent Built-ins and Fixed Y-Coordinates


gigopogo:
### Related Pull Request
This issue is a continuation of the layout stabilization work in:
- **PR #3594:** [Fixing frames coordinates and preventing overlap of data and frames](https://github.com/source-academy/frontend/pull/3594)

### Description
In PR #3594, we addressed the horizontal overlapping of frames, but vertical instability remains a significant issue. Currently, the Global Frame's height expands and shrinks dynamically as built-in functions are called and then removed from the environment. This constant resizing causes the global frame and program frame to overlap if y coordinates of frames are fixed.

### The Problem
- **Built-in Volatility:** The disappearance of called functions makes it difficult for students to reference the "history" of the environment.
- **Vertical Instability:** Without fixed Y-coordinates, frames shift vertically as the number of bindings in parent frames changes, causing visual fatigue.
- **Layout Gaps:** Implementing fixed Y-coordinates based on maximum height can lead to excessive vertical white space (gaps) during steps where frames are sparsely populated.

### Proposed Improvements
1. **Permanent Global Bindings:** Keep built-in function signatures visible in the Global Frame once they have been invoked.
2. **Fixed Y-Coordinates:** Include caching of Y-coordinates too so frames do not jump vertically.

### Limitations of fixed Y-coordinates
- **Layout Gaps:** Implementing fixed Y-coordinates based on maximum height can lead to excessive vertical white space (gaps) between frames during steps.

martin-henz:
Nice description of the issue!

PR #3594, opened 2026-02-20
Fixing frames coordinates and preventing overlap of data and frames (additional feature: center alignment) 

gigopogo:
### Description

This PR introduces a fixed-coordinate layout system for the CSE Machine to prevent visual drifting when scrubbing through the execution timeline. By caching the coordinates of environment frames at the last step, this ensures that frames and data objects remain completely stationary throughout the program's execution. 

Additionally, this PR introduces two key visual improvements:
1. **Center Alignment Toggle:** Allows users to cleanly center the execution environment within the viewport for better readability.
2. **Sibling Frame Data Spacing:** Implements specific layout spacing calculations for sibling frame data (e.g., arrays and lists) to properly allocate space and prevent visual overlaps between sibling structures.

This aligns with the pedagogical goal of achieving completely stationary frames, providing a more consistent and less confusing visual tracing experience for students. 

**Note on Known Issues / Out of Scope (Discussed & Approved by Prof):**
The following behaviors are known and have been explicitly scoped out of this PR to be handled separately or prioritized as intended behaviors:
1. **Arrow Coordinates:** Arrows are currently pointing to incorrect frame coordinates in the fixed layout. This is being passed over to the Arrow Team for a follow-up fix.
2. **Printable Mode Overlap:** Frames occasionally overlap in printable mode (verified as a pre-existing issue on the live site). A separate issue will be made.
3. **Non-Sibling Array Spacing Overlap:** While spacing for *sibling* frame data is now handled, there are edge cases where arrays generated outside of sibling assumptions visually overlap with environment frames (verified as a pre-existing issue). A separate issue will be made.
4. **Empty Spaces with PR #3592:** Hiding dead frames leaves empty visual gaps. CSE liveness team are working on this in our merge branch. 

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Code quality improvements

### How to test

1. Paste the provided code snippet into the Source Academy playground.
2. Run the code and use the timeline slider to scrub back and forth between the steps.
3. **Expected Behavior 1 (Fixed Layout):** The environment frames should remain completely anchored in their fixed positions without sliding horizontally while data (arrays, function objects , etc.)  renders beside it . 
[...]

PR #3741, opened 2026-04-05
CSE Machine: Fixed y coordinates

gigopogo:
## Context
This PR builds upon the layout stabilization efforts introduced in **#3594** (Fixing frames coordinates and preventing overlap of data and frames) and **#3695** (Fixing referenced global functions throughout execution). 

Previously, Y-coordinates could shift dynamically between steps, causing visual jitter and making it difficult to track the execution flow. With this PR, the layout is now completely fixed throughout all execution steps. The machine calculates and renders the dynamic spacing required for the entire execution at the very beginning, ensuring that frames, bindings, and data structures remain anchored to their static Y-coordinate levels as the user steps through the program. (This also applies to printable mode)

This PR also fulfills the features outlined in #3607 . 

## Changes Made
* Locked Y-coordinates for Frames, Functions, Arrays, Bindings ,Primitive values , etc to their respective y coordinates
* Replaced `(this.text as any)` typecasting hacks with explicit `setX` and `setY` setter methods across visible components for safer coordinate reassignments. (Brought up by @RichDom2185 in  **#3594** )

## How to Test
1. Paste the following code 

```javascript
function new_stream(a, b) { 
  const c = list(1, 2, 3);
  let d = list(1, c);
  return pair(a, () => stream_map(x => -x, new_stream(-b, a + 2))); 
} 
const st2 = new_stream(1, 2); 
eval_stream(st2, 4);

```

Step through the execution., the layout remains fixed throughout.

Tip: Turn ON Center Alignment! This makes it incredibly easy to appreciate the clean, static horizontal levels of the newly fixed layout as you step through the code

https://github.com/user-attachments/assets/de142ce1-6859-48ca-925d-7866a16cc62a

  ### Checklist
  
  
  
  - [x] I have tested this code

### DIFF
--- a/src/features/cseMachine/components/Text.tsx
+++ b/src/features/cseMachine/components/Text.tsx
@@ -62,7 +62,7 @@ export class Text extends Visible implements IHoverable {
     options: Partial<TextOptions> = {}
   ) {
     super();
-    this._x = x; 
+    this._x = x;
     this._y = y;
     this.options = { ...this.options, ...options };
 


src/features/cseMachine/components/Text.tsx
```diff
@@ -56,8 +56,8 @@ export class Text extends Visible implements IHoverable {
 
   constructor(
     readonly data: Data,
-    readonly _x: number,
-    readonly _y: number,
+    protected _x: number,
+    protected _y: number,
```
RichDom2185: Why do we need to change to protected if we have a setter? Did you mean private?

gigopogo: I tried private first. But because Visible already defines them as protected, TypeScript threw an inheritance error for trying to change the access modifier in the child class. I just changed them to normal parameters and assigned them to the inherited properties.

martin-henz: "I just changed..."

I don't see the change. I still see "protected" in front of the two constructor parameters.

Issue #3727, opened 2026-04-01
[Specs] CSE Machine: Stream Visualizer


DaRealTristan:
**Basic Criterion for visualising streams**
When a nullary function returns a pair, draw a yellow dotted arrow between the nullary function and the pair.
There may be multiple dotted arrows between the nullary function and the pair..

martin-henz:
We can track CALL 0 instructions and look at the top of the stack when all computation resulting from the call is done. The return value of the nullary function  will now be on top of the stash. If that value is a pair, we draw the dotted line.

In more detail: Let's say the control pointer is at position i when the CALL 0  instruction gets executed. In other words, "CALL 0" sits at control[i]. We should check the first time the control pointer reaches (i - 1). At that time the return value of the nullary function will be on top of the stash. If that value is a pair we can draw a yellow dotted line from the nullary function to that pair.

In even more detail: In the CSE machine, introduce a stack for nullary function calls, nullaryStack. Whenever, we execute a CALL 0 instruction at control[cc], push the tuple (cc, top-of-stask) onto the stack. Here cc refers to "control counter", and top-of-stack is the function to be called. Whenever cc is decreased, check top of nullaryStack. Let's say it's the tuple (cc', nullaryFun). If cc = cc' - 1, pop the tuple from nullaryStack and check if the top of the stash is a pair. If yes, draw a yellow dotted line from nullaryFun to the pair.

PR #3747, opened 2026-04-06
CSE machine: StreamViz arrow drawing

of89p:
### Description

Created new panel in CSE machine for Stream Visualization (pairs and nullary functions belonging to streams).
 
Displays results from streamViz js-slang PR (https://github.com/source-academy/js-slang/pull/1939). Specifically, we use results from streamLineage from the Context object to draw dotted arrows from a nullary  function to the pair(s) it returns.

The map streamLineage (Found in context object, types.ts) that maps the ID of nullary functions to the pairs it returns.

Everytime a CALL0 is called in the control, we push a tuple (nullaryFnID, controlStackHeight) to the pendingStreamFnStack[]. Within generateCSEMachineStateStream, we check the whether the height of the control stack has reached the height of i-1, where i is the controlStackHeight when we pendingStreamFnStack.peek(). If so, we pop pendingStreamFnStack.pop() and we add the pair that is at the top of the stash to the array of pairs that the nullaryFnID maps to, using streamLineage.

See also frontend issue: https://github.com/source-academy/frontend/issues/3727

### Type of change

### Checklist

- [ ] I have tested this code


package.json
```diff
@@ -200,6 +200,6 @@
   },
   "resolutions": {
     "@types/estree": "1.0.8",
-    "vite": "^8.0.0"
+    "vite": "^8.0.0",
```
Akshay-2007-1: This needs to be urgently reverted! It prevents me from running any yarn commands and essentially blocks me from doing 'yarn run start' for localhost testing to do further reviews!

Akshay-2007-1: fixed at 02c8ea1

tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
...ound.test.tsx","./src/pages/sicp/Sicp.tsx","./src/pages/sicp/__tests__/Sicp.test.tsx","./src/pages/sicp/subcomponents/CodeSnippet.tsx","./src/pages/sicp/subcomponents/SicpExercise.tsx","./src/pages/sicp/subcomponents/SicpIndexPage.tsx","./src/pages/sicp/subcomponents/SicpLatex.tsx","./src/pages/sicp/subcomponents/SicpToc.tsx","./src/pages/sicp/subcomponents/__tests__/CodeSnippet.test.tsx","./src/pages/sicp/subcomponents/__tests__/SicpExercise.test.tsx","./src/pages/sicp/subcomponents/__tests__/SicpIndexPage.test.tsx","./src/pages/sicp/subcomponents/__tests__/SicpLatex.test.tsx","./src/pages/sicp/subcomponents/__tests__/SicpToc.test.tsx","./src/pages/sicp/subcomponents/chatbot/ChatBox.tsx","./src/pages/sicp/subcomponents/chatbot/Chatbot.tsx","./src/pages/sicp/subcomponents/chatbot/ChatbotCodeSnippet.tsx","./src/pages/sicp/subcomponents/chatbot/types.tsx","./src/pages/sourcecast/Sourcecast.tsx","./src/pages/stories/Stories.tsx","./src/pages/stories/StoriesTable.tsx","./src/pages/stories/Story.tsx","./src/pages/stories/StoryActions.tsx","./src/pages/welcome/Welcome.tsx","./src/routes/routeGuard.ts","./src/routes/routerConfig.ts","./src/sw/precache.ts"],"errors":true,"version":"6.0.2"}
```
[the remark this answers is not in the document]
Akshay-2007-1: fixed at 0013166

martin-henz:
This program seems to get stuck at step 81, now:
```
function make_ints(i) {
    return pair(i, () => make_ints(i + 1));
}

stream_ref(make_ints(0), 10);
```
Screenshot 2026-04-22 at 10 57 53 AM

RichDom2185 (changes requested):
I can reproduce the issue flagged by Prof @martin-henz. It's a blocker

Akshay-2007-1 (commented):
1. FnValue.tsx — the main fix (root cause of the bug)

  The bug: The draw() method runs on every render step. At the very start of draw(), it did this._streamArrows = [] —
  wiping the arrow list. But before that wipe, it was calling getStreamLineage() and trying to use the result. So the
  arrows were built, then immediately thrown away before they could be rendered. The stream visualization never actually
   drew anything.

  The fix: Move getStreamLineage() inside the if (getPairCreationMode()) block, after the wipe. Now arrows are built
  fresh each render, in the right order.

  Also removed: ~30 lines of dead constructor code that tried to draw stream arrows during object construction — but
  since draw() always wiped _streamArrows = [] at the start, anything set in the constructor was always discarded before
   rendering. It was completely unreachable code that had no effect.

  Also fixed: centerX (the horizontal center of the function circles) was being computed after the arrows were created.
  Arrows use centerX for their starting position, so they could start from x=0. Moved centerX calculation to the very
  top of draw().

  Also removed: A throw new Error('Closure has no main reference...') that could never actually fire (the value is
  always set before draw() is called), but if it ever did, it would crash the whole visualization silently.

  ---
  2. ArrayValue.tsx — removed dead constructor code

  The pair (array) constructor had code that tried to notify its parent function value about itself, so the function
  could draw an arrow to it. This was the other half of the dead code above — it set Layout.pendingFnLink = true as a
  signal. Since FnValue's constructor code reading that flag was already dead, this code had no effect either. Removed
  it, along with the unused Closure import.

  ---
  3. CseMachineLayout.tsx — removed dead state variables

  Six static fields (currentDarkPairs, currentLightPairs, etc.) were declared and reset in setContext(), but never read
  Also removed: A throw new Error('Closure has no main reference...') that could never actually fire (the value is always set before draw() is called), but
  if it ever did, it would crash the whole visualization silently.

  ---
  2. ArrayValue.tsx — removed dead constructor code

  The pair (array) constructor had code that tried to notify its parent function value about itself, so the function could draw an arrow to it. This was the
   other half of the dead code above — it set Layout.pendingFnLink = true as a signal. Since FnValue's constructor code reading that flag was already dead,
  this code had no effect either. Removed it, along with the unused Closure import.

  ---
  3. CseMachineLayout.tsx — removed dead state variables

  Six static fields (currentDarkPairs, currentLightPairs, etc.) were declared and reset in setContext(), but never read anywhere — they were leftovers from
  an earlier implementation. Also removed pendingFnLink (the boolean flag used by the dead constructor code above) and streamNodes = null which was assigned
   but never used in the JSX render.

  ---
  4. updateInspector.ts — stop silently swallowing errors

  The catch block previously had a comment saying errors were "most likely harmless, we can pretty much ignore this." This meant if the CSE machine
  visualization crashed during rendering, it would look like the diagram was just "stuck" — no error, no feedback. Changed to console.error(...) so any
  crash now shows up in the browser console with a real message.

  ---

I made an AI written summary for the changes I made in the recent commit for better understanding. @DaRealTristan do look into if the actual Stream dotted arrows work or my changes are breaking!

DaRealTristan:
> 1. FnValue.tsx — the main fix (root cause of the bug)
> 
> The bug: The draw() method runs on every render step. At the very start of draw(), it did this._streamArrows = [] — wiping the arrow list. But before that wipe, it was calling getStreamLineage() and trying to use the result. So the arrows were built, then immediately thrown away before they could be rendered. The stream visualization never actually drew anything.
> 
> The fix: Move getStreamLineage() inside the if (getPairCreationMode()) block, after the wipe. Now arrows are built fresh each render, in the right order.
> 
> Also removed: ~30 lines of dead constructor code that tried to draw stream arrows during object construction — but since draw() always wiped _streamArrows = [] at the start, anything set in the constructor was always discarded before rendering. It was completely unreachable code that had no effect.
> 
> Also fixed: centerX (the horizontal center of the function circles) was being computed after the arrows were created. Arrows use centerX for their starting position, so they could start from x=0. Moved centerX calculation to the very top of draw().
> 
> Also removed: A throw new Error('Closure has no main reference...') that could never actually fire (the value is always set before draw() is called), but if it ever did, it would crash the whole visualization silently.
> 
> 2. ArrayValue.tsx — removed dead constructor code
> 
> The pair (array) constructor had code that tried to notify its parent function value about itself, so the function could draw an arrow to it. This was the other half of the dead code above — it set Layout.pendingFnLink = true as a signal. Since FnValue's constructor code reading that flag was already dead, this code had no effect either. Removed it, along with the unused Closure import.
> 
> 3. CseMachineLayout.tsx — removed dead state variables
> 
> Six static fields (currentDarkPairs, currentLightPairs, etc.) were declared and reset in setContext(), but never read Also removed: A throw new Error('Closure has no main reference...') that could never actually fire (the value is always set before draw() is called), but if it ever did, it would crash the whole visualization silently.
> 
> 2. ArrayValue.tsx — removed dead constructor code
> 
> The pair (array) constructor had code that tried to notify its parent function value about itself, so the function could draw an arrow to it. This was the other half of the dead code above — it set Layout.pendingFnLink = true as a signal. Since FnValue's constructor code reading that flag was already dead, this code had no effect either. Removed it, along with the unused Closure import.
> 
> 3. CseMachineLayout.tsx — removed dead state variables
> 
> Six static fields (currentDarkPairs, currentLightPairs, etc.) were declared and reset in setContext(), but never read anywhere — they were leftovers from an earlier implementation. Also removed pendingFnLink (the boolean flag used by the dead constructor code above) and streamNodes = null which was assigned but never used in the JSX render.
> 
> 4. updateInspector.ts — stop silently swallowing errors
> 
> The catch block previously had a comment saying errors were "most likely harmless, we can pretty much ignore this." This meant if the CSE machine visualization crashed during rendering, it would look like the diagram was just "stuck" — no error, no feedback. Changed to console.error(...) so any crash now shows up in the browser console with a real message.
> 
> I made an AI written summary for the changes I made in the recent commit for better understanding. @DaRealTristan do look into if the actual Stream dotted arrows work or my changes are breaking!

Just checked with the stream examples we have and it works! Thanks for all the help @Akshay-2007-1 ! Will work on the js-slang map issue when I am back in Singapore

PR #3758, opened 2026-04-10
CP3108_26MH

elsxyss:
Small aesthetic adjustments following meeting:
- error message updated to "Binary Tree View" instead of "Render binary tree", etc
- arrow rendering made nicer (vertical alignment to boxes)


martin-henz (approved):
Looks good.

PR #3783, opened 2026-04-15
Data visualizer (draw_data): Minor Bug: Retaining previous Call step when toggling between view modes + Tightening tree verification

elsxyss:
Minor Bug: Retaining previous Call step when toggling between view modes + Tightening tree verification


sayomaki (approved):
Looks good to me, thanks!

PR #3802, opened 2026-04-24
Adapt to js-slang 1.0.92 API changes

martin-henz:
## Summary

Bumps `js-slang` from 1.0.90 to 1.0.92 and adapts the frontend to the
breaking changes introduced upstream in
[source-academy/js-slang#1893](https://github.com/source-academy/js-slang/pull/1893).

Without these fixes, 31 tests across 3 test files were failing after the
bump. With them, all 644 tests pass and `tsc -b` is clean.

## Changes

- **`TypeError` → `RuntimeTypeError`.** `js-slang/dist/utils/rttc`
  renamed the exported class. Updated `ContextMocks.ts` accordingly.

- **`loadSourceModules` API.** The 1st argument is now
  `Record<string, ModuleInfo>` (was `Set<string>`) and the 3rd is
  `Partial<ImportLoadingOptions>` (was a boolean `loadTabs`). The
  `ModuleFunctions` type was removed. Updated `CToWasmHelper.ts` and
  `JavaHelper.ts` to look up each module's `ModuleInfo` via
  `memoizedGetModuleManifestAsync` and pass `{ loadTabs: true }`.

- **Frozen `moduleContexts` proxy.** `createContext` now initializes
  `context.moduleContexts` as a lazy-init `Proxy` whose `get` trap
  writes to its target. Once Redux Toolkit (Immer) freezes the
  workspace state, any read of an unknown key — including
  `JSON.stringify` probing `toJSON` — throws
  `Cannot add property toJSON, object is not extensible`. This caused
  the entire `WorkspaceReducer.test.ts` suite (30 tests) to fail. The
  wrapper `createContext` in `JsSlangHelper.ts` now replaces
  `moduleContexts` with a plain `{}` after delegating to js-slang.
  *(Arguably a js-slang bug worth reporting upstream — the proxy
  shouldn't mutate its target, and certainly not for well-known method
  names like `toJSON`.)*

- **CSE machine: built-ins misclassified as source objects.** The
  js-slang PR added a `toReplString` method to built-in functions.
  `isSourceObject` in `CseMachineUtils.ts` checks
  `isObject(x) && 'toReplString' in x && isFunction(x.toReplString)`,
  but lodash `isObject` returns `true` for functions too. As a result,
  `pair`, `accumulate`, etc. were classified as source objects,
  rendered as `PrimitiveValue`, and no arrow was drawn to them from
  array units. Added a `!isFunction(x)` guard.

## Test plan

- [x] `yarn tsc -b` is clean
- [x] `yarn test` — all 644 tests pass (previously 31 failing)
- [ ] Manual smoke test: open a Source 4 playground, evaluate
      `list(1, pair, accumulate); debugger;` and verify the CSE
      machine draws arrows from the list cells to the global
      `pair`/`accumulate` functions
- [ ] Manual smoke test: run a C program that `#include`s a Source
      module (e.g. `<pix_n_flix>`) and verify the module functions
      still load
- [ ] Manual smoke test: run Java code that imports a Source module
      and verify the proxy functions work


RichDom2185 (changes requested):
Why are we adding typecasts?

martin-henz:
> Why are we adding typecasts?

You mean: https://github.com/source-academy/frontend/pull/3802/changes#diff-7b5e1d5740828c0dabd1c4d2f2ffeecd3ec6f01dc5ce7790809217b918da9b7aR88
as { [k: string]: Function | object } in JavaHelper.ts, right?

I think it's because of this change in js-slang:
src/modules/moduleTypes.ts:                                                                                                                     
                                                                                                                                                                       
  export type ModuleBundle = (require: RequireProvider) => ModuleFunctions;                                                                                                               
  export type ModuleFunctions = {                                                                                                                                                         
    [name: string]: any;
  };                                                                                                                                                                                      
                  
--->

  export type PartialSourceModule = (require: RequireProvider) => LoadedBundle;                                                                   
  export type LoadedBundle = {                                                                                                                                                            
    [name: string]: unknown;  
  };
  
The imported createModuleProxy function from java-slang takes { [k: string]: Function | object } as moduleFuncs. how can we satisfy this without a typecast?

RichDom2185 (approved):
LGTM, thanks!

RichDom2185:
Supersedes #3817

Issue #3796, opened 2026-04-20
CSE Machine: Corrupted layout table upon switiching to different code file


ThatLi:
A major visual error happens when user switched between code files in their folder and running them on the CSE Machine. The error persists until you reload the page.

### Code used:

1. foo
```
function foo(n) {
    const msg = "frame " + stringify(n);
}

for (let i = 0; i < 3; i = i + 1) {
    foo(i);
}
```
2. add_stream
```
function add_streams(s1, s2) {
    return pair(head(s1) + head(s2), 
           () => add_streams(stream_tail(s1), stream_tail(s2)));
}
const strm = pair(1, () => pair(2, () => add_streams(strm, strm)));
eval_stream(strm, 10);
```
--- 
Here is what happened when I run the add_stream program first, then switched to foo:

<img width="1284" height="734" alt="Image" src="https://github.com/user-attachments/assets/6227616b-1d3d-4307-a3d0-7c1bc7a089ac" />

> Intended foo layout

<img width="503" height="831" alt="Image" src="https://github.com/user-attachments/assets/4599056b-cf39-4597-ad99-2de762f181d5" />

> Corrupted foo layout

And vice versa:

<img width="629" height="839" alt="Image" src="https://github.com/user-attachments/assets/d9780838-1b2a-46be-bf63-d5b344fe2f99" />

> Corrupted add_stream layout

Akshay-2007-1:
#3824 fixes this bug. Can now be closed!

I have tested this and it is no more an issue!

PR #3824, opened 2026-05-02
Fix: resolve CSE Machine layout corruption when switching files in Folder Mode

gigopogo:
### Description

**Resolves:** #3796 

**Bug Context:**
A major visual bug occurs in the CSE Machine when users switch between different files in Folder Mode and evaluate them. The layouts from the old file and the new file overlap, corrupting the layout table until the page is fully reloaded.

This was caused by the `updateCse` boolean flag. In `handleEditorEval`, the layout caches were only being cleared if `updateCse === true`. However, simply switching active editor tabs does not trigger a code change event, meaning `updateCse` remains `false`. When the user pressed "Run" on the newly selected file, the `if (updateCse)` check bypassed `CseMachine.clearCachedLayouts()`, causing the new execution to draw using the old file's cached ghost coordinates.

**Changes Made:**
* **`Playground.tsx`**: Removed the `if (updateCse)` condition inside `handleEditorEval`. 
* `CseMachine.clearCachedLayouts()` now triggers on *every* evaluation. The tiny computational save of skipping a cache clear if the code hasn't changed is redundant and not worth the side effect of breaking Folder Mode.
* Removed `updateCse` from the `useMemo` dependency array for `autorunButtonHandlers` as it is no longer needed in that block.

**How to Test:**
1. Open the CSE Machine in Folder Mode.
2. Create two files with the code snippets from issue #3796 (`foo` and `add_stream`).
3. Run the first file.
4. Switch the active tab to the second file and hit Run.
5. The CSE Machine should now properly wipe the layout cache and render the new layout cleanly without any visual overlap.

### DIFF
--- a/src/pages/playground/Playground.tsx
+++ b/src/pages/playground/Playground.tsx
@@ -425,9 +425,7 @@ const Playground: React.FC<PlaygroundProps> = props => {
   const autorunButtonHandlers = useMemo(() => {
     return {
       handleEditorEval: () => {
-        if (updateCse) {
-          CseMachine.clearCachedLayouts();
-        }
+        CseMachine.clearCachedLayouts();
         // reset stepper before evaluation
         dispatch(WorkspaceActions.updateCurrentStep(-1, workspaceLocation));
         dispatch(WorkspaceActions.updateStepsTotal(0, workspaceLocation));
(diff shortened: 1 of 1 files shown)


RichDom2185 (approved):
LGTM, thanks. Agree we shouldn't need to over optimize, keeping it simple makes sense

PR #3857, opened 2026-05-14
Switch to Monaco Editor

RichDom2185:
### Type of change

### Checklist

- [ ] I have tested this code


leeyi45:
I was experimenting with migrating the devserver's editor, and basically the biggest hurdle now is that if we want autocomplete to work properly (or at least behave similarly to how it does right now with ace) we need to redo the entire autocompletion system.

There are other things like disabling the command palette or hiding irrelevant context menu options that need to be done (but the solutions for those are readily available).

PR #3903, opened 2026-05-29
Redirect to home tab on Stepper/CSE evaluation errors

### LINKED ISSUE #1737: Stepper: errors silently ignored
Some errors are silently ignored in stepper. Is validation happening before evaluation in the stepper? Examples of programs whose execution is prevented but no error is reported:
```
function f(x, x) {}
```
```
const x = 1; const x = 2;
```

Shrey5132:
Closes https://github.com/source-academy/js-slang/issues/1737

### Description

Redirects the user to the main introduction (home) tab when a preprocessing or evaluation error occurs during a Stepper or CSE Machine run. 

Currently, when code contains preprocessing/compile-time errors and is executed while the Stepper or CSE Machine tab is active, the visualizer remains on a blank welcome screen and does not display the error. This change ensures that any evaluation failure (`result.status !== 'finished'`) under these active views automatically triggers a redirect to the introduction tab (`SideContentType.introduction`), making the printed REPL/console error clearly visible to the user.

No dependencies are required for this change.

### Type of change

### How to test

1. Open the Playground.
2. Select either the Stepper tab or the CSE Machine tab.
3. Write a program containing a preprocessing or syntax error (for example, `const x = 1; const x = 2;` or `funyun f(x) {}`).
4. Click **Run**.
5. Verify that the side content tab automatically switches back to the introduction (home) tab and the error is printed to the console.

### Checklist

- [x] I have tested this code


Akshay-2007-1 (commented):
The code does solve the targeted fix as it states [here](https://github.com/source-academy/js-slang/issues/1737). 

But when the conductor feature flag is turned on, say, and we type random gibberish in the pre-existing but non-functional (as of today) Stepper tab, it should cause errors and redirect to the home tab. But this doesnt happen.

So, maybe a similar minor change required even when the Conductor flag is turned on. This will be good QOL later on because eventually we will only be using the new framework after the migration of the plugins is completed!

AaravMalani:
> The code does solve the targeted fix as it states [here](https://github.com/source-academy/js-slang/issues/1737).
> 
> But when the conductor feature flag is turned on, say, and we type random gibberish in the pre-existing but non-functional (as of today) Stepper tab, it should cause errors and redirect to the home tab. But this doesnt happen.
> 
> So, maybe a similar minor change required even when the Conductor flag is turned on. This will be good QOL later on because eventually we will only be using the new framework after the migration of the plugins is completed!

Ideally, the CSE and Stepper should not show up unless the language provides it. For the current languages, should the viz tools show up, albeit non-functional?

Akshay-2007-1:
> Ideally, the CSE and Stepper should not show up unless the language provides it. For the current languages, should the viz tools show up, albeit non-functional?

# In the normal framework

In `Playground.tsx` (line 735):
```ts
const shouldShowCseMachine = languageConfig.supports.cseMachine;
```

`languageConfig` is an `SALanguage` object from the old `ApplicationTypes.ts` system. It has a `supports: { cseMachine: boolean }` field. The `sourceLanguages` array in `ApplicationTypes.ts` sets `cseMachine: true` for Source 3+. Java is manually set to `cseMachine: true`. Python has **no `SALanguage` entry at all**.

# In the conductor framework

When `conductor.enable` is ON and you select a language from the Language Directory dropdown:
```ts
dispatch(LanguageDirectoryActions.setSelectedLanguage('python'))
    → LanguageDirectorySaga: setSelectedLanguage handler
        → sets selectedEvaluatorId
        → calls preloadConductorEvaluatorSaga(evaluator.path)
        → DONE — nothing else happens
```

`playgroundConfigLanguage` is **never dispatched** when selecting from the Language Directory. So `state.playground.languageConfig` never updates. `shouldShowCseMachine` stays whatever it was from the last chapter the old selector set (probably Source 1, which has `cseMachine: false`).

The Language Directory has `capabilities: EvaluatorCapability[]` on every evaluator entry - exactly the right place to declare "this evaluator supports the CSE Machine". But:

```ts
// @sourceacademy/language-directory
declare enum EvaluatorCapability {
    // EMPTY — exactly like CseMessageType
}
```

All current evaluator entries have `"capabilities": []`. These most probably need to be filled in. So, what we can do is only fill in the plugins specifically into those capability arrays of those languages that support them!

RichDom2185 (approved):
LGTM, thanks!

PR #3921, opened 2026-06-01
CSE Machine : Arrow Routing QOL

### LINKED ISSUE #3919: CSE Machine: arrows run through boxes from which they originate
This PR 
https://github.com/source-academy/frontend/pull/3899
introduced the following problem: 

Image

(example given the the PR already)

The arrow should not run through the box, and instead should run something like this:

Image

### LINKED ISSUE #3920: CSE Machine : Arrows run over function objects
A while ago #3722 introduced Arrow Layering whereby arrows were in a different Konva layer going behind the frames, and objects.

But recently, despite that happening for the most part, some arrows go over the function eyeballs.

Image

An easy fix to include these function closure objects as well in that same Konva layer!

Akshay-2007-1:
### Description

Closes #3920 
Closes #3919 

The routing logic for arrows originating from array cells was producing paths that either pierced target objects, ran along cell borders, or got hidden behind frames and other canvas elements (arrows are rendered as underlays). This PR rewrites `calculateSteps` in `ArrowFromArrayUnit` to route arrows correctly based on the spatial relationship between source and target.

---

## Before / After

Some examples are as follows :

| Case | Before | After |
|---|---|---|
| Array cell → closure (from right) | image | image |
| Array cell → array (target below) | image | image |
| Array cell → array (target above) | image | image |
| Arrow running over Closure/FnObject | image | image |

### Type of change

### How to test

```javascript
const y = 2;
const f = () => null;
const q = pair(y, f);

{
    const x = 1;
    const p = pair(x, 2);
    set_tail(p, q);
}
```

---

```javascript
// Test B: sibling frames with closures + nested arrays
function make_frame(id, n) {
  const arr = [];
  let i = 0;
  while (i < n) {
    const k = id * 100 + i;
    function add(x) {
      return x + k;
    }
    function combo(x) {
      return add(x) + k;
    }
    arr[i] = [add, combo, [list(add, combo), [i, i + 1, i + 2]]];
    i = i + 1;
  }
  return () => arr;
}

const frames = [];
let j = 0;
while (j < 4) {
  frames[j] = make_frame(j, 4);
  j = j + 1;
}

frames[0]();
frames[1]();
frames[2]();
frames[3]();
```

The above is a more complicated code with lots of frames and steps with possibility of catching bugs in the logic.

---

### Checklist

- [X] I have tested this code


RichDom2185 (approved):
LGTM, thanks!

PR #3959, opened 2026-06-10
feat(stepper): integrate Conductor plugin tabs and dynamic loading of web plugins

Shrey5132:
This PR integrates the dynamic plugin loading framework for the Conductor stepper plugin. Key changes include:
- Importing and registering external web plugins dynamically using `BrowserHostPlugin`.
- Exposing shared dependencies globally via a bootstrap script (`conductorSharedDeps`).
- Injecting custom stepper plugin tabs into the SideContent of the Playground when Conductor is enabled.
- Hiding legacy stepper tabs/resizing when the Conductor stepper is active to match legacy behaviors.
- Adding a one-time, generic import map + `public/shims/*` so dynamically-imported web-plugin bundles share the frontend's single React tree and Blueprint styling.

### A note on evaluators and plugins (re: review feedback)

The evaluator bundles and the stepper web-plugin bundle are **not** committed to the frontend. They are built from their source repos (js-slang, py-slang, the plugins monorepo) and deployed to GitHub Pages, then referenced via the hosted language- and plugin-directories. The relevant feature flags already default to those hosted directories:

- `directory.language.url` → `https://source-academy.github.io/language-directory/directory.json`
- `directory.plugin.url` → `https://source-academy.github.io/plugin-directory/directory.json`

so the frontend never needs the files locally. The local copies under `public/evaluators`, `public/languages` and `public/plugins` were only a turnkey local-test harness; they have been untracked and gitignored (`dev.sh` still copies them for local testing without re-adding them to the repo).


src/pages/playground/Playground.tsx
```diff
@@ -957,7 +976,10 @@ function Playground(props: PlaygroundProps) {
     sourceVariant: languageConfig.variant,
     externalLibrary: ExternalLibraryName.NONE, // temporary placeholder as we phase out libraries
     hidden:
-      selectedTab === SideContentType.substVisualizer || selectedTab === SideContentType.cseMachine,
+      selectedTab === SideContentType.substVisualizer ||
+      selectedTab === SideContentType.cseMachine ||
+      // When the conductor stepper plugin tab is active, also hide the REPL (matches legacy behaviour)
+      (conductorEnabled && (selectedTab as string) === 'stepper'),
```
AaravMalani: Could you abstract this out so other tabs can use this as well? I was going to implement this feature before seeing there's already a PR open for this

public/evaluators/py-slang/PyCseEvaluator2.js
AaravMalani, line 1: Do the evaluators need to be added to the frontend?

Akshay-2007-1: @Shrey5132 I was confused bout this as well. They can be built from js-slang just like how py-slang does it. 

py-slang CI:

1. Builds all evaluators (yarn run build --all) on every push to main
2. Uploads the dist/ folder as a GitHub Pages artifact
3. Deploys it to GitHub Pages

So the evaluator .js files live at a GitHub Pages URL like https://source-academy.github.io/py-slang/PyCseEvaluator1.js. The frontend's language-directory (or local-directory.json) references those URLs — it doesn't need the files locally at all.

Shrey5132: Hi, this issue is resolved in the latest commit.

Akshay-2007-1 (changes requested):
1. The portal: entries must be reverted, conductor should stay at npm:^0.5.0 (or bumped to a new published version if you made breaking changes to conductor), and plugin-directory should go to the proper #0.0.3 git tag (which is the intended change in package.json)

2. If you did change conductor's API, you need to publish a new npm version of conductor first, then reference that

Akshay-2007-1 (approved):
SideContentManager hardcoded to 'playground'. If there are snippets (which there will be, in SICPy textbook), the Stepper might not work!
Please check this out, @Shrey5132!

LGTM otherwise!

Shrey5132:
> comment

Thanks, I'll look into it.
Might post it as an issue to be fixed in another PR, in case this one gets merged before that.

PR #35, opened 2018-05-25
Add Source §2

ningyuansg:
This PR is blocked by #22, as the changes in `createContext.ts` here follow that in said PR.

---

### Features
- Add §2 (list library and the empty list, `[]`)
- Fix an issue where arrow functions were not treated as functions (`fb954f3`)
- Add HTML select for selecting chapters

### Issues fixed
- #18 
- #32

Issue #33, opened 2018-05-24
REPL buttons should be in a button group


remo5000:
As of now, they are `col`s and thus do not stick together when there are size changes.

remo5000:
Addressed in #21

PR #3996, opened 2026-06-22
fix: bump conductor to 0.5.0

### LINKED ISSUE #34: run button not working as intended when the conductor feature flag is enabled
When the conductor feature flag is enabled, clicking on the run button with a valid/invalid program in the editor behaves unexpectedly:

The program does not always produce an output when the run button is clicked. There are some scenarios where I am able to click on the run button multiple times after which the run button stops registering clicks altogether, and some scenarios where clicking on the run button hangs the page (may be an infinite loop), requiring a refresh.

These issues prevent me from testing changes to the python stepper, since unless the run button works, no tool which relies on the program running works either.

To test these issues yourself, set the language to Python 1 or 2, although I believe the issue exists regardless of the language selected.

Akshay-2007-1:
## Summary
- Bumps `@sourceacademy/conductor` from `0.4.0` to `0.5.0`

Closes [issue](https://github.com/source-academy/conductor/issues/34)

## Why
Conductor 0.4.0 used `const enum` for `ServiceMessageType`, `InternalChannelName`, and other enums. Vite/esbuild (used by this frontend) compiles with `isolatedModules`, which erases `const enum` values from external packages to `undefined` at runtime. This caused the Conductor protocol handshake to silently fail — making the **Run button appear to do nothing**.

Conductor 0.5.0 converts all exported `const enum`s to regular `enum`s, fixing the runtime erasure. It also adds `BasicHostPlugin.receiveResult` so evaluation results are correctly received by the host.

## Changes in Conductor 0.5.0 (vs 0.4.0)
- `fix: remove const keyword from all remaining exported enums (#35)`
- `feat(host): add receiveResult callback to BasicHostPlugin (#33)`
- `fix(stepper): handle evaluator errors gracefully in BasicEvaluator and RunnerPlugin (#30)`
- `Modules: Switch Module Closures to be AsyncGenerators (#29)`

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@sentry/react": "^10.5.0",
     "@sourceacademy/autocomplete": "github:source-academy/autocomplete#e669d9ed98753350a3c8433a92985227eb789663",
     "@sourceacademy/c-slang": "^1.0.21",
-    "@sourceacademy/conductor": "https://github.com/source-academy/conductor.git#0.4.0",
+    "@sourceacademy/conductor": "https://github.com/source-academy/conductor.git#0.5.0",
     "@sourceacademy/language-directory": "https://github.com/source-academy/language-directory.git#0.0.6",
     "@sourceacademy/plugin-directory": "https://github.com/source-academy/plugin-directory.git#0.0.2",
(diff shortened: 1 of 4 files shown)


martin-henz (approved):
looks good

PR #4000, opened 2026-06-23
feat(cse): Python CSE Machine via Conductor

Akshay-2007-1:
## Summary

- Wires the Conductor framework's CSE plugin into the frontend so Python 3 and 4 (and any future conductor evaluator declaring `"cse"` capability) can drive the existing CSE Machine visualizer
- Introduces a language-agnostic `CseSnapshot` protocol and `CseSnapshotAdapter` that translates snapshots into the Konva renderer's expected shape without touching Source-specific paths
- Adds tab visibility logic: CSE tab appears only for evaluators with `"cse"` capability; hides for Python 1/2 and Source
- Fixes several interaction bugs: animation ordering, slider behaviour in snapshot mode, tab flush on switch, home-tab isolation

## Key changes

**New files**
- `src/features/conductor/CseMachineHostPlugin.ts` — concrete subclass of the abstract npm plugin; makes `receiveSnapshots` reassignable for saga wiring
- `src/features/conductor/CseSnapshotAdapter.ts` — translates `CseSnapshot[]` → fake js-slang AST nodes for the Konva renderer
- `src/features/cseMachine/CseMachineLayout.snapshotMode.ts` — snapshot rendering entry point that bypasses Source-specific env mutations

**Modified**
- `conductorEvaluatorCache.ts` — threads `csePlugin` through `PreparedConductor`
- `evalCode.ts` — forks `handleCseSnapshots` saga; wires `csePlugin.receiveSnapshots` to a Redux eventChannel
- `SideContentCseMachine.tsx` — snapshot-mode rendering path; tab flush/isolation via `isOnCseTab` from Redux; animation fix; arrow filter additions
- `Playground.tsx` — capability-based `shouldShowCseMachine`; passes `isActive` to CSE tab
- `CseMachineAnimation.tsx`, `CseMachineLayout.tsx`, `CseMachineUtils.ts` — ported fixes from `CSE_liveness_frontend`
- `public/local-directory.json` — Python 3/4 default evaluators declare `"cse"` capability; URLs point to py-slang GitHub Pages
- `package.json` / `.yarnrc.yml` — `@sourceacademy/web-cse-machine` switched from local portal to npm `^1.0.0`

## Dependencies

| Package | Status |
|---|---|
| `@sourceacademy/web-cse-machine` | npm `^1.0.0` (published) |
| `@sourceacademy/conductor` | git tag `#0.5.0` (public) |
| `@sourceacademy/language-directory` | git tag `#0.0.6` (public) |
| `PyCseEvaluator{1-4}.js` | py-slang GitHub Pages — **active after py-slang PR merges** |

## Test plan

- [ ] Python 3 / Python 4 with Default evaluator: CSE Machine tab appears, run code, step through slider, animations play
- [ ] Python 1 / Python 2 / Source: CSE Machine tab does not appear
- [ ] Run code from home tab → switch to CSE tab: standard template shown (not visualization)
- [ ] Run code from CSE tab → switch away → switch back: standard template shown
- [ ] Java CSE machine unaffected
- [ ] Source CSE machine (stepper) unaffected


src/features/conductor/CseMachineHostPlugin.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Concrete subclass of the abstract CseMachineHostPlugin from @sourceacademy/web-cse-machine.
+ *
+ * The abstract base handles channel transport (subscribes to the __cse channel, calls
+ * receiveSnapshots). This subclass makes receiveSnapshots reassignable so evalCode.ts
+ * can wire it to a Redux eventChannel after construction.
+ */
+import {
+  CseMachineHostPlugin as _CseMachineHostPluginBase,
+  type CseSnapshot,
+} from '@sourceacademy/web-cse-machine';
```
Akshay-2007-1, line 11: This is the CSE Machine web side Plugin from plugins repo!

src/features/conductor/createConductor.ts
```diff
@@ -3,15 +3,17 @@ import { Conduit } from '@sourceacademy/conductor/conduit';
...
   onRequestFile: (fileName: string) => Promise<string | undefined>,
   onRequestLoadPlugin: (pluginName: string) => void,
-): { hostPlugin: BrowserHostPlugin; conduit: IConduit } {
+): { hostPlugin: BrowserHostPlugin; csePlugin: CseMachineHostPlugin; conduit: IConduit } {
   const worker = new Worker(evaluatorPath);
   const conduit = new Conduit(worker, true);
   const hostPlugin = conduit.registerPlugin(BrowserHostPlugin, onRequestFile, onRequestLoadPlugin);
   hostPlugin.registerPlugin(AutoCompletePlugin);
-  return { hostPlugin, conduit };
+  const csePlugin = conduit.registerPlugin(CseMachineHostPlugin);
```
AaravMalani, line 17: I wouldn't recommend this, I think a better mechanism is for the CSE runner-side plugin (or the runner when it loads the CSE runner-side plugin, I don't think plugins have access to the BasicRunnerPlugin instance) to load the host-side plugin.

(Yes, the autocomplete plugin should migrate to that as well and I'll make a PR for that)

The issue is
a) plugins are loaded unnecessarily which will cause unnecessary code execution every worker load
b) eventually, if we have 10 plugins even, the object returned by the `createConductor` call is going to become very large

src/features/cseMachine/CseMachine.tsx
```diff
@@ -59,6 +61,17 @@ export default class CseMachine {
   private static currentEnvId: string;
   private static control: Control | undefined;
   private static stash: Stash | undefined;
+  private static pairCreationMode: boolean = false;
+  /** Last snapshot passed to renderSnapshot(); used by redraw() in snapshot mode. */
+  private static lastSnapshot: CseSnapshot | null = null;
+
+  public static togglePairCreationMode(): void {
```
AaravMalani, line 69: Could there be some documentation regarding what `pairCreationMode` does? Not a blocker, I'll approve even if this isn't fixed, but I'm not fully sure what it does

src/features/cseMachine/CseMachineAnimation.tsx
```diff
@@ -125,22 +133,48 @@ export class CseAnimation {
           new ControlToStashAnimation(lastControlComponent, currStashComponent!),
         );
         break;
-      case 'Identifier':
+      case 'Identifier': {
+        const identNode = node as any;
```
AaravMalani: I'd prefer proper type validation here
(I've not cloned it locally yet, why is the widening required?)

src/commons/sideContent/content/SideContentCseMachine.tsx
AaravMalani, line 1: What's with the formatting changes in this file out of curiosity? (Not a blocker, again)

src/features/cseMachine/CseSnapshotAdapter.ts
```diff
@@ -0,0 +1,450 @@
...
+
+function toJsValue(
+  v: CseSerializedValue,
+  envMap: Map<string, Environment>,
+  closureCache: Map<string, unknown>,
+  listCache: Map<string | number, unknown>,
+): unknown {
+  const label = v.label.toLowerCase();
+
+  if (label === 'int' || label === 'number') {
```
AaravMalani, line 53: Shouldn't we migrate the JS value conversion to the runner side, since the frontend is supposed to be language agnostic. This would work for JS and Python, but for languages that don't follow this specification, it wouldn't work. For example, Go uses `nil` instead of `null`, `none` or `nonetype`. Even if the label matches, the display values might not conform to the subset present here. For example, Racket uses `#t` and `#f` instead of `true` and `false`

AaravMalani (approved):
For the sake of time, and since I don't see any fatal issues, I'm approving the PR. However, if there is some extra time, I do think the design of the adapter needs to be changed (which I think will happen when the code is migrated out of the frontend).

Either way, good work!

src/features/conductor/createConductor.ts
Akshay-2007-1: I have made this an issue at #4004. Major work for another PR. Out of scope for now ig. 
Thanks for pointing out!

src/features/cseMachine/CseSnapshotAdapter.ts
Akshay-2007-1: I have made this an issue at #4005. Major work for another PR. Out of scope for now ig. 
Thanks for pointing out!

src/commons/sideContent/content/SideContentCseMachine.tsx
Akshay-2007-1: The large diff here is unavoidable, this file was refactored into a function component in a prior commit, but our PR needs it as a class component (we rely on instance methods and setState callbacks for animation timing). Restoring it to a class component naturally produces a large diff even though the logic is identical to what was there before.

src/features/cseMachine/CseMachineAnimation.tsx
Akshay-2007-1: fixed

src/features/cseMachine/CseMachine.tsx
Akshay-2007-1: added!

PR #4013, opened 2026-06-29
chore: bump language-directory to 0.0.10 (§ signs in chapter names)

Akshay-2007-1:
## Problem

The § signs added by Martin Henz in language-directory#30 were not showing up in the frontend. The language-directory was tagged `0.0.9` _before_ that PR landed, so three commits were stranded on `main` with no released tag:

- `cda61b04` feat(stepper): register python1Stepper in Python 1
- `ee2701a9` **adding chapter sign (§) in front of Python and Source chapter numbers**
- `7a9b32d7` feat(python): add Python §1 and §2 Stepper evaluators

The root cause is a short-circuit in `LanguageDirectorySaga.ts`: when the feature flag URL equals the default (`source-academy.github.io/language-directory/directory.json`), the saga skips the network fetch and uses the locally-installed npm package directly. That package was pinned to `0.0.9` → no § signs.

## Fix

- Tagged `0.0.10` on language-directory `main` HEAD (`7a9b32d7`)
- Bumped `package.json` from `#0.0.9` → `#0.0.10`
- Re-ran `yarn install` to update `yarn.lock`

After this change the installed package exports `"Source §1"`, `"Python §1"`, etc.

## Test plan
- [ ] Language selector shows "Source §1" … "Source §4" and "Python §1" … "Python §4"

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@sourceacademy/autocomplete": "github:source-academy/autocomplete#e669d9ed98753350a3c8433a92985227eb789663",
     "@sourceacademy/c-slang": "^1.0.21",
     "@sourceacademy/conductor": "^0.5.0",
-    "@sourceacademy/language-directory": "https://github.com/source-academy/language-directory.git#0.0.9",
+    "@sourceacademy/language-directory": "https://github.com/source-academy/language-directory.git#0.0.10",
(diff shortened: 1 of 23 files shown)


Akshay-2007-1:
Superseded by #4014 — the branch was accidentally cut from a local master commit that included unrelated SCSS changes. #4014 is a clean 2-file diff (package.json + yarn.lock only).

PR #4035, opened 2026-07-01
feat(playground): drive the conductor stepper from its side-content tab

Shrey5132:
## Summary
Improves the conductor Python stepper UX in the playground:

- **Tab-driven evaluator.** A Stepper tab is shown for any conductor language that offers stepping (Python §1/§2, via the new `stepper` evaluator capability). Opening the tab selects the stepper evaluator so a Run produces steps; leaving it restores the language's default evaluator.
- **Stepper hidden from the dropdown.** Stepper-capability evaluators are filtered out of the evaluator dropdown, and the dropdown is disabled while the Stepper tab is active — so the stepper is reachable only through its tab.
- **No more silent failures.** Conductor evaluation errors that arrive by rejection (e.g. the Python stepper's preprocessing/syntax errors, delivered into the run saga rather than over the error channel) are now surfaced: appended to the REPL and, since the REPL is hidden on the Stepper tab, the view switches to the Introduction tab so the error is visible.
- Side-content tabs are de-duplicated by rendered id (so the placeholder Stepper tab and the plugin's live one never double up), and pending result/error/output messages are drained before teardown so a trailing message is not dropped.

## Test plan
- `tsc` clean; `Playground`, `WorkspaceSaga`, `PlaygroundSaga`, and `sideContent` suites pass.
- Verified in-browser: on the Python §1 Stepper tab, running a program with an undefined name jumps to Introduction with the `NameError` shown in the REPL; the evaluator dropdown is greyed while the Stepper tab is active.

Companion changes: `source-academy/language-directory`, `source-academy/py-slang` (feat/conductor-stepper).


Akshay-2007-1 (approved):
I have personally cloned, built and tested the changes of this PR. The change where Stepper evaluator is removed from the dropdown as well as choosing it by default when in the Stepper tab is a commendable required change!

One thing that is quite different from, say the CSE tab which is quite similar in terms of a plugin is that on the first time selection of that tab when in Python chapters 1 and 2, the Stepper tab loads for quite a long time untile reload.

Image

---

I dont know if this will still be there in production but its worth a look!

PR #4063, opened 2026-07-04
fix(playground): reset side content tab to Introduction on language switch

### LINKED ISSUE #4061: Python : Stepper Tab loading forever
After the PR #4035 got merged, as I had already stated in the approval of the same PR, there is a bug where Stepper tab loads forever and hangs at the following status:

Image

### Steps to Reproduce : 
be on the Stepper tab
then switch to Python 2 if you are on 1 
or switch to 1 if you are on 2

---

Stepper only works and the tab appears when on Python chapters 1 and 2.

Shrey5132:
## Summary
- Fixes #4061 (Stepper tab loading forever after switching Python §1 ⇌ §2).
- Whenever the Conductor directory's selected language changes, the side content tab now unconditionally resets to Introduction, regardless of which tab was active.
- Not stepper-specific: a tab tied to the previous language (a plugin-provided one, or one only some languages offer, like the Stepper) has no guarantee of applying to the new language, so preserving it across a switch risks stranding the user on a tab that's no longer valid — exactly what happened in #4061.
- Replaces the previous, narrower fix in the Stepper-evaluator-sync effect (which only handled switching *away* from a stepper-capable language, missing the reported case of switching *between* two stepper-capable languages, e.g. Python §1 → §2).
- The legacy Source/full-JS/Java/C chapter+variant picker already resets the tab correctly today (its `chapterSelect` saga dispatches `resetSideContent` whenever the chapter or variant changes), so this change only needed to cover the Conductor language-directory picker, which had no equivalent.

## Test plan
- Added a regression test reproducing the exact repro steps from #4061: land on the Stepper tab under Python §1, switch to Python §2, assert the tab resets to Introduction. Verified it fails without the fix (`expected 'stepper' to be 'introduction'`) and passes with it.
- Full `Playground.test.tsx` suite (4 tests) and related suites (playground/sideContent/controlBar/directory/sagas, 87 tests) pass.

### DIFF
--- a/src/pages/playground/Playground.test.tsx
+++ b/src/pages/playground/Playground.test.tsx
@@ -11,7 +11,10 @@ import {
   type OverallState,
 } from 'src/commons/application/ApplicationTypes';
 import type { Router } from 'src/commons/application/types/CommonsTypes';
+import { visitSideContent } from 'src/commons/sideContent/SideContentActions';
+import { SideContentType } from 'src/commons/sideContent/SideContentTypes';
 import { EditorBinding, WorkspaceSettingsContext } from 'src/commons/WorkspaceSettingsContext';
+import LanguageDirectoryActions from 'src/features/directory/LanguageDirectoryActions';
 import { createStore } from 'src/pages/createStore';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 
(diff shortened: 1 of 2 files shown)


Akshay-2007-1 (changes requested):
I notice that the fix of switching to Intro tab on switching chosen language is not specific to the Python chapters 1 and 2 which are Stepper specific. Thats a good regression safety feature beyond the issue this PR fixes.

Could you also fix the issue of switching to Intro tab on running invalid code irrespective of what tab we are in for all Python chapters in the same PR? Its quite a similar process and changes are required in the very same files this PR touches.

Akshay-2007-1 (approved):
The CSE Machine issue request I made got its own PR. 
LGTM, approving!

PR #4065, opened 2026-07-04
fix(conductor): stop conductor.enable from overriding directory URLs

Shrey5132:
## Summary
- In Settings → Feature Flags, enabling `conductor.enable` was silently force-setting `directory.language.url` to `/languages/directory.json` and `directory.plugin.url` to `/plugins/directory.json`; disabling it reset both. This was a local-dev convenience (one click to point the directories at the frontend's own dev-served JSON) that leaked into production, where those paths don't serve anything meaningful.
- `flagConductorEnable`'s `createFeatureFlag` call took a 4th argument — a saga callback — that did this. Removed the callback entirely; the flag is now a plain boolean with no side effects on any other flag. Toggling it now only affects the Conductor framework itself, matching the intended behaviour.
- `directory.language.url`/`directory.plugin.url` already default to the correct production remote URLs on their own, so nothing else needed to change.

## Test plan
- Added a regression test asserting that enabling/disabling `conductor.enable` triggers no callback at all. Verified it fails against the old code (the callback ran and returned a generator, not `undefined`) and passes against the fix.

### DIFF
--- a/src/features/conductor/flagConductorEnable.test.ts
+++ b/src/features/conductor/flagConductorEnable.test.ts
@@ -1,0 +1,13 @@
+import { describe, expect, test } from 'vitest';
+
+import { flagConductorEnable } from './flagConductorEnable';
+
+describe('flagConductorEnable', () => {
+  test('enabling or disabling it has no side effect on other flags (e.g. the directory URLs)', () => {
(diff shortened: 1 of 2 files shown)


Akshay-2007-1 (approved):
LGTM! Quite an urgent PR tbh, we shouldnt have such local stuff merge into prod.

PR #4070, opened 2026-07-05
docs: add local backend setup and py-slang testing instructions

Akshay-2007-1:
## Summary
- The README's "Backend configuration" section explains what each `REACT_APP_*_BACKEND_URL` env var means, but never explains how to actually get a backend running locally — without one, the frontend just shows a "Source Academy is under maintenance" page. Adds a "Running a local backend" subsection with the full clone-to-running sequence (asdf toolchain, postgres via Docker, dev secrets, `mix ecto.setup`, `mix phx.server`).
  - Also corrects the Docker command copied from the backend repo's own README: it has a stray `-e` flag placed right before `-p 5432:5432`, which causes `-p` to be consumed as `-e`'s value instead of being parsed as its own flag — so the port never actually gets published if run verbatim.
- Adds a "Running your own py-slang" section to `CONTRIBUTING.md`, mirroring the existing "Running your own js-slang" section. Links out to py-slang's own README (source-academy/py-slang#199), which now documents the workflow, rather than duplicating it here — consistent with how the js-slang section is written.

(Originally two separate PRs — #4071 folded in here via cherry-pick so only one review is needed. #4071 is now closed.)

## Test plan
- [x] Docs-only change, verified against the backend's actual router/config/README and cross-checked against a working local backend setup.

PR #4071, opened 2026-07-05
docs: add "Running your own py-slang" section to CONTRIBUTING.md

Akshay-2007-1:
## Summary
- Mirrors the existing "Running your own js-slang" section in `CONTRIBUTING.md`.
- Links out to py-slang's own README, which now documents this workflow (companion PR: source-academy/py-slang#199) — consistent with how the js-slang section links out rather than duplicating instructions.
- Notes the one key difference contributors should know upfront: py-slang is loaded at runtime via a URL from the Language Directory, not as a build-time npm dependency like js-slang, so the local-testing mechanics differ.

## Test plan
- [x] Docs-only change.


Akshay-2007-1:
Folded into #4070 via cherry-pick so only one review is needed. Closing this one.

PR #4085, opened 2026-07-11
Configure Louis chatbot for the Python SICP textbook

Isha-Sovasaria:
### What changed

1. Move the SICP chatbot (Louis) from the JavaScript textbook layout to the Python textbook layout.
2. Gate Louis behind a new per-course enableLouisChatbot configuration flag.
3. Add an admin-panel toggle and a configurable louisChatbotPrompt system-prompt field.
4. Pass the configured Louis system prompt through initChat and continueChat.


RichDom2185:
@Isha-Sovasaria I'm marking as draft since there are many merge conflicts. Please re-mark as ready when ready

Chatbot_Architecture_Louis_Pixel.pdf
RichDom2185, line 1: Why is this committed?

Chatbot_Architecture_Louis_Pixel.md
RichDom2185, line 1: Why is this committed?

src/new_routes/sicpjs/_layout.tsx
```diff
@@ -105,9 +76,7 @@ function SicpLayout() {
           )}
         </CodeSnippetProvider>
       </SicpErrorBoundary>
-      {isLoggedIn && Constants.featureFlags.enableSicpChatbot && (
-        <Chatbot getSection={getSection} getText={getText} />
-      )}
+      {/* Chatbot moved to Python layout */}
```
RichDom2185, line 79: Why is it removed?

src/new_routes/sicpjs/_layout.tsx
```diff
@@ -19,42 +16,16 @@ import {
...
-
-function getSection() {
-  // To discard the '/sicpjs/'
-  return location.pathname.replace('/sicpjs/', '') as SicpSection;
-}
+// Chatbot utilities removed from JS layout
 
 function SicpLayout() {
-  const { section } = useParams<{ section: string }>();
+  const { section } = useParams();
```
RichDom2185: Why did we remove the typing? This breaks type safety?

RichDom2185 (changes requested):

src/new_routes/sicpjs/_layout.tsx
Isha-Sovasaria: It is removed because Prof Martin said to keep the chatbot visible only for python versions of the books and it shouldnt be visible on the js side

RichDom2185: I see, in that case, hiding it for JS does not necessarily mean removing the feature. We should put it under a feature flag instead, similar to the Louis chatbot, but gated by language?

Isha-Sovasaria: I haven't removed the feature.The original frontend had this React Component in the layout for the js textbook.Now I added that to the python textbook's layout.However, if I choose to do a feature flag-i feel its unnecessary because there is no Louis support for js -the RAG pipeline is purely for python book in the backend.And then we would have the same code in two layouts (code duplication) with a guarantee that it never will be switched on for js -just to keep this bit of code in this place.I think best to delete from here and let it exist In the correct place-python layout

martin-henz (approved):
multiple thorough review rounds

sayomaki:
@coderabbitai review

sayomaki:
Some important pointers that I have noticed:
1. The `enableLouisChatbot` config is only checked in the frontend, and not on the backend. You can easily think of how this is bad: the course disables the chatbot, but users can bypass the frontend check and still be able to send crafted louis requests to the backend directly.
2. It appears that the entire prompt is sent from the frontend to the backend, and that could mean that any user in a valid course with louis enabled is able to get around the frontend by crafting a request with a custom prompt, bypassing any restrictions or limitations. The prompt should only be ever provided by the backend, and stored in the backend.

    Alternatively, you can consider attaching the user "prompt" as a message before the actual user message (you can send more than one message in a request), and have a safe system prompt in the backend that we have control over. Or you can just prepend the prompt to the user message, depending on what it is being used for (feel free to correct me if this is not the intent, but I assuming it the first one in the above paragraph)

PR #4152, opened 2026-07-22
fix: encode Python language selection in share links

Akshay-2007-1:
## Summary
- Share links only encoded the js-slang `Chapter`/`Variant`, so a Conductor-based language selection (currently Python) was silently dropped when generating a share link.
- Adds a `lang` query param carrying `languageDirectory.selectedLanguageId` in `PlaygroundSaga.updateQueryString`, and restores it via `LanguageDirectoryActions.setSelectedLanguage` in `Playground.handleHash` when present.

Fixes #4144

## Test plan
- [x] `yarn tsc --noEmit` passes
- [ ] Manually verify: select Python in the playground, generate a share link, open it in a new tab, confirm Python is selected
- [ ] Manually verify: existing Source share links (chap/variant) still restore correctly

PR #4153, opened 2026-07-22
fix: stop legacy chap param from shadowing Python share links

Akshay-2007-1:
## Summary
- Follow-up to #4152. That PR added `lang`/`evaluator` share-link params for Conductor-based languages (e.g. Python), but `handleHash` never actually reached that decode path: `playground.context.chapter` always holds a real js-slang `Chapter` value (defaults to Source 1) regardless of whether Conductor is active, so `chap` was always present and truthy in the URL, and the `if (chapter) {...} else if (qs.lang)` branch always took the legacy path.
- `updateQueryString` now only encodes `chap`/`variant` when Conductor is disabled, and only encodes `lang`/`evaluator` when it's enabled — the two are mutually exclusive, so this lets `handleHash` reach the correct branch on decode.

## Test plan
- [x] `yarn tsc --noEmit` passes
- [x] `yarn oxfmt --list-different` passes
- [ ] Manually verify: with `conductor.enable` on, select Python + a non-default evaluator (e.g. PVML), share, reopen in a new tab in the same browser — language and evaluator now correctly restore instead of defaulting
- [ ] Manually verify: regular Source chapter share links still restore chapter/variant correctly

### DIFF
--- a/src/commons/sagas/LanguageDirectorySaga.ts
+++ b/src/commons/sagas/LanguageDirectorySaga.ts
@@ -43,9 +43,21 @@ export function* getEvaluatorDefinitionSaga() {
 const languageDirectoryHandlers = combineSagaHandlers({
   [LanguageDirectoryActions.setLanguages.type]: function* () {
     const directory = yield select((state: OverallState) => state.languageDirectory);
-    if (directory.languages.length > 0) {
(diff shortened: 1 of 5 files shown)


martin-henz (approved):
looks good! two separate mechanisms

PR #59, opened 2018-05-29
Organise utilities

ningyuansg:
### Features
- Organise `notification.ts` and `registerServiceWorker.ts` into their own utility directory
- Slight refactor of name `genericButton` -> `controlButton` according to PR #35 
- Remove unnecessary `yield undefined` according to issue #37 
- Slight refactor in structure of `ReplControl.tsx` to list declarations in decreasing levels of abstraction
- Improve test coverage of `Repl.tsx` from 67% to 100%

```
------------------------------|----------|----------|----------|----------|-------------------|
File                          |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------------------|----------|----------|----------|----------|-------------------|
src/components/workspace      |     87.5 |    58.33 |    33.33 |    87.27 |                   |
  Repl.tsx (this PR)          |      100 |      100 |      100 |      100 |                   |
 src/components/workspace     |    70.69 |    29.17 |    33.33 |    70.18 |                   |
  Repl.tsx (master: 403ba81)  |    68.18 |    22.22 |      100 |    66.67 |... 66,74,75,81,90 |
------------------------------|----------|----------|----------|----------|-------------------|
```

### Issues fixed

- #36 
- #37

PR #4155, opened 2026-07-23
Fix conductor web plugin URLs resolving against the wrong origin

Shrey5132:
## Summary
The Python (and Source) Stepper side-content tab is stuck on "Loading the stepper…" in production, even after #59 in `source-academy/plugins` fixed the plugin directory's `.js`→`.mjs` mismatch.

The real bug is in `resolveWebPluginUrl()` (`src/commons/sagas/helpers/conductorEvaluatorCache.ts`): it returns the plugin directory's `resolutions.web` string as-is — e.g. `"./web/stepper/index.mjs"`, which is relative to where `directory.json` itself lives (`https://source-academy.github.io/plugins/`). That string is handed straight to a dynamic `import()` in `importAndRegisterWebPlugin`. A relative specifier passed to `import()` resolves against the **importing module's own URL**, not against wherever the string logically "came from" — so in production this resolves to a nonsensical same-origin path like `https://sourceacademy.org/static/js/web/stepper/index.mjs`, which just serves the SPA's `index.html` for any unmatched route. The dynamic import then fails with a MIME-type error, `StepperHostPlugin` never constructs, and the placeholder tab never gets replaced.

This explains why the plugins-repo fix (#59) didn't actually resolve the user-visible bug — that fix was necessary (the directory data was wrong) but not sufficient, since the frontend never even reaches `source-academy.github.io` for this fetch.

## Fix
Resolve the plugin's URL against the plugin directory's own address (`directory.plugin.url`) via `new URL(url, directoryUrl)` before returning it from `resolveWebPluginUrl()`, so `import()` always receives an absolute URL. (`new URL()` is a no-op passthrough if the resolution is already absolute, so this is safe for any future plugin that lists a fully-qualified URL too.)

## How this was found
Reproduced against the *live production* plugin directory and evaluators using a local dev server + Playwright: with `conductor.enable` toggled on, selecting Python and opening the Stepper tab showed the runner correctly loading `PyStepperEvaluator1.js`, but the web-plugin `import()` went to `sourceacademy.org/static/js/web/stepper/index.mjs` (served `index.html`, MIME-type error, "Conductor: failed to load web plugin \"stepper\""). After the fix, the same flow requests `source-academy.github.io/plugins/web/stepper/index.mjs` (200) and the Stepper tab renders real content end to end.

## Test plan
- [x] `npx tsc -b` — clean
- [x] `npx eslint src/commons/sagas/helpers/conductorEvaluatorCache.ts` — clean
- [x] `TZ=Asia/Singapore npx vitest run` — 678 passed, 0 failed
- [x] Manual end-to-end verification via local dev server against production plugin directory/evaluators (see above) — confirmed broken before the fix, working after

https://claude.ai/code/session_01PtafwFAYacm4Xyjgnq5yE2

### DIFF
--- a/src/commons/sagas/helpers/conductorEvaluatorCache.ts
+++ b/src/commons/sagas/helpers/conductorEvaluatorCache.ts
@@ -4,6 +4,7 @@ import { ModuleLoaderWebPlugin } from '@sourceacademy/web-module-loader';
 import type { SagaIterator } from 'redux-saga';
 import { call, select } from 'redux-saga/effects';
 import { selectDirectoryModulesUrl } from 'src/features/directory/flagDirectoryModulesUrl';
+import { selectDirectoryPluginUrl } from 'src/features/directory/flagDirectoryPluginUrl';
 
 import type { BrowserHostPlugin } from '../../../features/conductor/BrowserHostPlugin';
 import { createConductor } from '../../../features/conductor/createConductor';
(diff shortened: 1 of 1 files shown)


martin-henz (approved):
LGTM

PR #4159, opened 2026-07-23
Enable Conductor by default; fix Python None/builtin display and expression-statement coloring in CSE machine

### LINKED ISSUE #302: CSE Machine Visualization: misplaced quotation marks around functions on stash
Image

The quotation marks indicate a string, but these are function objects.

### LINKED ISSUE #4074: Conductor by default
I bet all developers use conductor by now. I suggest that we switch the frontend to use conductor.enable=true by default. The variable can be set back to false when needed.

### LINKED ISSUE #4111: CSE machine visualization: None should be None
Image

Image

Instead of
`null`
the stash should show
`None`

Akshay-2007-1:
## Summary
- Flip `conductor.enable`'s default to `true` — most development already happens through Conductor, so it shouldn't need to be turned on manually (fixes #4074).
- `CseSnapshotAdapter` mapped Python's `None` straight to JS `null`, which `PrimitiveValue`/`Text` then stringified as `"null"` on bindings and the stash. It now returns a source-object sentinel (same pattern already used for float/complex) so it renders via `toReplString()` as `"None"` (fixes #4111).
- `CseSnapshotAdapter`'s function-value branch only handled values carrying `metadata.closureFrameId`. Builtins (`abs`, `print`, ...) have none, so they fell through to the generic string fallback, and `StashItemComponent`'s `typeof val === 'string'` check then wrapped the bare name in quotes as if it were a Python string. Builtins now render as a plain JS function (picked up by `isBuiltInFn()`) so the unquoted name displays correctly (fixes source-academy/py-slang#302).
- Threaded py-slang's new `expressionStatement` control-item tag (source-academy/py-slang#270) through to `ControlItemComponent`, which now renders expression statements in a distinct color instead of being indistinguishable from a bare sub-expression until the "pop" that discards their value shows up on a later step.

## Test plan
- [x] `vitest run src/features/cseMachine/CseSnapshotAdapter.test.ts` — added regression tests for the `None` binding and the builtin-function stash value, all 11 tests pass.
- [x] Manually verify in the Python playground: a variable/stash value of `None` now shows `None`.
- [ ] Manually verify in the Python playground: a builtin like `abs` on the stash shows unquoted, and an expression statement's control item renders in the new color.


martin-henz (approved):
Looks good, as a temporary pragmatic fix.

PR #4202, opened 2026-07-30
Revert "Configure Louis chatbot for the Python SICP textbook"

RichDom2185:
Reverts source-academy/frontend#4085.

Breaks deploy


sayomaki (approved):
For deployment
LGTM

PR #4206, opened 2026-07-30
Remove ESLint step from yarn format

RichDom2185:
## Summary

`yarn format` was chaining `oxfmt` with `eslint --fix src`. This drops the
ESLint step so that `yarn format` only runs the formatter.

Linting is unchanged — it is still available via the standalone `yarn lint`
script (`eslint src`).

## Changes

- `package.json`: `format` script now runs only `oxfmt ...` (no longer chained
  with `eslint --fix src`).

## Test plan

- [ ] Run `yarn format` and confirm it only invokes `oxfmt` and no longer runs
  ESLint.
- [ ] Confirm `yarn lint` still works as before.


sayomaki (approved):
LGTM, makes `yarn format` so much better

PR #4216, opened 2026-07-31
Use more neutral language for the multi-language platform

martin-henz:
## Summary
Source Academy now hosts multiple languages via the Language Directory (Python/py-slang alongside Source/JavaScript - source-academy/py-slang#359), but three user- and admin-facing strings still described the whole platform as JavaScript-only. Found during a sweep for JavaScript remnants (same spirit as source-academy/py-slang#372/#373 on the py-slang side).

- **`DropdownAbout.tsx`**: the "About" dialog said students "write and run their programs... using sublanguages of JavaScript called Source." Now says the platform covers multiple languages, naming Source and Python as examples.
- **`IntroductionHelper.ts`**: `MAIN_INTRODUCTION` (the Playground's fallback welcome text, shown when a Conductor language doesn't supply its own `welcome` string via the language directory) opened with "The book ... uses JavaScript sublanguages that we call Source" as if that described the whole platform. Reworded to scope the JavaScript claim to "this language" instead - still accurate (this text only ever renders for Source/JS chapters, per the file's own `// TODO: Remove this after migrated to language config`), without asserting it platform-wide.
- **`PixelbotConfigPanel.tsx`**: `DEFAULT_ANSWER_PROMPT`, the default AI-tutor system prompt for the Pixelbot feature, unconditionally instructed the model to use "Source" syntax and forbade JavaScript features not in Source - actively wrong guidance if a Python-track course admin uses the default prompt as-is. Reworded to match the course's actual language, keeping the Source-specific guidance conditional.

Two other findings from the same sweep were considered but left as-is (out of scope for this PR): a hardcoded `mode="javascript"` in the admin assessment-template editor, and a JS-flavored placeholder in the course-level LLM prompt field - both minor/cosmetic rather than actively wrong.

## Test plan
- [x] `npx tsc -b --noEmit` - clean
- [x] `npx vitest run` (full suite) - 90 files / 746 tests passing (2 snapshot tests updated to reflect the intentional text change - `Markdown.test.tsx`, `Playground.test.tsx`)
- [x] `npx eslint` - clean
- [x] `npx oxfmt --list-different` - clean
- [x] Pre-push hook (tsc/lint/format:ci/test) - clean

Note: as with #4215, this environment's default Node (20.19.1) can't run the test suite (`jsdom`/`undici` incompatibility, reproduces on clean `master` too) - verified against Node 22 directly, matching CI.

### DIFF
--- a/src/commons/__snapshots__/Markdown.test.tsx.snap
+++ b/src/commons/__snapshots__/Markdown.test.tsx.snap
@@ -11,18 +11,7 @@ exports[`Markdown page renders correctly 1`] = `
     
 
     <p>
-      The book 
-      <a
-        href="https://sourceacademy.org/sicpjs/"
-        rel="noopener noreferrer"
-        target="_blank"
-      >
-        <em>
-          Structure and Interpretation of Computer Programs, JavaScript Edition
-        </em>
-      </a>
-      
-uses JavaScript sublanguages that we call 
+      This language is 
       <a
         href="https://docs.sourceacademy.org/"
         rel="noopener noreferrer"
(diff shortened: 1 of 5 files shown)


RichDom2185:
> Note: as with https://github.com/source-academy/frontend/pull/4215, this environment's default Node (20.19.1) can't run the test suite (jsdom/undici incompatibility, reproduces on clean master too) - verified against Node 22 directly, matching CI.

@martin-henz please update your NodeJS version to something not EOL. The repo expects Node 22

PR #4220, opened 2026-08-01
fix: conditionally show LLM API key

RichDom2185:
## Summary
- A previously entered `llmApiKey` was being dispatched with the create-course payload even after the user toggled LLM grading off. Hiding the input was purely a UI concern.
- `DropdownCreateCourse.submitHandler` now omits `llmApiKey` from the payload when `enableLlmGrading` is `false`.
- Added `DropdownCreateCourse.test.tsx` covering:
  - Enter a key → disable grading → submit → key is absent from payload.
  - Enter a key while grading is enabled → submit → key is present.

## Validation
- `yarn tsc -b` clean
- `yarn eslint src/commons/dropdown/DropdownCreateCourse.{tsx,test.tsx}` clean
- `yarn test` — 91 files / 748 tests pass


sayomaki:
I thought the plan we agreed on was just to remove the option to enable/disable from the create course dropdown? In that case course admins would still be able to enable it afterwards, and then add in the required API key.

PR #4221, opened 2026-08-01
Add local-file ("folder") imports for py2js (§2+)

martin-henz:
## Summary
Frontend-side support for py-slang's local-file ("folder") imports feature (source-academy/py-slang#379, merged): the Playground's multi-file Folder Mode already works with py2js's new local-import bundler out of the box (conductor's existing `requestFile` already backs it), but this PR adds two pieces of language-aware polish:

- **`foldersEnabled`**: reads the Language Directory's new `foldersEnabled` field (source-academy/language-directory#54) to disable the Folder Mode toggle for languages that don't support local imports — currently just Python §1, which has no pair/list library to build the exports-transfer structure they rely on.
- **`defaultFileExtension`**: reads the Language Directory's new `defaultFileExtension` field (source-academy/language-directory#55) so the Playground's fallback single-file entrypoint is named `program.py` for Python (was previously hardcoded to `program.js` for every language).

Also bumps `@sourceacademy/language-directory` to the tagged `0.0.20` release and drops the temporary type-cast this branch was using for `foldersEnabled` before that type was published.

## Notes for reviewers
- `defaultFileExtension` (source-academy/language-directory#55) isn't tagged yet, so `getDefaultFileExtension` still uses the same temporary-cast pattern `foldersEnabled` originally did — safe to merge either order, and the cast can be dropped once that PR lands and this repo bumps again.
- While testing this, found and fixed a stale UI snapshot: with the real bundled directory data, Python §1 (`foldersEnabled: false`) is the app's default selected language (`languages[0]`), so the Playground's Folder Mode button now correctly renders disabled by default. This is an intended consequence of the `foldersEnabled` feature that a pre-existing Node-version issue in local test runs had prevented from surfacing until now.

## Test plan
- [x] `yarn tsc` / `yarn lint` / `yarn format:ci` (via the pre-push hook)
- [x] `yarn test` — full suite green (90 files, 746 tests), including the updated Playground snapshot
- [x] Manually verified live in-browser earlier in this work (Folder Mode toggle correctly disabled for Python §1, correctly enabled and functional for §2–4)

https://claude.ai/code/session_011b8eJsqoDGiuesf5uQ42oG

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "@sourceacademy/common-cse-machine": "^0.3.0",
     "@sourceacademy/common-tabs": "^0.0.1",
     "@sourceacademy/conductor": "^0.8.3",
-    "@sourceacademy/language-directory": "https://github.com/source-academy/language-directory.git#0.0.20",
+    "@sourceacademy/language-directory": "https://github.com/source-academy/language-directory.git#0.0.21",
     "@sourceacademy/plugin-directory": "https://github.com/source-academy/plugin-directory.git#0.0.4",
     "@sourceacademy/runner-module-loader": "^0.1.0",
     "@sourceacademy/sharedb-ace": "2.1.1",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "@sourceacademy/common-cse-machine": "^0.3.0",
     "@sourceacademy/common-tabs": "^0.0.1",
     "@sourceacademy/conductor": "^0.8.3",
-    "@sourceacademy/language-directory": "https://github.com/source-academy/language-directory.git#0.0.19",
+    "@sourceacademy/language-directory": "https://github.com/source-academy/language-directory.git#0.0.20",
     "@sourceacademy/plugin-directory": "https://github.com/source-academy/plugin-directory.git#0.0.4",
     "@sourceacademy/runner-module-loader": "^0.1.0",
     "@sourceacademy/sharedb-ace": "2.1.1",
--- a/src/commons/controlBar/ControlBarToggleFolderModeButton.tsx
+++ b/src/commons/controlBar/ControlBarToggleFolderModeButton.tsx
@@ -8,6 +8,12 @@ type Props = {
   isSessionActive: boolean;
   isPersistenceActive: boolean;
   toggleFolderMode: () => void;
+  /** Whether the current language/evaluator supports folder mode at all —
+   * from the Language Directory's `foldersEnabled` (defaults to `true` when
+   * the field is omitted, so callers on languages predating this field are
+   * unaffected). `false` for e.g. Python §1, which has no pair/list library
+   * to build the exports-transfer structure local-file imports rely on. */
+  foldersEnabled?: boolean;
 };
 
 function ControlBarToggleFolderModeButton({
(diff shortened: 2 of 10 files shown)


martin-henz:
## Review

### Overview

This PR wires the frontend up to two new `@sourceacademy/language-directory` fields:

- **`foldersEnabled`** — disables the Playground's Folder Mode toggle for languages whose evaluator(s) don't support local-file imports (currently Python §1), reading the field now that `language-directory` is bumped to the tagged `0.0.20`.
- **`defaultFileExtension`** — replaces a previously hardcoded `program.js` fallback filename (used whenever a workspace has zero open editor tabs) with a per-language extension (`program.py` for Python), via a new `getDefaultFileExtension` helper. This field isn't tagged upstream yet, so the helper still casts around the missing type, mirroring how `foldersEnabled` was handled before its own tag existed.

Also includes a UI snapshot fix: with real bundled directory data, Python §1 (`foldersEnabled: false`) turns out to be `languages[0]` — the app's default selected language — so the Folder Mode button now correctly renders disabled on first load. The PR body calls this out explicitly as an intended consequence, not a regression.

### Code quality & style

- Both wiring sites (`WorkspaceSaga/index.ts`'s `setFolderMode` handler and `evalEditor.ts`'s `evalEditorSaga`) reach for the current language with the **identical** four-line `yield select(...)` block:
  ```ts
  const extension: string = yield select((state: OverallState) => {
    const { selectedLanguageId: langId, languageMap } = state.languageDirectory;
    return getDefaultFileExtension(langId ? languageMap[langId] : undefined);
  });
  ```
  Worth collapsing into one shared selector (e.g. `selectDefaultFileExtension(state: OverallState): string` alongside `getDefaultFileExtension` in `LanguageDirectoryTypes.ts`), so the "resolve current language, then extract the field" logic lives in one place rather than two.
- `getDefaultFileExtension`'s parameter type is `ILanguageDefinition | null | undefined`, but neither call site ever actually passes `null` (`langId ? languageMap[langId] : undefined` is `ILanguageDefinition | undefined`) — minor, harmless type over-generality.
- The temporary cast in `getDefaultFileExtension` is well-documented (links directly to `language-directory#55`), consistent with how the now-removed `foldersEnabled` cast was documented before it was dropped.
- Good, minimal snapshot diff — only the two lines that should have changed did.

### Specific suggestions

1. Consolidate the duplicated `yield select(...)` block (above) into one exported selector helper shared by both sagas.
2. Add a regression test for the actual new behavior (see Test coverage below) rather than relying solely on the snapshot picking it up incidentally.
3. Minor: drop `| null` from `getDefaultFileExtension`'s parameter type, or keep it if you want the function to double as a general-purpose helper beyond these two call sites — either is fine, just worth a conscious choice rather than leftover breadth.

### Potential issues / risks

- **Test coverage gap**: nothing in this PR directly exercises `getDefaultFileExtension` producing `"py"` for a Python-like language, or asserts that `evalEditorSaga`/`setFolderMode` actually build `program.py` (as opposed to `program.js`) when the current language sets `defaultFileExtension`. The existing `WorkspaceSaga.test.ts` assertions (`'/playground/program.js'`) all pass because their mock language configs simply don't set the field, defaulting to `"js"` — they don't cover the new path at all. The Playground snapshot fix incidentally proves `foldersEnabled` works end-to-end (because Python §1 happens to be the default test language), but nothing similarly proves `defaultFileExtension` does. A small saga test with a mock `languageDirectory` state (`defaultFileExtension: 'py'`) asserting the resulting file path would close this gap.
- **Sequencing risk (called out by the author, correctly)**: `defaultFileExtension` isn't tagged in `language-directory` yet, so this PR's behavior for that field depends on runtime JSON matching a type the pinned package doesn't declare. Low risk in practice (same pattern already proven safe by `foldersEnabled`'s own history), but worth double-checking after `language-directory#55` merges and this repo re-bumps, that the temporary cast is actually removed and not forgotten (the PR body says so, which is the right call to leave for a human, but flagging it here too since it's easy for a "quick removal" TODO to get lost).
- **Trust assumption on `defaultFileExtension`'s value**: it's string-interpolated directly into a file path (`` `program.${extension}` ``) with no validation. Not exploitable today (the directory data is either the bundled npm package or a URL gated behind `flagDirectoryLanguageUrl`, i.e. already a trusted-config surface — same trust level as `evaluators[].path`, which is already loaded as a script URL elsewhere), but if this field or its source ever becomes more dynamic/user-influenced, that assumption should be revisited.

### Test coverage

Full suite passes (per the PR's own test plan) and the one affected snapshot was correctly updated. As noted above, the gap is that nothing *new* is asserted for `defaultFileExtension`'s actual effect — only `foldersEnabled` gets incidental end-to-end proof via the snapshot.

### Security

No new attack surface beyond the pre-existing trust level of language-directory data (see above). No user input flows into either changed code path.

### Performance

Negligible — one extra `select()` per saga invocation, same shape as selectors already used nearby in both files.

PR #4236, opened 2026-08-02
fix: show Pixel chatbot in playground on backend-enabled deployments

### LINKED ISSUE #4234: Pixel bot: not in playground?
Currently the pixel bot does not appear in playground. It appears in the assessments and in Achievements.

martin-henz:
## Summary
In the full-academy router config (`getFullAcademyRouterConfig`, used by backend-enabled deployments like sourceacademy.nus.edu.sg), `playground/:playgroundCode?` was a sibling route of `courses/:courseId` rather than a child of it. `<RagChatbot />` (Pixel) was only mounted inside `courses/[courseId]/_layout.tsx`, so Assessments and Achievements (children of `courses/:courseId`) got Pixel, but Playground didn't.

The playground-only (serverless, e.g. sourceacademy.org) deployment is unaffected — it uses a separate router config that never included the chatbot, which is correct/intentional.

## Fix
Rather than nesting `playground` under `courses/:courseId` (which would change its URL to `/courses/:courseId/playground/...` and break existing links), this factors the chatbot mount out into a new pathless layout route (`src/new_routes/_chatbotLayout.tsx`) shared by both `courses/:courseId` and `playground`. `<RagChatbot />` is removed from `courses/[courseId]/_layout.tsx` since the new shared layout now owns it. `FloatingChatbot` renders with `fixed` positioning so moving its mount point doesn't affect layout.

Fixes #4234

## Testing
- `yarn tsc --noEmit`
- `yarn lint` (via eslint on changed files)
- `yarn vitest run src/new_routes src/commons/application/ApplicationWrapper` (5 passed)
- Full pre-push hook (tsc/lint/format/test) passed under Node 22 (matching CI's `node-version: 22`)

Note: my local Node (20.19.1, via nvm) fails the whole suite with a pre-existing `jsdom`/`undici` incompatibility unrelated to this change (reproduces identically on a clean `master` checkout); verified everything under Node 22.17.0 instead.


RichDom2185 (changes requested):
This violates our coding conventions on file names

RichDom2185:
Let me fix this by EOD

martin-henz:
@RichDom2185 is this naming better?

PR #4239, opened 2026-08-02
Fix Autograder Run not triggering all testcases

### LINKED ISSUE #4232: Autograder (client-side): Not triggered by "Run"
The autograding tab should run all test cases when "Run" is pressed. That currently doesn't work. Test cases can be run individually by clicking on them.

Image

Akshay-2007-1:
Fixes #4232.

## Summary
Three separate bugs contributed to "Run" not triggering all testcases from the Autograder tab:

- **The actual crash** (`ConductorInternalError: Conduit already terminated`, uncaught, aborting the whole batch): `runAllTestcases` switched `state.languageDirectory`'s selection to force the test-case language/evaluator and back for **every single testcase**. Each switch tears down and rebuilds the whole Conductor session (`ensureConductorSessionSaga` in `conductorEvaluatorCache.ts`). With N testcases that's up to 2N+1 rapid session teardown/rebuild cycles in a row - on a real network connection (reproduced on staging, not locally, since it's a latency-dependent race) an in-flight async teardown could race a new one and throw uncaught from inside the vendored `@sourceacademy/conductor` library. Now switches once for the whole batch and restores once afterward, regardless of testcase count.
- `EditingWorkspace.tsx` (the question-editing/preview "Mission Control" view) had **no** autograder-tab check at all - Run always just evaluated the editor regardless of which side-content tab was active.
- `GradingWorkspace.tsx` fired **both** a plain editor eval and the full testcase run back-to-back when the Autograder tab was active (the testcase run itself already re-evaluates the editor internally).

## Test plan
- [x] `yarn tsc --noEmit`, `yarn eslint`, `yarn oxfmt` all clean
- [x] `WorkspaceSaga.test.ts` (115 tests, including a new one asserting the language/evaluator switch happens exactly once per batch instead of once per testcase) passing
- [x] Verified locally against a real mission with testcases: Run correctly triggers all testcases from the Autograder tab in the student assessment view

### DIFF
--- a/src/commons/editingWorkspace/EditingWorkspace.tsx
+++ b/src/commons/editingWorkspace/EditingWorkspace.tsx
@@ -50,6 +50,7 @@ import Markdown from '../Markdown';
 import SideContentToneMatrix from '../sideContent/content/SideContentToneMatrix';
 import type { SideContentProps } from '../sideContent/SideContent';
 import { changeSideContentHeight } from '../sideContent/SideContentActions';
+import { useSideContent } from '../sideContent/SideContentHelper';
 import type { SideContentTab } from '../sideContent/SideContentTypes';
 import { SideContentType } from '../sideContent/SideContentTypes';
 import { useAppSelector } from '../utils/Hooks';
@@ -91,6 +92,8 @@ function EditingWorkspace(props: EditingWorkspaceProps) {
     currentAssessment: storedAssessmentId,
     currentQuestion: storedQuestionId,
   } = useAppSelector(store => store.workspaces[workspaceLocation]);
+
+  const { selectedTab } = useSideContent(workspaceLocation);
 
   /**
    * After mounting (either an older copy of the assessment
@@ -328,6 +331,23 @@ function EditingWorkspace(props: EditingWorkspaceProps) {
     const editorTestcases = [testcase];
     handleUpdateWorkspace({ editorTestcases });
     dispatch(WorkspaceActions.evalTestcase(workspaceLocation, 0));
+  };
+
+  const handleRunAllTestcases = () => {
+    const question = assessment!.questions[formatedQuestionId()] as IProgrammingQuestion;
+    const editorTestcases = question.testcases.concat(question.testcasesPrivate ?? []);
+    handleUpdateWorkspace({ editorTestcases });
+    dispatch(WorkspaceActions.runAllTestcases(workspaceLocation));
+  };
+
+  // Run all testcases (instead of just evaluating the editor) when the Run button/shift-enter
+  // is used while the Autograder tab is active - matches AssessmentWorkspace/GradingWorkspace.
+  const handleEval = () => {
+    if (selectedTab === SideContentType.editorAutograder) {
+      handleRunAllTestcases();
+    } else {
+      handleEditorEval();
+    }
   };
 
   const handleSave = () => {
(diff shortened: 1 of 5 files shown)


martin-henz (approved):
Looks good!

Can't wait to get rid of all these conductorEnabled checks.

martin-henz:
Claude code review, FYI:

PR #4239: Fix Autograder Run not triggering all testcases

Fixes: #4232 — clicking "Run" from the Autograder tab was silently aborting the batch partway through.

Overview

Three independent bugs bundled into one fix:

1. The actual crash — runAllTestcases (in WorkspaceSaga/index.ts) used to switch the shared languageDirectory selection to the test-case language/evaluator and back for every testcase. Under Conductor, each switch tears down and rebuilds the whole session (ensureConductorSessionSaga in conductorEvaluatorCache.ts — confirmed by reading that file: a session is torn down whenever selectedLanguageId changes). With N testcases that's up to 2N+1 rapid teardown/rebuild cycles, which on real network latency could race an in-flight async teardown and throw an uncaught Conduit already terminated, aborting the batch. Fixed by switching once for the whole batch (guarded by isConductorEnabled, since the legacy js-slang path never touches languageDirectory at all) and restoring once in a finally.
2. EditingWorkspace.tsx had no autograder-tab check on Run at all — verified against the repo: resetWorkspaceValues() never populates editorTestcases for this workspace, so the new handleRunAllTestcases correctly builds it from question.testcases.concat(question.testcasesPrivate ?? []) before dispatching. This mirrors AssessmentWorkspace/GradingWorkspace, using the workspace-specific SideContentType.editorAutograder enum value (correctly distinct from SideContentType.autograder).
3. GradingWorkspace.tsx fired both a plain editor eval and the full testcase run back-to-back — removed, since runAllTestcases already re-evaluates the editor internally (evalEditorSaga at the top of the saga handler).

Verification performed

- Read the full diff and the surrounding files (runTestCase.ts, WorkspaceSaga/index.ts, conductorEvaluatorCache.ts, testCaseLanguage.ts, AssessmentWorkspace.tsx, GradingWorkspace.tsx, EditingWorkspaceSideContentAutograderTab.tsx) to confirm the story holds together — it does.
- Checked out the PR branch in an isolated worktree and ran yarn tsc --noEmit (clean) and the saga test file (65/65 passing, including the new batch-switch-count test).
- Confirmed the individual per-testcase "click to test" button (AutograderTab → handleTestcaseEval) still goes through runTestCase with skipLanguageSwitch unset, so it keeps its own switch/restore — consistent with the code comment explaining why.
Code quality- The try/finally restore pattern is correct: an early returramSucceeded === false) still runs the finally block, so thelanguage selection is never left stranded mid-batch.
- Threading skipLanguageSwitch through runTestCase → runTestpositional boolean arg is a little easy to misread at call
sites (runTestCase, workspaceLocation, idx, isConductorEnabllag as the skip flag is correct but non-obvious without the
comment). The doc comments carry the weight here and are tho't a real risk, just a minor readability note.
- isConductorEnabled is selected twice (once in runAllTestcaTestCase call) — harmless, negligible duplication, not worth
changing.
- Comments are dense but genuinely explain non-obvious invarmantics, race conditions, why the single-testcase path is
untouched) rather than restating code — appropriate given ho

Risks

- None found. The fix is narrowly scoped, gated correctly bend the new test directly targets the regression (asserts theswitch happens exactly once per batch, not once per testcase).

Overall: solid, well-tested fix. No blocking issues.

PR #4245, opened 2026-08-04
Correct off-by-one prepend line shift in Conductor error messages

### LINKED ISSUE #4244: Assessments: In error messages, PREPEND lines are counted
If the PREPEND has one line, the line numbers are off-by-one in the error messages.

martin-henz:
## Summary
- Fixes #4244: under Conductor, `evalEditorSaga` concatenates `${prepend}\n${entrypoint}` before sending it to the evaluator, so a reported error's line number was relative to that combined text — a one-line PREPEND made every error line off-by-one.
- `toConductorSourceError` now uses the evaluator's structured `line`/`column`/`rawMessage` fields (duck-typed, since a Worker's `postMessage` clone drops the original `EvaluatorError` prototype) and subtracts the same prepend-line offset already applied to `breakpointLines`, instead of passing through the evaluator's pre-baked message string untouched.
- The offset is tracked per Conductor REPL session via a mutable ref, reset to 0 for a later REPL chunk sent to an already-live session (a REPL line has no prepend concatenated onto it, unlike the session's initial entrypoint chunk).

## Test plan
- [x] `tsc --noEmit` — clean
- [x] `yarn test src/commons/sagas/WorkspaceSaga.test.ts` — 67/67 passing, including two new cases:
  - a one-line PREPEND shifts a reported error line back to the student's own line
  - a REPL chunk sent to an already-live session is not shifted by that session's own prepend
- [x] `eslint` — clean

https://claude.ai/code/session_01HZ2e6QFxwCn1rkTjdhKCSS

### DIFF
--- a/src/commons/sagas/WorkspaceSaga/helpers/evalCode.ts
+++ b/src/commons/sagas/WorkspaceSaga/helpers/evalCode.ts
@@ -51,7 +51,49 @@ import { selectWorkspace } from '../../SafeEffects';
 import { dumpDisplayBuffer } from './dumpDisplayBuffer';
 import { updateInspector } from './updateInspector';
 
-function toConductorSourceError(error: unknown): SourceError {
+/**
+ * `lineOffset` is the number of prepend lines concatenated onto the entrypoint before it was sent
+ * to the evaluator (see evalEditorSaga's Conductor branch, which joins `${prepend}\n` onto the
+ * entrypoint, and evalCodeConductorSaga's own `preludeLineOffset`, which shifts breakpointLines the
+ * other way for the same reason) - without subtracting it back out here, an error at the student's
+ * own line 1 would be reported as `prepend line count + 1` (source-academy/frontend#4244).
+ *
+ * A Worker's postMessage clones a thrown EvaluatorError via the structured clone algorithm, which
+ * drops its prototype chain (EvaluatorError's `name` isn't a recognised built-in Error subtype, so
+ * the clone lands on plain `Error.prototype`) but keeps custom own properties like
+ * `line`/`column`/`rawMessage` intact - so this checks shape rather than `instanceof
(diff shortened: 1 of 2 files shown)


src/commons/sagas/WorkspaceSaga/helpers/evalCode.ts
```diff
@@ -51,7 +51,49 @@ import { selectWorkspace } from '../../SafeEffects';
...
+ * the clone lands on plain `Error.prototype`) but keeps custom own properties like
+ * `line`/`column`/`rawMessage` intact - so this checks shape rather than `instanceof
+ * EvaluatorError`.
+ */
+function toConductorSourceError(error: unknown, lineOffset: number = 0): SourceError {
+  if (
+    typeof error === 'object' &&
+    error !== null &&
+    'line' in error &&
+    typeof (error as { line: unknown }).line === 'number'
```
Akshay-2007-1, line 72: the check is only 'line' in error && typeof error.line === 'number', there's no discriminant tying this to an actual EvaluatorError (e.g. checking name). If a structured object happens to have a numeric .line but no rawMessage and no .message, String(evaluatorError.message) evaluates to the literal string "undefined", which would be shown to the student as the error explanation. Low probability given EvaluatorError always sets these fields, but there's no test covering that partial-shape case.

Akshay-2007-1 (approved):
LGTM mostly. Its a narrow scoped fix. A few questions I had : 

Errors truly located inside the prepend get silently relabeled as line 1. Is this documented somewhere? I am okay with this behavior since it does make sense and normally there should never be errors in the PREPEND. 

Note that non-structured Conductor errors still get line: 0 (the pre-existing fallback branch, unchanged): fine as pre-existing behavior, just noting the fix only improves the structured-error path, not this baseline.

PR #4251, opened 2026-08-04
Clear side content tabs when switching questions

### LINKED ISSUE #4248: Left-over tabs when navigating between workspaces
Image

After visiting runes and csg missions, the user is attempting an unrelated path, but the old tabs are still there.

Akshay-2007-1:
Fixes #4248

The side content tab list only got refreshed after running code, so switching to a different question kept showing module tabs (Runes, Curves, etc) from the previous one until you hit run again. Now it clears out as soon as the question actually changes, in the assessment, grading, and question editor workspaces.

Tested locally against a seeded backend by opening a path with Runes/Curves questions, switching to an unrelated path without running any code, and confirming the leftover tabs are gone.


martin-henz (approved):
Great. Looks like an easy fix!

PR #4271, opened 2026-08-09
Modified Pixel for Python 

yiilinzhang:
### Description

This PR adds 2 main features for pixel
1. Allow the admin to upload course documents for pixel via the admin panel. This generates an ai description and saves it to the s3 bucket. This supports the update to python where the new slides can be added to the store when it is ready and the prompt can also be tested and tweaked on the go
2. Let the chatbot view the current page content so it can provide better responses
3. Fixed UI bug ther louis cannot be minimised after it is expanded

### Type of change

### How to test

1. Pull both the frontend and the backend and run the necessary migrations
2. An S3 bucket also has to be provided with aws creds
4. Navigate to the admin user login admin panel for pixel setting. After documents have been uploaded the user can test out the chatbot

### Checklist

- [x] I have tested this code


src/features/adminPanel/subcomponents/PixelbotDocumentsPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+                              type="button"
+                              className={classNames(classes.iconButton, classes.iconButtonDanger)}
+                              onClick={() => setConfirmDeleteId(doc.id)}
+                              aria-label="Delete document"
+                            >
+                              <Icon icon={IconNames.TRASH} size={13} />
+                            </button>
+                          </span>
+                        </div>
+                  ))}
```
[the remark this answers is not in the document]
yiilinzhang: reran format

src/features/adminPanel/subcomponents/PixelbotDocumentsPanel.tsx
```diff
@@ -0,0 +1,410 @@
...
+                    onClick={e => handleDeleteCategory(e, category)}
+                    aria-label="Delete category"
+                  >
+                    <Icon icon={IconNames.TRASH} size={13} />
+                  </button>
+                </span>
+              </div>
+
+              {isOpen && docs.length > 0 && (
+                <div className={classes.categoryBody}>
```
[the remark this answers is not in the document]
yiilinzhang: intended design

src/components/ui/chatbot/FloatingChatbot.tsx
```diff
@@ -163,6 +164,7 @@ function FloatingChatbot({
         <FloatingChatbotButton
           src={avatarSrc}
           alt={avatarAlt}
+          positionClassName={isOpen ? 'right-2.5 bottom-full mb-2' : 'right-2.5 bottom-2.5'}
```
sayomaki: This seems rather clunky and makes it hard to figure out how the button will look like just from the `FloatingChatbotButton` component alone, and needing to come back here to tell which classes are applied. Instead, you should directly pass the `isOpen` into the component, and have the component decide which classes to apply.

src/pages/academy/ragChatbot/RagChatbot.tsx
```diff
@@ -1,17 +1,36 @@
...
 import RagMessageRenderer from './RagMessageRenderer';
 
 const init = (tokens: Tokens) => initRagChat(tokens);
-const send = (tokens: Tokens, userInput: string) => sendRagMessage(tokens, userInput);
 
 function RagChatbot() {
   const { feedbackUrl } = useSession();
+  const assessmentId = useAppSelector(state => state.workspaces.assessment.currentAssessment);
+  const questionId = useAppSelector(state => state.workspaces.assessment.currentQuestion);
+  const onAssessment = assessmentId !== undefined && questionId !== undefined;
```
sayomaki, line 21: This doesn't seem like the appropriate check here. The current implementation only ensures that if the user has ever opened an assessment workspace, it will load the context from that assessment. However, it fails to consider that it is actually never unloaded, if the user then changes to the playground page for example, it will still load in the assessment context. 

A suggested approach here would be to instead check within the `_chatbot/_layout.tsx` whether if the user is currently in an assessment by the path, and pass that information to this component. Alternatively you can instead fix the logic in assessments to properly also reset `currentAssessment` and `currentQuestion` as well.

src/pages/academy/ragChatbot/RagChatbot.tsx
sayomaki, line 43: Also minor thing, while its technically fine to have this as CS1101S for now, I do think it should read from the course information and maybe use the course short name instead. Unimportant, it's alright if you decide not to fix this for now.

src/features/adminPanel/subcomponents/AddDocumentsModal.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+function AddDocumentsModal({ isOpen, categories, tokens, onClose, onSaved }: Props) {
+  const [batch, setBatch] = useState<BatchFile[]>([]);
+  const [selectedFileId, setSelectedFileId] = useState<string | null>(null);
+  const [isSaving, setIsSaving] = useState(false);
+
+  const selected = batch.find(f => f.id === selectedFileId) ?? null;
+
+  // The upload endpoint needs a category up front; the one that actually sticks is chosen on save.
+  const uploadCategoryId = categories[0]?.id;
```
sayomaki: Can you maybe just slightly elaborate on this a bit? What is it actually used for, and what are initial values and how does it change when modified/saved?

sayomaki (changes requested):
Overall the changes are good, but there's a few things that I would like to point out and/or ask about.

src/components/ui/chatbot/FloatingChatbot.tsx
yiilinzhang: fixed

src/pages/academy/ragChatbot/RagChatbot.tsx
yiilinzhang: switched to routing method

src/features/adminPanel/subcomponents/AddDocumentsModal.tsx
yiilinzhang: the categories are things like "lectures", "reading assesments" that the user can add on it go. Its used in the document map to give the llm more context on what the file is about. 

But i updated the code slightly so this is only used on save and not on upload.

sayomaki (approved):
LGTM, but do remember to update the `category_id` changes on the backend! Will leave a comment there.

PR #4291, opened 2026-08-14
Exclude prepend from Conductor testcase error line numbers

Shrey5132:
## Summary
- `runTestCaseConductor` concatenates `prepend`/student code/`postpend`/testcase into one combined file before sending it to the Conductor evaluator, but never subtracted the prepend's line count back out of a resulting error's reported line - so an error on the student's own first line was reported at `prepend line count + 1`.
- This is the same off-by-N bug #4244/#4245 fixed for the plain Run path (`evalCode.ts`'s `toConductorSourceError`/`preludeLineOffset`), but that fix doesn't reach this call site: it's gated on an editor Run's `actionType`, and `runTestCaseConductor` calls `evalCodeSaga` with `EVAL_SILENT` instead. Testcase/grading errors in assessments were still shifted.
- Shifts the line back out locally in `runTestCase.ts`, using the same `prepend` value this saga already concatenates into the combined file. A location's line of `0` (toConductorSourceError's "no location info" fallback) is left untouched rather than shifted into a misleading line `1`.

## Test plan
- [x] Added a test asserting a testcase error's line number is corrected back to the student's own line (multi-line prepend case).
- [x] Added a test asserting a line-`0` (unknown location) error is left unshifted.
- [x] `yarn test src/commons/sagas/WorkspaceSaga.test.ts` - 69 passed.
- [x] `yarn eslint` on changed files - clean.
- [x] `tsc --noEmit` - clean.

https://claude.ai/code/session_01FYm9pttcp3nixtDLKiPh2k

### DIFF
--- a/src/commons/sagas/WorkspaceSaga.test.ts
+++ b/src/commons/sagas/WorkspaceSaga.test.ts
@@ -1175,6 +1175,108 @@ describe('EVAL_TESTCASE under Conductor (runTestCaseConductor)', () => {
       })
       .silentRun();
   });
+
+  test('excludes prepend from a Conductor testcase error location (source-academy/frontend#4244)', () => {
+    const context = createContext();
+    // Two lines, so combinedCode's own line 3 is `editorValue`'s line 1 - see
+    // runTestCaseConductor's combinedCode, which joins prepend/value/.../testcase with '\n'.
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Looks good!

PR #4381, opened 2026-09-17
fix(cseMachine): don't prune a frame that's still a closure's home (#4380)

### LINKED ISSUE #4380: CSE Machine visualization: toplevel function values not shown
If there is no global environment, the function values coming from lambda expressions are not visualized

Compare 
Image

```
dummy = 1
lambda x: x + 2
```

with 
Image

```
lambda x: x + 2
```

In the latter, the `closure` entry  on the stash does not have any function object to point to.

martin-henz:
## Summary
Fixes #4380.

- `isEmptyEnvironment`-based frame collapsing (`getNextChildren`, and `getVisibleChildren` under clear-dead-frames) treated any bindingless frame as boring and safe to hide, without checking whether it's still the defining environment of a live closure or array referenced elsewhere (stash, control, or another frame's bindings).
- When it was — e.g. a bare top-level `lambda x: x + 2` with no other statement, so its Python "Globals" frame has no bindings at all — the frame vanished from the layout, and the closure's arrow had nothing left to point to (the bug reported against py-slang as [py-slang#429](https://github.com/source-academy/py-slang/issues/429), traced here to the frontend and reopened as #4380).
- `computeLiveState` (`CseMachineUtils.ts`) now also returns `closureHomeEnvIds`: the subset of live environments that are some closure's/array's *own* environment, as opposed to merely reachable via another live environment's tail chain. Both frame-collapsing paths now skip an empty frame only when it's outside that set.
- Also includes an unrelated one-file reformat (`TypeHelper.ts`) — `oxfmt --list-different` already failed on it on a clean `master`, which blocked the pre-push hook for this branch.

## Test plan
- [x] Added a regression test (`CseSnapshotAdapter.test.ts`) that renders a Python-conductor snapshot with a bare stash closure whose home frame has zero bindings, both with and without clear-dead-frames, and asserts the frame and the closure's arrow both exist. Confirmed it fails on the pre-fix code and passes with the fix.
- [x] `yarn vitest run src/features/cseMachine`: all 31 tests pass, no regressions.
- [x] `tsc -b --noEmit` and `eslint` clean on changed files.


martin-henz:
@codex review

PR #4396, opened 2026-09-21
fix(cse): render Source's empty list and global frame correctly

martin-henz:
Companion to [source-academy/js-slang#2061](https://github.com/source-academy/js-slang/pull/2061), which adds Conductor CSE machine evaluators for Source §3–§4.

`CseSnapshotAdapter` was written against py-slang's shape. Running js-slang through it surfaces two places where that shape does not fit Source — plus one latent bug that turns out not to be Conductor-specific at all.

## 1. The empty list

The adapter maps the labels `nonetype`/`none`/`null` to a `{ toReplString: () => 'None' }` stand-in, deliberately avoiding the `ArrayNullUnit` visual. That is right for Python. It is wrong for Source, where the empty list **is** that visual and terminates every box-and-pointer diagram.

js-slang therefore labels Source's `null` as `empty_list`, which this PR maps to a real JS `null`. The Python labels are untouched.

## 2. The global frame

py-slang's global frame is genuinely empty — its builtins live in `nativeStorage` — so the adapter injects a `:::pre-declared names::` sentinel to stand in for them. js-slang puts every builtin in the global environment's own `head`, so that frame arrives with ~40 real bindings and the sentinel is a spurious extra row above them.

It is now injected only when the frame has no bindings of its own, so both shapes render correctly.

## 3. A latent bug the above exposed

`Frame` moves the sentinel to the front of the global frame's bindings:

```ts
const index = entries.findIndex(([key]) => key === Config.GlobalFrameDefaultText);
entries.unshift(entries.splice(index, 1)[0]);
```

When the sentinel is absent, `findIndex` returns `-1`, and `splice(-1, 1)` reads that as *"the last entry"* — silently promoting an unrelated binding to the front of the global frame.

Demonstrated by mutation rather than asserted: with the guard removed, a global frame bound `first`/`middle`/`last` renders as `last`/`first`/`middle`.

**This is not Conductor-specific.** js-slang's global frame has never carried the sentinel, so the legacy Source CSE tab has been hitting this too.

## Tests

Four cases added to `CseSnapshotAdapter.test.ts`, covering both shapes so neither regresses: the empty list becoming a real `null`, the sentinel being omitted for a populated global frame, the sentinel being *kept* for an empty one (py-slang), and binding order surviving a sentinel-less global frame.

`src/features/cseMachine` and `src/features/conductor`: 56 tests pass. `tsc`, `eslint` and `format:ci` clean.

## Note on scope

Diff is deliberately minimal — 3 changed lines plus comments, and 95 lines of tests. Worth mentioning for anyone reproducing this: running `prettier` over these files reformats them wholesale, because this repo formats with **`oxfmt`**, not prettier.


martin-henz:
@codex review

PR #4400, opened 2026-09-21
fix(cse): render NaN and Infinity as themselves, not as null

martin-henz:
Two bugs in the CSE machine's number path. Both are visible in the **global frame of every Source run**, because `NaN` and `Infinity` are predeclared globals there.

## 1. `JSON.stringify` renders non-finite numbers as the string `"null"`

`components/Text.tsx`:

```ts
? JSON.stringify(data) || String(data)
```

`JSON.stringify(Infinity)` is `"null"` — a **truthy string**, so the `|| String(data)` fallback never fired. `NaN`, `Infinity` and `-Infinity` all displayed as `null`.

This is in the **renderer**, not the snapshot adapter, so it affects the live Source CSE tab too — not just Conductor. It is long-standing rather than new.

## 2. The adapter turned `NaN` into `0`

`CseSnapshotAdapter.ts`:

```ts
const n = parseFloat(v.displayValue);
return isNaN(n) ? 0 : n;
```

`parseFloat('NaN')` is `NaN`, so the fallback rewrote it to `0`. The same fallback also turned *any* unreadable `displayValue` into `0` rather than admitting it did not know — and `parseFloat` stops at the first non-numeric character, so partial garbage parsed as a number.

Now the three non-finite spellings are recognised explicitly, parsing uses `Number` rather than `parseFloat`, and the `0` fallback is kept only for genuinely unparseable input.

## How these surfaced

I went looking for the **empty-list rendering** from #4396, to confirm it. It works: the `ArrayNullUnit` — the diagonal stroke terminating a box-and-pointer diagram — is drawn for `list(1, 2)`'s terminator.

The `null` I could see nearby turned out to be the `Infinity` binding two rows above it in the global frame:

```
NaN:=       0        ← bug 2
Infinity:=  null     ← bug 1
math_abs:=  "function math_abs(a) {…"
```

## Verification

Both fixes are **mutation-verified** — reverting either fails its own tests:

- reverting the adapter fix → `renders NaN as NaN, not 0` fails
- reverting the `Text` fix → all three of `NaN`/`Infinity`/`-Infinity` `does not render as null` fail

`src/features/cseMachine`: 42 tests pass. `tsc`, `eslint` and `format:ci` clean.

## Related, not fixed here

While investigating I found why the Source global frame is enormous: `CseMachineLayout.removeUnreferencedGlobalFns()` — which prunes the global frame to the builtins a program actually uses — early-returns when `Layout.snapshotMode` is true. Filed separately; it needs a different fix and probably an evaluator-side one.


martin-henz:
Good catch, and it is the exact failure mode I had been guarding against elsewhere in this PR — a test that can pass without asserting anything.

Fixed: the assertion is now unconditional.

Worth noting the guard was never load-bearing. The mutation check that validated this fix produced `expected 'null' not to be 'null'`, so `text` was always defined; the `if` was defensive clutter that happened to hide a hole rather than protect anything.

Re-verified by mutation after the change — reverting the `Text.tsx` fix now fails all three cases with the stronger message:

```
expected 'null' to be 'NaN'
expected 'null' to be 'Infinity'
expected 'null' to be '-Infinity'
```

42 tests pass in `src/features/cseMachine`.

@coderabbitai review

